PR #96, opened 2015-10-01
(CTH-390) Add PXP agent init scripts

geoffnichols:
This commit adds init scripts and service
configuration files for initialization of
the PXP agent service.

These are derived as wholly as possible from
the facilities currently used for Puppet in
the AIO puppet-agent.

### DIFF
--- a/ext/debian/pxp-agent.default
+++ b/ext/debian/pxp-agent.default
@@ -1,0 +1,5 @@
+# Defaults for pxp-agent - sourced by /etc/init.d/pxp-agent
+
+# Startup options
+#DAEMON_OPTS=""
+#PXP_AGENT_EXTRA_OPTS=--loglevel=debug
--- a/ext/debian/pxp-agent.init
+++ b/ext/debian/pxp-agent.init
@@ -1,0 +1,113 @@
+#! /bin/sh
+### BEGIN INIT INFO
+# Provides:          pxp-agent
+# Required-Start:    $network $named $remote_fs $syslog
+# Required-Stop:     $network $named $remote_fs $syslog
+# Should-Start:      pxp-agent
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+### END INIT INFO
+
+PATH=/opt/puppetlabs/puppet/bin:/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+DAEMON=/opt/puppetlabs/puppet/bin/pxp-agent
+DAEMON_OPTS="--daemonize "
+PROCNAME="pxp-agent"
+DESC="PXP agent"
+PIDFILE="/var/run/puppetlabs/${PROCNAME}.pid"
+
+test -x $DAEMON || exit 0
+
+[ -r /etc/default/pxp-agent ] && . /etc/default/pxp-agent
+
+. /lib/lsb/init-functions
+
+reload_pxp_agent() {
+    start-stop-daemon --stop --quiet --signal HUP --pidfile $PIDFILE --name $PROCNAME
+}
+
+start_pxp_agent() {
+    start-stop-daemon --start --quiet --pidfile $PIDFILE --startas $DAEMON $DAEMON_OPTS $PXP_AGENT_EXTRA_OPTS
+}
+
+stop_pxp_agent() {
+    start-stop-daemon --stop --retry TERM/10/KILL/5 --quiet --oknodo --pidfile $PIDFILE --name $PROCNAME
+    rm -f "$PIDFILE"
+}
+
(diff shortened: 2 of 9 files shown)


ext/osx/pxp-agent.plist
```diff
@@ -0,0 +1,27 @@
...
+<dict>
+        <key>EnvironmentVariables</key>
+        <dict>
+                <key>PATH</key>
+                <string>/opt/puppetlabs/puppet/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+                <key>LANG</key>
+                <string>en_US.UTF-8</string>
+        </dict>
+        <key>Label</key>
+        <string>pxp-agent</string>
```
heathseals: @stahnma @kylog we decided that having it fully qualified was the correct way to do it ( and would require managing it with the full name ), but there would be additional work elsewhere.

kylog: @heathseals does that mean:
a) we're going to leave it as-is in this PR, i.e. not fully qualified
OR
b) we're going to change it to be fully qualified, and the additional work is ticketed
OR
something else?

ext/debian/pxp-agent.init
```diff
@@ -0,0 +1,113 @@
...
+### END INIT INFO
+
+PATH=/opt/puppetlabs/puppet/bin:/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+DAEMON=/opt/puppetlabs/puppet/bin/pxp-agent
+DAEMON_OPTS="--daemonize "
+PROCNAME="pxp-agent"
+DESC="PXP agent"
+PIDFILE="/var/run/puppetlabs/${PROCNAME}.pid"
+
+test -x $DAEMON || exit 0
```
haus: I think this should exit 5 according to lsb.

ext/debian/pxp-agent.init
```diff
@@ -0,0 +1,113 @@
...
+    start-stop-daemon --stop --quiet --signal HUP --pidfile $PIDFILE --name $PROCNAME
+}
+
+start_pxp_agent() {
+    start-stop-daemon --start --quiet --pidfile $PIDFILE --startas $DAEMON $DAEMON_OPTS $PXP_AGENT_EXTRA_OPTS
+}
+
+stop_pxp_agent() {
+    start-stop-daemon --stop --retry TERM/10/KILL/5 --quiet --oknodo --pidfile $PIDFILE --name $PROCNAME
+    rm -f "$PIDFILE"
```
haus: do we always want the pidfile blown away or only on success?

ext/debian/pxp-agent.init
```diff
@@ -0,0 +1,113 @@
...
+}
+
+restart_pxp_agent() {
+    log_begin_msg "Restarting $DESC"
+    stop_pxp_agent
+    start_pxp_agent
+    log_end_msg $?
```
haus: We don't ever check to see if the stop was successful here. Do we care? Should the restart stop if the stop failed?

haus:
@geoffnichols the variables used and cases vary greatly for the svsv scripts (deb, el, sles). These would be easier to follow if they used the same variables (DAEMON vs pxpagentd for example). Also the sysconfig/default files should probably have the same options exposed, which currently isn't the case.

ext/redhat/pxp-agent.init
```diff
@@ -0,0 +1,145 @@
...
+export PATH
+
+[ -f /etc/sysconfig/pxp-agent ] && . /etc/sysconfig/pxp-agent
+lockfile=/var/lock/subsys/pxp-agent
+piddir=/var/run/puppetlabs
+pidfile=${piddir}/pxp-agent.pid
+pxp_agentd=/opt/puppetlabs/puppet/bin/pxp-agent
+pid=$(cat $pidfile 2> /dev/null)
+RETVAL=0
+
```
haus: Any reason we aren't checkng to see if pxp_agentd is executable here but we did in the debian? Same question for sles.

ext/debian/pxp-agent.default
```diff
@@ -0,0 +1,5 @@
+# Defaults for pxp-agent - sourced by /etc/init.d/pxp-agent
+
+# Startup options
+#DAEMON_OPTS=""
```
haus: DAEMON_OPTS is called PXP_AGENT_OPTS in the el init script. We should probably pick one.

geoffnichols:
Since the puppetlabs/pxp-agent repo was made public after this PR was opened, it looks like I need to close this PR and open a new one (with a new fork of puppetlabs/pxp-agent). 

Pleas see https://github.com/puppetlabs/pxp-agent/pull/112.

PR #104, opened 2015-10-02
(CTH-387) Update default paths and configuration options

parisiale:
- Updating paths as per Puppet Specifications PR#50
- Using logdir instead of logfile
- New console-logger option
- Replacing boost filesystem's create_directory calls with
  create_directories, to ensure we create entire dir paths
- Processing spool directory option value as other path options (i.e. we
  don't ensure a trailing slash anymore)

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -168,9 +168,11 @@ void Configuration::validateAndNormalizeConfiguration() {
 
         if (!fs::exists(spool_dir)) {
             LOG_INFO("Creating spool directory '%1%'", spool_dir);
-            if (!fs::create_directories(spool_dir)) {
-                throw Configuration::Error { "failed to create the results "
-                                             "directory '" + spool_dir + "'" };
+            try {
+                fs::create_directories(spool_dir);
+            } catch (const fs::filesystem_error& e) {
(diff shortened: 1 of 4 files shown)


lib/src/configuration.cc
```diff
@@ -30,41 +30,40 @@ namespace lth_log = leatherman::logging;
...
+    static const std::string DEFAULT_LOG_DIR { "/var/log/puppetlabs/pxp-agent" };
 #endif
 
-static const std::string DEFAULT_MODULES_DIR = (DEFAULT_SHARE_DIR / "modules").string();
-static const std::string DEFAULT_MODULES_CONF_DIR = (DEFAULT_CONF_DIR / "modules.d").string();
-static const std::string DEFAULT_CONFIG_FILE = (DEFAULT_CONF_DIR / "pxp-agent.cfg").string();
+
+// TODO(ale): temporary fix; update this for CTH-373
+static const std::string DEFAULT_MODULES_DIR {
+    (fs::current_path() / "modules").string() };
```
MikaelSmith: We've done that through CMAKE defines in packaging for Facter (https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/facter.rb#L157).

PR #105, opened 2015-10-02
(maint) Run ssh-agent in Rake so ssh keys are available to beaker

james-stocks:


er0ck:
most of our tests are internal, eg jenkins -> vmpooler.  so any risk of
using ssh-agent (someone *on* that system has keys to many of our other
systems) is extremely low (if someone we don't want is on that system we
have bigger problems).

but maybe this is a bigger risk if testing on a SUT outside of puppet?
in this case maybe using the clone-on-test-runner and scp to the SUT is a
better alternative (if slower).

BUT, there is also some security risk of reverse tunneling using the SSH
session from the SUT back to the test-runner.
basically, there is always risk when using SUTs outside of puppet.

i'm not thinking of any decent alternatives to the two above, right now.

On Sun, Oct 4, 2015 at 8:45 AM, Ethan J. Brown <notifications@github.com>
wrote:

> @james-stocks <https://github.com/james-stocks> @er0ck
> <https://github.com/er0ck> @parisiale <https://github.com/parisiale>
> @joshcooper <https://github.com/joshcooper> are we sure using ssh-agent
> here is a good idea? http://rabexc.org/posts/pitfalls-of-ssh-agents
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/puppetlabs/pxp-agent/pull/105#issuecomment-145359968>.
>



-- 
Eric Thompson
eric.thompson@puppetlabs.com <erict@puppetlabs.com>
Senior Software Quality Assurance Engineer

PR #115, opened 2015-10-06
(PCP-30) Agent should be able to start unconfigured

ploubser:
Here we allow the agent to start in a unconfigured mode to help
facilitate acceptance testing.

When the key parameters like server, ca, cert and key are not set,
instead of starting normally the server will start a busy loop. At this
point the agent should still behave normally to signals.

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -14,6 +14,7 @@
 #include <horsewhisperer/horsewhisperer.h>
 
 #include <memory>
+#include <thread>
 
 namespace PXPAgent {
 
@@ -21,8 +22,6 @@ namespace HW = HorseWhisperer;
 namespace lth_file = leatherman::file_util;
 
 int startAgent(std::vector<std::string> arguments) {
-    const auto& agent_configuration =
-        Configuration::Instance().getAgentConfiguration();
 
 #ifndef _WIN32
     std::unique_ptr<Util::PIDFile> pidf_ptr;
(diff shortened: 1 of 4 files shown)


exe/main.cc
```diff
@@ -62,21 +74,17 @@ int main(int argc, char *argv[]) {
...
-    } catch(const Configuration::Error& e) {
-        std::cout << "A configuration error has occurred:\n  ";
-        err_msg = e.what();
-    }
-
-    if (!err_msg.empty()) {
-        std::cout << err_msg << "\nCannot start pxp-agent" << std::endl;
+        std::cout << e.what() << "\nCannot start pxp-agent'n";
         return 1;
+    } catch(const Configuration::Error& e) {
```
parisiale: You will start the dumb thread after for any error thrown in configuration.cc.
Wouldn't it be better to start it only if we miss the config file?
You could do that by throwing a specific error.

exe/main.cc
```diff
@@ -41,7 +40,20 @@ int startAgent(std::vector<std::string> arguments) {
     }
 #endif
 
+    if (!Configuration::Instance().isInitialized()) {
+        std::thread idle_thread { []() {
```
parisiale: Adding a comment here would be nice.

parisiale: I mean, about what we're doing with the dumb thread when Configuration is not initialized.

lib/src/configuration.cc
```diff
@@ -354,8 +358,7 @@ void Configuration::parseConfigFile() {
 
     // TODO(ale): update this for CTH-380
     if (!lth_file::file_readable(config_file_)) {
-        throw Configuration::Error { "config file '" + config_file_
```
parisiale: If you agree, you can throw a specific error.

lib/src/configuration.cc
```diff
@@ -354,8 +358,7 @@ void Configuration::parseConfigFile() {
 
     // TODO(ale): update this for CTH-380
```
parisiale: This can go now.

exe/main.cc
```diff
@@ -41,7 +40,20 @@ int startAgent(std::vector<std::string> arguments) {
...
+            LOG_INFO("pxp-agent started unconfigured. No connection can be established");
```
parisiale: Perhaps at WARNING?

PR #116, opened 2015-10-06
(PCP-49) Removing PIDFile lock functionality temporarily

parisiale:
Removing the file locking functionality based on flock(); temporary fix.


kylog:
I did a quick test build on osx as a sanity check. Merging but leaving the ticket open for the real fix.

PR #118, opened 2015-10-07
(PCP-49) Use fcntl-based file locking

parisiale:
In PIDFile, we now replace the flock() system call with fcntl() for
extending portability to Solaris. The resulting lock entity has
fundamentally different semantics.

Adding read / write locking calls, that can be blocking or not.
Adding lock status check calls.
Crating log file with PIDFile ctor.
Acquiring read lock in each spawned process; file locks are not
inherited anymore with fcntl.
Using SIGUSR1-based synchronization to mantain file lock accross fork()
calls.
Adding unit tests for new functions.

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -15,6 +15,7 @@
 
 #include <memory>
 #include <thread>
+#include <chrono>
 
 namespace PXPAgent {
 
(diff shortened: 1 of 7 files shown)


ploubser:
:+1:

parisiale:
@kylo, @joshcooper  ^^

PR #120, opened 2015-10-09
(PCP-45) Replace --daemonize with --foreground

parisiale:


exe/main.cc
```diff
@@ -27,7 +27,7 @@ int startAgent(std::vector<std::string> arguments) {
     std::unique_ptr<Util::PIDFile> pidf_ptr;
 
     try {
-        if (Configuration::Instance().get<bool>("daemonize")) {
+        if (!Configuration::Instance().get<bool>("no-daemonize")) {
```
ploubser: foreground maybe?

parisiale: yes

PR #123, opened 2015-10-12
(PCP-57) Use known paths for modules directory

parisiale:
For Windows and *nix we now use the paths indicated in Puppet
Specifications for the modules directory default path.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -42,12 +42,12 @@ namespace lth_log = leatherman::logging;
 
     static const fs::path DEFAULT_CONF_DIR { DATA_DIR / "etc" };
     const std::string DEFAULT_SPOOL_DIR { (DATA_DIR / "var" / "spool").string() };
-    static const std::string PID_DIR { (DATA_DIR / "var" / "run").string() };
+    const std::string PID_DIR { (DATA_DIR / "var" / "run").string() };
     static const std::string DEFAULT_LOG_DIR { (DATA_DIR / "var" / "log").string() };
 #else
     static const fs::path DEFAULT_CONF_DIR { "/etc/puppetlabs/pxp-agent" };
     const std::string DEFAULT_SPOOL_DIR { "/opt/puppetlabs/pxp-agent/spool" };
-    static const std::string PID_DIR { "/var/run/puppetlabs" };
+    const std::string PID_DIR { "/var/run/puppetlabs" };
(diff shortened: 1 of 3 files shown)


lib/src/configuration.cc
```diff
@@ -30,33 +30,34 @@ namespace lth_log = leatherman::logging;
...
                 (boost::format("failure getting Windows AppData directory: %1%")
                     % lth_w::system_error()).str() };
         }
-        return fs::path(szPath) / "PuppetLabs" / "pxp-agent";
-    }();
+        return fs::path { szPath };
+    }
+
+    static const fs::path DATA_DIR = getWinPath(CSIDL_COMMON_APPDATA)
+                                        / "PuppetLabs" / "pxp-agent";
```
MikaelSmith: This looks wrong, should have a space in it.

lib/src/configuration.cc
```diff
@@ -30,33 +30,34 @@ namespace lth_log = leatherman::logging;
...
+    static const fs::path BIN_DIR = getWinPath(CSIDL_PROGRAM_FILES)
+                                        / "Puppet Labs" / "Puppet" / "pxp-agent";
```
MikaelSmith: I think @Iristyle and I agree that it'd be better to use a relative path to the binary, as in http://stackoverflow.com/questions/2647429/c-windows-path-to-the-folder-where-the-executable-is-located.

Iristyle: I recall @kylog had some objections to relative paths, but I'm not sure what they were exactly.  On *nix the pathing structure is not configurable, and there are standard / well-documented / uniform paths - so I can understand why that's the better approach there... given I don't believe there's a difference between "default" and "actual"

However, in the Windows installer we have the ability to install applications anywhere, so we probably shouldn't use `CSIDL_PROGRAM_FILES` at all since that would only work with the defaults.

Anything under `CSIDL_COMMON_APPDATA` (aka `C:\ProgramData`) is not configurable (aside from at the OS level, where the path may live on a different volume), so those paths can remain fixed.

But anything in the spec at https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md that lives under `Program Files` should be referred to via relative path from the `.exe` given the general simplicity of that approach (i.e nothing hardcoded, nor any registry lookups required)

parisiale: Ok.
Will use a relative path for Windows and the known "/opt/..." path for *nix.

PR #125, opened 2015-10-12
(maint) Use boost instead of std::thread on Windows

MikaelSmith:
Transition to boost::thread from std::thread missed a spot.


Iristyle:
:+1:

ferventcoder:
Also :+1:

PR #133, opened 2015-10-13
(PCP-63) pxp-agent will always demonize when unconfigured

ploubser:
Here we change the when we start in unconfigured mode from "no config
file is specified, the default config file doesn't exist, drop all flags
and start daemonized" to "If options are not set that we require to
connect, start unconfigured". We will continue to fail in the case of
config errors that we cannot recover from, such as parsing invalid
flags, not being able to process a given config file due to bad values,
not being able to log and (in the future) not being able to write a
pidfile.

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -27,7 +27,8 @@ int startAgent(std::vector<std::string> arguments) {
     std::unique_ptr<Util::PIDFile> pidf_ptr;
 
     try {
-        if (!Configuration::Instance().get<bool>("foreground")) {
+        // Using HW because configuration may not be initialized at this point
+        if (!HW::GetFlag<bool>("foreground")) {
             // Store it for RAII
             // NB: pidf_ptr will be nullptr if already a daemon
             pidf_ptr = Util::daemonize();
@@ -79,7 +80,6 @@ int main(int argc, char *argv[]) {
     Configuration::Instance().setStartFunction(startAgent);
 
     HW::ParseResult parse_result;
-    std::string err_msg {};
 
     try {
         parse_result = Configuration::Instance().initialize(argc, argv);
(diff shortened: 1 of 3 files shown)


exe/main.cc
```diff
@@ -27,7 +27,7 @@ int startAgent(std::vector<std::string> arguments) {
     std::unique_ptr<Util::PIDFile> pidf_ptr;
 
     try {
-        if (!Configuration::Instance().get<bool>("foreground")) {
+        if (!HW::GetFlag<bool>("foreground")) {
```
parisiale: It would be nice to have the usual comment about using HW because Configuration may not be initialized.

exe/main.cc
```diff
@@ -88,16 +88,11 @@ int main(int argc, char *argv[]) {
...
+        std::cout << "pxp-agent will start unconfigured\n";
     } catch(const Configuration::Error& e) {
-        std::cout << "A configuration error has occurred:\n  ";
-        err_msg = e.what();
-    }
-
-    if (!err_msg.empty()) {
-        std::cout << err_msg << "\nCannot start pxp-agent" << std::endl;
+        std::cout << "A configuration error has occurred: " << e.what() << "\n";
+        std::cout << "pxp-agent cannot start\n";
```
parisiale: We don't use the err_msg variable anymore.

parisiale: Maybe using std::endl before returning to flush.

parisiale:
I suggest to update README, in the "Configuring the agent" section where we talk about "unconfigured mode".
Also, we have the server and cert options marked as "required"; perhaps we can specify that it's required to actually connect the agent to a broker.

ploubser:
Updated config to reflect how unconfigured mode works now.

PR #136, opened 2015-10-13
(PCP-69) Update service configs for daemonization by default

geoffnichols:
By default, pxp-agent now daemonizes if unconfigured, and runs as
a foreground process if invoked with --foreground.

This commit removes the unsupported --daemonize argument from the
pxp-agent SYSV init scripts and adds the --foreground argument for
the platforms where daemonization is not needed.


haus:
:+1:

PR #138, opened 2015-10-13
(PCP-67) Bump cmake requirement to 3.2.2, set CMAKE_INSTALL_RPATH

geoffnichols:
Prior to this changeset, execution of pxp-agent on OS X fails with ```dyld: Library not loaded:
libcpp-pcp-client.so```. To fix this, we borrow from Facter, bumping the cmake requirement to 3.2.2 and setting the CMAKE_INSTALL_RPATH based on the install prefix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,12 @@ if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
     set(CMAKE_INCLUDE_PATH ${CMAKE_INCLUDE_PATH} /opt/local/include)
 endif()
 
+# Set RPATH if not installing to a system library directory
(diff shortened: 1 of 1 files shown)


geoffnichols:
See https://github.com/puppetlabs/puppet-agent/pull/338 for related discussion.

geoffnichols:
I haven't tested this yet.

geoffnichols:
Seeing the following error when testing this:

```
# /opt/puppetlabs/puppet/bin/pxp-agent
A configuration error has occurred: server value must be defined
pxp-agent will start unconfigured
An unexpected code was returned when trying to parsecommand line arguments - 2115506936. Aborting
```

PR #151, opened 2015-10-15
(PCP-17) Ctrl-C handling and PID files on Windows

MikaelSmith:
Adds Ctrl-C and Shutdown handling, along with creating a PID file to
ensure only one pxp-agent is running at a time.


MikaelSmith:
This is terrible code, and I'd like to clean it up. Doing so will probably require changing how control flow works, or reproducing more of the Linux version.

However, it works.

MikaelSmith:
@e suggests [CreateMutex](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682411(v=vs.85).aspx) is simpler if we don't care about others checking for the existence of the PID file.

exe/main.cc
```diff
@@ -26,7 +23,17 @@ namespace HW = HorseWhisperer;
...
+    try {
+        Util::daemonize();
+    } catch (const std::exception& e) {
+        LOG_ERROR("Failed to daemonize: %1%", e.what());
+        return 1;
+    } catch (...) {
+        LOG_ERROR("Failed to daemonize");
+        return 1;
+    }
+#else
```
parisiale: Perhaps we can move the platform branch to Util::daemonize() and use a specific error for failures.

exe/main.cc
```diff
@@ -26,7 +23,17 @@ namespace HW = HorseWhisperer;
 namespace lth_file = leatherman::file_util;
 
 int startAgent(std::vector<std::string> arguments) {
-#ifndef _WIN32
+#ifdef _WIN32
+    try {
+        Util::daemonize();
```
parisiale: You're not considering the --foreground option.

exe/main.cc
```diff
@@ -76,6 +83,9 @@ int startAgent(std::vector<std::string> arguments) {
         LOG_ERROR("Unexpected error");
     }
 
+#ifdef _WIN32
+    Util::cleanup();
```
parisiale: Perhaps renaming to something with "daemon" would make the call more clear.

lib/inc/pxp-agent/util/windows/pid_file.hpp
```diff
@@ -0,0 +1,70 @@
...
+    PIDFile(const std::string& dir_path);
+
+    // Perform clean up unless already done.
+    ~PIDFile();
+
+    // Disable copies, and provide move semantics.
+    PIDFile(PIDFile const&) = delete;
+    PIDFile& operator=(PIDFile const&) = delete;
+    PIDFile(PIDFile &&) = default;
+    PIDFile& operator=(PIDFile &&) = default;
```
parisiale: We should add that to the POSIX side.

lib/inc/pxp-agent/util/windows/pid_file.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+    // Perform clean up; use logging if log is true
+    // logging is disabled in the destructor, as we can't guarantee
+    // it'll be destroyed before Boost.Log is shut down
```
parisiale: Can we try to log anyway?

parisiale: Anyway, this should be taken care of on the POSIX side as well.

parisiale:
I left minor comments.

Do we want to prevent multiple pxp-agent foreground instances on POSIX, as done in Windows (the {{--foreground}} option is not considered)? We don't do that currently. If so, the PID file creation and the daemonization() call should be two distinct steps in the main.cc start function.

On POSIX, we error (if {{--foreground}} is unflagged) when: 1) the PID reported in pxp-agent.pid is relative to a running process or 2) pxp-agent.pid is locked (read or write). In Windows, I suggest to go with the mutex solution suggested by Ethan and use a more specific and simple workflow. The important thing is that we prevent multiple pxp-agent services running at once with same configuration (broker and certs); in that case, a livelock would take place at the PXP messaging level.

exe/main.cc
MikaelSmith: Foreground should allow multiple instances to run?

On Windows it's not going to start a process in the background, as that's not really how services work.

lib/inc/pxp-agent/util/windows/pid_file.hpp
MikaelSmith: This only happens because I implemented cleanup differently. I'm using global static destruction on Windows, where I think you circumvent that on Linux. I'll double check.

MikaelSmith:
Since you can specify a different config-file on the command-line, it makes sense to allow running multiple foreground pxp-agent instances under some circumstances. The logic would be more complicated to ensure it's safe to do, so I'm tempted to leave it up to the user to be safe when running foreground.

parisiale:
I agree.

exe/main.cc
MikaelSmith: Not quite sure what you mean a specific error for failures.

exe/main.cc
parisiale: I know. I put in evidence here since I refer to it in my global comment.
The point is that we have a different behaviour in POSIX.

exe/main.cc
parisiale: Simply throw a PID::Error.

MikaelSmith: Part of the try is also catching Horsewhisperer exceptions.

Iristyle:
Yeah, sorry I didn't mention the mutex based solution earlier, which is the canonical kernel object solution for ensuring a single process on Windows.

I had assumed the pid file management code was leveraging something cross-platform to guarantee atomic operations when I stated Windows could share the same PID file management  When I saw @MikaelSmith had written a bunch of Windows specific code, it seemed to make more sense to do the Windows appropriate thing here.  That actually re-raises the question of managing PID files in the init system, since Windows has a different code path anyway (which I hadn't previously understood, as mentioned).

So as long as there isn't a need for other processes to use the file system to see if the `pxp-agent` service is running, then a mutex is the way to go.  On Windows, we should also be able to simply check the service status from other processes as well, given we only care about not daemonizing multiple times.

MikaelSmith:
I'd rather not rewrite the POSIX behavior for PCP-17. Have a lot of other work left to do.

MikaelSmith:
This works and I'm happy with it for now.

PR #152, opened 2015-10-16
(PCP-77) Don't create directories 

ploubser:


ploubser:
@parisiale Can you check for sanity?

@Iristyle Will this wreck us on Windows? IIRC we created the logdir for windows, since the MSI wasn't doing it. With us moving to using the event log, logdir no longer becomes relevant, right?

parisiale:
@ploubser sure

Also @ferventcoder ^^

lib/src/configuration.cc
```diff
@@ -193,15 +193,11 @@ void Configuration::validateAndNormalizeConfiguration() {
         spool_dir = lth_file::tilde_expand(spool_dir);
 
         if (!fs::exists(spool_dir)) {
+            throw Configuration::Error { "spool-dir does not exists" };
             LOG_INFO("Creating spool directory '%1%'", spool_dir);
```
parisiale: Missed the log.

lib/src/util/posix/pid_file.cc
```diff
@@ -36,12 +36,9 @@ PIDFile::PIDFile(const std::string& dir_path_)
...
         }
     } else {
-        LOG_INFO("Creating PID directory '%1%'", dir_path);
-        try {
-            fs::create_directories(dir_path);
-        } catch (const fs::filesystem_error& e) {
-            throw PIDFile::Error { e.what() };
-        }
+        std::string err_msg { "'" + dir_path + "' doesn't exist. Cannot create PID file" };
+        LOG_ERROR(err_msg);
```
parisiale: Do we need to log here? I think it will be logged again in main.exe.

parisiale:
:+1: for me, just minor stuff

Iristyle:
@ploubser we will create the default log dirs in the MSI per the spec.   As I mentioned earlier, I don't think it's a good idea to assume long term that only 1 active logging method is supported - they shouldn't be treated as mutually exclusive options IMHO.  Short-term that's probably fine though.

Many .NET Windows based applications use log4net or NLog, where the log sinks can be configured at run-time, and thus end users can choose event log + Windows debug streams + console + files, etc, etc.  I think a flexible runtime selection is the best when it comes to logging.

The ticket doesn't enumerate the concerns with pxp-agent creating directories, but I assume that they're security motivated?

And just to clarify - the path moving forward is that if a user overrides a disk based setting (i.e. log), they must always have created the directory manually prior to using it?

ploubser:
@Iristyle Correct. However we end up implementing logging, when a user wants to log to a file the target directory must be there.

PR #153, opened 2015-10-16
(PCP-83) Fix unit tests

ploubser:
Here we fix the status action's unit tests to be able to pass when run
as a non-root user or when the default spool-dir directories don't
exist.

### DIFF
--- a/lib/src/modules/status.cc
+++ b/lib/src/modules/status.cc
@@ -1,5 +1,4 @@
 #include <pxp-agent/modules/status.hpp>
-#include <pxp-agent/configuration.hpp>
 
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/operations.hpp>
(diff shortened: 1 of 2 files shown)


lib/tests/unit/modules/status_test.cc
```diff
@@ -95,7 +101,7 @@ TEST_CASE("Modules::Status::executeAction", "[modules]") {
                     0 };
             ActionRequest request { RequestType::Blocking, other_chunks };
 
-            boost::filesystem::path dest { DEFAULT_SPOOL_DIR };
+            boost::filesystem::path dest { SPOOL_DIR };
```
parisiale: What about calling remove_directories on SPOOL_DIR once we're done?

lib/src/modules/status.cc
```diff
@@ -38,7 +40,10 @@ Status::Status() {
 ActionOutcome Status::callAction(const ActionRequest& request) {
     lth_jc::JsonContainer results {};
     auto t_id = request.params().get<std::string>("transaction_id");
-    fs::path results_path { Configuration::Instance().get<std::string>("spool-dir") };
+    // We use HW instead of Configuration because we may be called from tests
+    // where Configuration is still unconfigured. We cannot get here during
+    // normal flow unconfigured, so using HW should be safe.
```
parisiale: Calling HW should be safe anyway, no?

lib/tests/unit/modules/status_test.cc
ploubser: Line 155. It could be clearer, but I don't want to touch too many loc here.

lib/src/modules/status.cc
ploubser: Good point, I'll remove the comment.

PR #148, opened 2015-10-15
(PCP-6) Reopen log file after SIGUSR2 on *nix

parisiale:
Registering a handler for SIGUSR2 that will reopen the file associated
with the log file stream (new class var of Configuration), in case we
log on file on *nix platforms.

Adding new platform specific section on configuration.hpp and related
implementation files in src/configuration/posix  and /windows.
Adding reopenLogFile() method to Configuration.

PR #155, opened 2015-10-16
(PCP-92) Add logrotate configuration

geoffnichols:
This commit adds a logrotate config for the pxp-agent.

### DIFF
--- a/ext/pxp-agent.logrotate
+++ b/ext/pxp-agent.logrotate
@@ -1,0 +1,11 @@
+/var/log/puppetlabs/pxp-agent/*.log {
+    daily
+    missingok
+    rotate 30
+    compress
+    notifempty
+    sharedscripts
+    postrotate
+        [ -s /var/run/puppetlabs/pxp-agent.pid ] && kill -USR2 `cat /var/run/puppetlabs/pxp-agent.pid`
+    endscript
+}


ext/pxp-agent.logrotate
```diff
@@ -0,0 +1,8 @@
+/var/log/puppetlabs/pxp-agent/pxp-agent.log {
```
haus: Is this the only log generated by pxp-agent?

stahnma: Seems like doing *.log here would be more future proof.

haus:
Who owns the pxp-agent directory? root?

haus:
Does logrotate compress by default? If not, that's probably worth turning on.

mhaskel:
Don't you have to specify one of time or size?

ext/pxp-agent.logrotate
```diff
@@ -0,0 +1,11 @@
+/var/log/puppetlabs/pxp-agent/*.log {
+    daily
+    missingok
+    rotate 30
+    compress
```
geoffnichols: It looks like logrotate compresses (with gzip) by default, but I have added that explicitly here.

ext/pxp-agent.logrotate
```diff
@@ -0,0 +1,11 @@
+/var/log/puppetlabs/pxp-agent/*.log {
+    daily
```
geoffnichols: I'm not sure if daily or weekly will make the most sense here.

ext/pxp-agent.logrotate
```diff
@@ -0,0 +1,11 @@
...
+    missingok
+    rotate 30
```
geoffnichols: Is this overkill? Maybe 10 would be sufficient?

ext/pxp-agent.logrotate
shrug: Probably depends on what kind of info is in the log? Daily & keep 30 works for me for now? Or weekly, keep 4-ish?

geoffnichols:
@haus `/var/log/puppetlabs/pxp-agent/` is currently owned by `root:root`.

geoffnichols:
Relates to https://github.com/puppetlabs/pxp-agent/pull/148.

mhaskel:
:+1:

PR #160, opened 2015-10-20
(PCP-23) Renaming configuration options

parisiale:


lib/inc/pxp-agent/configuration.hpp
```diff
@@ -35,7 +35,7 @@ struct EntryBase {
     // Config option name
     // HERE: must match one of the flag names and config file option
     std::string name;
-    // CLI option aliases (e.g. --server --serv -s)
+    // CLI option aliases (e.g. --broker --bro -b)
```
ploubser: I think we should get rid of the alias. --broker isn't very long and nothing else is shortened.

lib/src/configuration.cc
```diff
@@ -260,38 +262,38 @@ void Configuration::defineDefaultValues() {
...
 
     defaults_.insert(
-        Option { "server",
+        Option { "broker-ws-uri",
                  Base_ptr { new Entry<std::string>(
-                    "server",
-                    "s",
-                    "PCP server URL",
+                    "broker-ws-uri",
+                    "b",
```
ploubser: See previous comment about shortening.

lib/inc/pxp-agent/configuration.hpp
parisiale: Ok.

PR #162, opened 2015-10-20
(PCP-78) New --pidfile option

parisiale:
Here we allow specifying the path of the PID file on *nix.
Note that on Windows we don't write the PID to file; we ensure that only
one instance of pxp-agent is running at time by using a mutex.

Modifying the PIDFile interface.
Checking the pidfile path.
Updating unit tests.

### DIFF
--- a/lib/src/util/posix/pid_file.cc
+++ b/lib/src/util/posix/pid_file.cc
@@ -10,7 +10,6 @@
 
 #include <cstdio>           // remove()
 #include <cassert>
-
 #include <sys/file.h>       // open()
 #include <sys/stat.h>
 #include <fcntl.h>          // fcntl(), open() flags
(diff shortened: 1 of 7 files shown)


lib/src/util/posix/pid_file.cc
```diff
@@ -23,13 +22,16 @@ namespace Util {
...
-
-PIDFile::PIDFile(const std::string& dir_path_)
-        : dir_path { dir_path_ },
-          file_path { dir_path + "/" + FILENAME },
+PIDFile::PIDFile(const std::string& file_path_)
+        : dir_path { fs::path(file_path_).parent_path().string() },
+          file_path { file_path_ },
           pidfile_fd {},
           cleanup_when_done { false } {
+    if (fs::exists(file_path) && !(fs::is_regular_file(file_path))) {
```
ploubser: Aren't we in trouble if the PID file already exists? Should this be a check for, "if the pidfile exists, terminate with an error"?

parisiale: No.
We don't rely on the PID file existence. That would be very fragile.
We 1) ensure that the written PID is not associated with a running process and 2) use a writer lock.

ploubser: Cool, I dig.

PR #164, opened 2015-10-20
(WIP) Run Puppet with --onetime and --no-daemonize

Iristyle:
- If an agent run is not triggered with --onetime, it might not exit
   if, for instance, the master hostname is not resolvable or the master
   is not reachable.

   Ensure that the process exits right away by running puppet with
   --onetime

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -169,6 +169,7 @@ def run(params_and_config)
   end
 
   params = params_and_config["params"]
+  params.merge!({'flags' => ['--onetime', '--no-daemonize']})
   config = params_and_config["config"]
 
   if config["puppet_bin"].nil? || config["puppet_bin"].empty?


Iristyle:
This was removed as part of https://tickets.puppetlabs.com/browse/PCP-86 ... but not sure that was the appropriate thing to do.

@jonathannewman can you comment on why `--onetime` is universally undesirable?

jonathannewman:
@Iristyle because we don't know what the future holds...  It is better to assume NO flags than to assume any.  Fewer moving parts.

nicklewis:
I think this flag (potentially along with `--no-daemonize`) is *always* required in order for the module to function correctly. I'm in favor of explicit behavior, but I think excluding these flags just subverts what the module is meant to do, which is to trigger a puppet run and report on its results.

If we later want behavior that requires omitting the default flags, we could create a new action (one whose semantics are explicitly *not* "trigger a puppet run and report on its results").

Iristyle:
Agreed on `--no-daemonize` as well @nicklewis.

PR updated accordingly.

kylog:
I'm all in favor of getting things right now and getting consensus on what "right" means. 

However, can we:
a) discuss this on the thread where the decision was originally made: "assumed flags in pxp-module-puppet"?
b) reopen https://tickets.puppetlabs.com/browse/PCP-86 and reference it in the commit message here? (this changes the pxp-module-puppet contract and so doesn't feel like "maint" to me)

Thanks!

Iristyle:
Yup, `(maint)` was a placeholder until I knew which ticket this was being attached to (i.e. re-open existing or create new).  I posted some proposed changes here for the sake of discussion - should have used `WIP`.

ploubser:
The discussed flags as well as --verbose have been added in https://github.com/puppetlabs/pxp-agent/pull/182. Closing this one. Thanks @Iristyle

PR #165, opened 2015-10-20
(PCP-79) Replace --logdir with --logfile

parisiale:
Removing --logdir option and --console-logger flag.
Introducing the new --logfile option.
README improvements.

Logging to stdout can be specified by using "--logfile -".

Changing default logging settings to continue logging on file by
default, but in case --foreground is flagged we don't error;
instead, we set the log level to none.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,10 +119,7 @@ console by using an hyphen instead of a file path: `--logfile -`.
 
 By default, log messages are written to the stdout stream.
 
-You can specify a log file with `--logfile` option. In that case, no
-file rotation will take place.
-
-Note that it is not possible to have log messages displayed only on console.
+You can specify a log file with `--logfile` option.
 
 ## Log level
 
--- a/README.md
+++ b/README.md
@@ -105,23 +105,14 @@ During its execution, the daemon PID will be stored in:
 
 ### Logging
 
-## *nix
-
-By default, log messages will be written to
-*/var/log/puppetlabs/pxp-agent/pxp-agent.log*.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -95,11 +105,26 @@ During its execution, the daemon PID will be stored in:
...
+
+## Windows
 
-You can specify a different directory with the `--logdir` flag.
+By default, log messages are written to the stdout stream.
+
+You can specify a log file with `--logfile` option. In that case, no
+file rotation will take place.
+
+Note that it is not possible to have log messages displayed only on console.
```
ploubser: What does this mean? If we default to logging to stdout, why won't it write to the console?

parisiale: That's wrong. I forgot to remove it after copying from https://gist.github.com/parisiale/e9ba57b2062cb5eaeed0, which was meant for PCP-100 discussion.

Sorry for that.

lib/src/configuration.cc
```diff
@@ -517,32 +514,26 @@ static void validateLogDirPath(fs::path logdir_path) {
             throw Configuration::Error { "log directory is not a directory" };
         }
     } else {
-        throw Configuration::Error { "--logdir '" + logdir_path.string() +
-                                     "' doesn't exist" };
+        throw Configuration::Error { "the log directory '" + logdir_path.string()
```
ploubser: Can we change this to say something like, "Cannot write $logfile: No such file or directory $logdir"? The term log directory doesn't really have that much context on its own when it is not exposed.

PR #169, opened 2015-10-21
(PCP-105) Fix unit tests

parisiale:
- Creating and removing directories as requested for each unit test, as
   after PCP-77 pxp-agent does not create directories anymore;
 - ensuring that Configuration is initialized for each unit test.

### DIFF
--- a/lib/tests/unit/configuration_test.cc
+++ b/lib/tests/unit/configuration_test.cc
@@ -1,5 +1,4 @@
 #include "certs.hpp"
-
 #include "root_path.hpp"
 
 #include <pxp-agent/configuration.hpp>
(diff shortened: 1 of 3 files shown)


lib/tests/unit/configuration_test.cc
```diff
@@ -9,56 +7,62 @@
...
+    "--ssl-cert", CERT.data(),
+    "--ssl-key", KEY.data(),
+    "--modules-dir", MODULES_DIR.data(),
+    "--spool-dir", SPOOL_DIR.data(),
+    "--foreground=true" };
+static const int ARGC = 16;
 
 static void configureTest() {
+    HW::Reset();
+    if (!fs::exists(SPOOL_DIR) && !fs::create_directories(SPOOL_DIR)) {
```
parisiale: Ensure that the testing spool directory is created, since we don't do that anymore.

lib/tests/unit/modules/status_test.cc
```diff
@@ -40,7 +41,32 @@ boost::format STATUS_FORMAT {
...
+    "--ssl-key", TEST_KEY_PATH.string().data() };
+static const int ARGC = 13;
+
+static void configureTest() {
+    HW::Reset();
+    if (!fs::exists(SPOOL_DIR) && !fs::create_directories(SPOOL_DIR)) {
+        FAIL("Failed to create the results directory");
+    }
+
+    Configuration::Instance().initialize(ARGC, const_cast<char**>(ARGV), false);
```
parisiale: Ensure Configuration is initialized with the options requested by the test..

lib/tests/unit/modules/status_test.cc
```diff
@@ -152,26 +177,22 @@ TEST_CASE("Modules::Status::executeAction", "[modules]") {
...
-            testResultFiles(false, result_path);
+            auto results = (TEST_RESOURCES_PATH / "delayed_result_failure").string();
+            testResultFiles(false, results);
         }
     }
 
-    if (!boost::filesystem::exists(SPOOL_DIR)) {
-        boost::filesystem::remove_all(SPOOL_DIR);
-    }
+    resetTest();
```
parisiale: Remove the testing spool directory.

parisiale:
Each test file has been executed in isolation, to exclude order-dependent failures.

PR #171, opened 2015-10-22
(PCP-92) Add systemd-specific logrotate config

geoffnichols:
Since the pxp-agent does not write a pid file
unless daemonized, we need to use "native"
systemd facilities to signal pxp-agent.

This commit adds a systemd-specific logrotate
config that uses a systemctl facility to do
postrotate signalling of pxp-agent.


ploubser:
:+1:

PR #179, opened 2015-10-22
(maint) Add AppVeyor and Travis CI

MikaelSmith:


Iristyle:
Looks like we need a rebase @MikaelSmith 

This was ready to go otherwise, right?

MikaelSmith:
It was, I'll rebase.

MikaelSmith:
Added the modules rspec run, but I don't have the same control over Ruby versions. rvm is only available in a Ruby build VM.

Iristyle:
Hmm - that's a bummer wrt Ruby versions.  It looks like they have `1.9.3-p551` installed by default in Travis - but we'll only ever run the module / module tests against `2.1.x+`.

This probably isn't worth spending a ton of time on, but it looks like you might still be able to switch Ruby versions by hand - http://rubies.travis-ci.org/usage

Precompiled / downloadable Ruby packages are available at http://rubies.travis-ci.org/

MikaelSmith:
I'll give it a try.

Iristyle:
"Disallowing packages: ruby-2.1.7" with your approach.

What I meant was calling `install-ruby` - which I think Travis has installed (source at https://github.com/postmodern/ruby-install)

That usage page had proposed that Travis can use `ruby-install --binary -M https://rubies.travis-ci.org/ ruby 2.1.1` for instance, but that PR at https://github.com/postmodern/ruby-install/pull/138 wasn't ever merged.

MikaelSmith:
Ah yeah, that looks like it might not work without sudo yet.

MikaelSmith:
Dropped the last commit, this is probably ok.

PR #180, opened 2015-10-23
(PCP-122) Don't set PATH in pxp-agent service scripts

geoffnichols:
Prior to this commit, some of the pxp-agent service
scripts were setting PATH. This appears to be
unneccesary. Additionally, the specific PATH that
was being set could lead to unexpected use of the
puppet-agent's "private" utilities, such as gem.

This commit removes PATH where it was set in the
pxp-agent service scripts.

### DIFF
--- a/ext/debian/pxp-agent.init
+++ b/ext/debian/pxp-agent.init
@@ -8,8 +8,6 @@
 # Default-Stop:      0 1 6
 ### END INIT INFO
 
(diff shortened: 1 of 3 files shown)


kylog:
:+1: 

Not sure what's up with Travis? I'll give it a retry in case it was something in travis/github infra

kylog:
Oh I think travis was turned on for #179, but doesn't actually work yet. Let's ignore travis for this PR.

PR #186, opened 2015-10-23
(PCP-117) Fix racy checks in running? and disabled?

ploubser:
In the past it was possible for a run to start between checking if there
was a run and running.

Here we no longer check if the lock files exist and instead kick a run
regardless and inspect the output to determine if a run was already
underway or if puppet was disabled. We still report the same message
back to the consumer.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -45,27 +45,19 @@ def check_config_print(cli_arg, config)
   return process_output.to_s.chomp
 end
 
-def running?(config)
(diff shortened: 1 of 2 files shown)


kylog:
I'm not a fan of scraping stdout's output. That's really not a part of the documented puppet contract and seems quite likely to expose us to surprising breakage when someone rewords a message.

ploubser:
@kylog neither am I, for all the same reasons. PCP-117 does seem potentially terrible. Any suggestions?

PR #188, opened 2015-10-26
(PCP-119) Add machine readable error_type to response

ploubser:
Here we add a new field to the response, error_type. This field is
intended to be machine readable so that consumers of the response
message can easier determine how to react to failures.

We are also able to refactor #get_result_from_report by adding
a #make_error_result function. This simplifies it's calling convention
by removing the need for the optional error string and fail_hard
boolean. Calls that do not require the last_run_report to be parsed can
now directly call #make_error_result.

### DIFF
--- a/modules/spec/unit/modules/puppet_spec.rb
+++ b/modules/spec/unit/modules/puppet_spec.rb
@@ -12,6 +12,19 @@ describe "pxp-module-puppet" do
     {"puppet_bin" => "puppet"}
   }
 
+  describe "last_run_result" do
+    it "returns the basic structure with exitcode set" do
+      expect(last_run_result(42)).to be == {"kind" => "unknown",
+                                            "time" => "unknown",
+                                            "transaction_uuid" => "unknown",
+                                            "environment" => "unknown",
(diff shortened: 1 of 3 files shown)


modules/spec/unit/modules/puppet_spec.rb
```diff
@@ -170,6 +201,7 @@
            "transaction_uuid" => "ac59acbe-6a0f-49c9-8ece-f781a689fda9",
            "environment"      => "production",
            "status"           => "changed",
+           "error_type"       => "",
```
richardc: Shouldn't this be an "All clear" error_type (if it should be here at all)

ploubser: I would prefer it if these fields weren't present on a successful run. This could however make it harder for NUI and App Mgmt to integrate these changes. I suggest we keep the behavior as it for now, run the change past them and then make "error" and "error_type" optional in another commit.

ploubser:
@nicklewis @jonathannewman FYI, this should make it simpler to determine what went wrong.

PR #189, opened 2015-10-26
(maint) Acceptance suite cleanup

Iristyle:


acceptance/tests/pxp_agent_version.rb
```diff
@@ -1,12 +1,17 @@
 test_name 'C93805 - pxp-agent - Versioning test'
 
-agent1 = agents[0]
+def pxp_version_cmd(platform)
+  if platform.start_with?('windows')
+    # simple hack to ensure OpenSSL is found w/out modifying PATH
+    "cd '/cygdrive/c/Program\\ Files/Puppet\\ Labs/Puppet/pxp-agent/bin' && ../../../pxp-agent/bin/pxp-agent --version"
```
Iristyle: Something is slightly off here

acceptance/setup/packages/010_Install.rb
```diff
@@ -32,15 +41,20 @@
...
+  ],
+}
+install_packages_on(master, MASTER_PACKAGES, :check_if_exists => true)
+
 step 'Install MSIs on any Windows agents'
 agents.each do |agent|
   if agent.platform.start_with?('windows')
     logger.info "Installing Puppet agent msi #{sha} on #{agent}"
     install_puppet_agent_dev_repo_on(agent, :version => sha)
-    logger.info 'Prevent Puppet Service from Running'
```
james-stocks: Preventing Puppet automatically running every 30 minutes is standard for test system setups. It may confuse people later if our setup does not stop the Puppet service.

acceptance/tests/pxp_agent_version.rb
```diff
@@ -1,12 +1,17 @@
...
+end
 
 step 'cd into pxp-agent bin folder and check the version'
-pxp_agent_folder = ''
-agent1.platform.start_with?('windows') ?
-  pxp_agent_folder = "/cygdrive/c/Program\\ Files/Puppet\\ Labs/Puppet/pxp-agent/bin" :
-  pxp_agent_folder = "/opt/puppetlabs/puppet/bin"
-on(agent1, "cd #{pxp_agent_folder} && ./pxp-agent --version") do |result|
-  assert(/[0-9\.]*/ =~ result.stdout, "Version number should be numbers and periods but was \"#{result.stdout.to_s}\"")
+agents.each do |agent|
```
james-stocks: There is no value in repeating the test command on every agent; it was deliberate that only the first agent was tested

acceptance/setup/packages/010_Install.rb
Iristyle: This is unnecessary, as Beaker automatically sets a startup type of `Manual` by default - see https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/windows_utils.rb#L111

If you look at the Beaker logs, you will also see something like this:

```
localhost $ scp /var/folders/1b/rzv96bhd5h9bwtdnz7g0xmlm0000gn/T/install-puppet-msi-2015-10-23_18.17.28.bat20151023-36854-1rusm7n win2012r2-64-1:C:\Windows\TEMP\install-puppet-msi-2015-10-23_18.17.28.bat {:ignore => }

pj6odfexaonohan.delivery.puppetlabs.net (win2012r2-64-1) 18:17:28$ cmd.exe /c "C:\Windows\TEMP\install-puppet-msi-2015-10-23_18.17.28.bat"

  C:\cygwin64\home\Administrator>start /w msiexec.exe /i "C:\PROGRA~3\puppet-agent-x64.msi" /qn /L*V C:\Windows\TEMP\install-puppet-2015-10-23_18.17.28.log PUPPET_AGENT_STARTUP_MODE=Manual

  C:\cygwin64\home\Administrator>exit /B 0

pj6odfexaonohan.delivery.puppetlabs.net (win2012r2-64-1) executed in 28.92 seconds

pj6odfexaonohan.delivery.puppetlabs.net (win2012r2-64-1) 18:17:57$ cmd.exe /c sc query puppet || sc query pe-puppet

  SERVICE_NAME: puppet
          TYPE               : 10  WIN32_OWN_PROCESS
          STATE              : 1  STOPPED
          WIN32_EXIT_CODE    : 1077  (0x435)
          SERVICE_EXIT_CODE  : 0  (0x0)
          CHECKPOINT         : 0x0
          WAIT_HINT          : 0x0

pj6odfexaonohan.delivery.puppetlabs.net (win2012r2-64-1) executed in 0.11 seconds
```

acceptance/tests/pxp_agent_version.rb
Iristyle: Until you add more agents to the config, which is something I would expect we would be adding shortly.

IMHO, it's a bad idea to hardcode the single agent assumption, when it's so easy not to.

acceptance/tests/pxp_agent_version.rb
Iristyle: Used the wrong dir :facepunch:

Iristyle:
@james-stocks this is ready to go AFAIK

Iristyle:
@james-stocks though you might want to have a look at the concessions made as part of 37b0370 ... that commit might be worth throwing out given it requires specifying `PC1`

MikaelSmith:
@Iristyle can you rebase to fix Travis?

Iristyle:
updated @MikaelSmith

joshcooper:
I'm assuming this is not the case, but if this is required for puppet-agent 1.3.0, please retarget for stable.

acceptance/tests/pxp_agent_version.rb
james-stocks: The test is logically passed if it works on one agent. Repeating on multiple agents is redundant. The current code is assuming there will be multiple agents in future - this test does NOT need to execute on every agent once we have multiple agents

acceptance/setup/packages/010_Install.rb
james-stocks: Ah. I copied the code from another project; on the assumption that this was the code that was needed to stop Puppet automatically running. Thanks for explaining

richardc:
Is this PR still needed?  It's currently conflicting.

acceptance/tests/pxp_agent_version.rb
james-stocks: On second thought; I agree with you here. However, all of the tests should be updated to make use of every agent. I've raised PCP-178 to update all the tests

james-stocks:
The current acceptance code, and the PR https://github.com/puppetlabs/pxp-agent/pull/259 , supersede most of this PR.
The change in this PR to have a test run on every agent needs applied to all the test cases - have raised separate ticket PCP-178 to handle this

PR #190, opened 2015-10-26
(PCP-95) Shell-escape command line options

parisiale:
With this commit we use Shellwords.escape on environment entries and
command line options.

Passing environment entries as a hash via :custom_environment to the
Puppet::Util::Execution.execute function.

Updating spec tests.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -3,6 +3,7 @@
 require 'json'
 require 'yaml'
 require 'puppet'
+require 'shellwords'
 
 module Errors
   InvalidJson = "invalid_json"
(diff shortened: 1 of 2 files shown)


ploubser:
@Iristyle Is this safe and sensible on windows? Shellwords does state that it escapes for the Bourne shell and I haven't had time to test this just yet so I'm fishing for a sanity check :)

Iristyle:
@ploubser Google says Shellwords + Windows = no bueno.

http://stackoverflow.com/questions/5636601/is-there-an-equivalent-to-the-ruby-shellwords-module-for-the-windows-shell

parisiale:
Thanks @Iristyle, I'll find another way to do that.

ploubser:
Closing this in favour of #196

PR #193, opened 2015-10-26
(maint) Fix compiling on Solaris

MikaelSmith:
Commit e0537e0289514db634dfe458075e9344f9d98f89 broke compilation on
Solaris, as stdin/stdout/stderr are macros and can't be assigned to.

The original change was to avoid cppcheck warnings. Change to checking
the return value and logging if there was a failure.

### DIFF
--- a/lib/src/util/posix/daemonize.cc
+++ b/lib/src/util/posix/daemonize.cc
@@ -242,9 +242,15 @@ std::unique_ptr<PIDFile> daemonize() {
 
     // Redirect standard files; we always use boost::log anyway
 
-    stdin = freopen("/dev/null", "r", stdin);
-    stdout = freopen("/dev/null", "w", stdout);
-    stderr = freopen("/dev/null", "w", stderr);
+    if (freopen("/dev/null", "r", stdin) == nullptr) {
+        LOG_WARNING("Failed to redirect stdin to /dev/null; errno=%1%", std::to_string(errno));
+    }
+    if (freopen("/dev/null", "w", stdout) == nullptr) {
+        LOG_WARNING("Failed to redirect stdout to /dev/null; errno=%1%", std::to_string(errno));
+    }
+    if (freopen("/dev/null", "w", stderr) == nullptr) {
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Testing a Solaris build.

kylog:
Makes sense. Should these be WARNINGs rather than DEBUGs? In the unlikely event that one of these fails, it seems like we'd want to know.

MikaelSmith:
I wouldn't be against that, but deferring to @parisiale.

MikaelSmith:
Verified with a build on Solaris 11.

parisiale:
:+1:  for me.
Perhaps logging at WARNING is more appropriate.

MikaelSmith:
Updated.

lib/src/util/posix/daemonize.cc
```diff
@@ -242,9 +242,15 @@ std::unique_ptr<PIDFile> daemonize() {
 
     // Redirect standard files; we always use boost::log anyway
 
-    stdin = freopen("/dev/null", "r", stdin);
-    stdout = freopen("/dev/null", "w", stdout);
-    stderr = freopen("/dev/null", "w", stderr);
+    if (freopen("/dev/null", "r", stdin) == nullptr) {
+        LOG_WARNING("Failed to redirect stdin to /dev/null; errno=%1%", std::to_string(errno));
```
kylog: Just noticed one more thing: instead of `errno=%1%", std::to_string(errno)` would we want something like `%1% (%2%)", strerror(errno), errno)`?

MikaelSmith: I was following prior art in this file, I could update all of them.

kylog: Ohhh. I didn't look around :)

We can save that for a later PR that normalizes the errno reporting.

parisiale: Thanks @kylog , I'll take care of that in a separate PR.

MikaelSmith:
This showed success even though Travis shows cppcheck failing...

MikaelSmith:
PR to circumvent it: https://github.com/puppetlabs/pxp-agent/pull/201

PR #196, opened 2015-10-27
(PCP-95) Safely pass command line options to Puppet::Util::Execution

ploubser:
Passing environment entries as a hash via :custom_environment to the
Puppet::Util::Execution.execute function.

Updating spec tests.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -53,11 +53,20 @@ def disabled?(run_result)
   !!(run_result =~ /disabled(.*?)Use 'puppet agent --enable' to re-enable/m)
 end
 
-def make_command_string(config, params)
-  env = params["env"].join(" ")
-  flags = params["flags"].join(" ")
+def make_environment_hash(params)
+  env_hash = {}
+  params["env"].each do |entry|
+    key, value = entry.split('=')
+    env_hash[key] = value
+  end
 
-  return "#{env} #{config["puppet_bin"]} agent #{flags}".lstrip.rstrip
+  return env_hash
+end
+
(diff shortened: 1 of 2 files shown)


modules/pxp-module-puppet
```diff
@@ -53,11 +77,20 @@ def disabled?(run_result)
...
+    env_dict[escape_string(key)] = escape_string(value)
+  end
+
+  return env_dict
+end
 
-  return "#{env} #{config["puppet_bin"]} agent #{flags}".lstrip.rstrip
+def make_command_vector(config, params)
+  cmd_vector = [config["puppet_bin"], "agent"]
+  cmd_vector +=  params["flags"].map{|flag| escape_string(flag)}
```
richardc: I'm confused by the need for this shellquoting.  If Puppet::Util::Execution.execute is capable of taking a vector of arguments for the command line, then surely it must the thing responsible for keeping it as a vector, and so shellquoting if need be.

modules/pxp-module-puppet
```diff
@@ -53,11 +77,20 @@ def disabled?(run_result)
...
 end
 
-def make_command_string(config, params)
-  env = params["env"].join(" ")
-  flags = params["flags"].join(" ")
+def make_environment_map(params)
+  env_dict = {}
+  params["env"].each do |entry|
+    key, value = entry.split('=')
+    env_dict[escape_string(key)] = escape_string(value)
```
richardc: Same as the other comment.  As the API is taking a Hash here, surely that API should ensure the keys/values are passed through in a safe manner.

PR #200, opened 2015-10-28
(PCP-136) Use static libraries to resolve AIX linking issues, and don't use SYSTEM

MikaelSmith:
##### Use static libraries to resolve AIX linking issues

AIX has a limit on shared library TOC size. Try static linking to avoid the issue.

##### Remove Boost.Process, don't use SYSTEM for headers

On AIX GCC assumes `-isystem` headers are all C, and will wrap
them in `extern "C"`. We were misusing the SYSTEM tag for C++ header
files. Stop doing that.

Boost.Process was previously removed and is no longer used. Fix the
CMake file that still referenced it.


MikaelSmith:
Needs a rebase after https://github.com/puppetlabs/pxp-agent/pull/201 is merged.

PR #206, opened 2015-10-30
PCP-136 Update submodule back to dynamic linking cpp-pcp-client

richardc:


jcoconnor:
Merge on recommendation of @richardc

PR #207, opened 2015-10-30
PCP-136 Correct target name

richardc:
We were attempting to set target properties on a target we don't create.


jcoconnor:
:+1:

PR #209, opened 2015-10-30
(PCP-136) Remove PIDFile ctor tests that assert errors

parisiale:
Removing trivial tests based on REQUIRE_THROWS_AS as they were creating
running problems on different platforms.


MikaelSmith:
I doubt that was the only problem, but I'll give it a shot later.

PR #210, opened 2015-11-02
(PCP-125) Configuration refactor; "unconfigured" after TLS init failure

parisiale:


exe/main.cc
```diff
@@ -10,18 +13,28 @@
...
 
+#include <memory>
+
 namespace PXPAgent {
 
 namespace HW = HorseWhisperer;
 namespace lth_file = leatherman::file_util;
 
+// Start a thread that just busy waits to facilitate acceptance testing
+void loopIdly() {
```
parisiale: Just factored out to make it easier to read.

exe/main.cc
```diff
@@ -40,90 +53,118 @@ int startAgent(std::vector<std::string> arguments) {
...
-        agent.start();
-        success = true;
-    } catch (const Agent::Error& e) {
-        LOG_ERROR("Fatal error: %1%", e.what());
-    } catch (const std::exception& e) {
-        LOG_ERROR("Unexpected error: %1%", e.what());
-    } catch (...) {
-        LOG_ERROR("Unexpected error");
+    bool success { true };
+    if (!Configuration::Instance().valid()) {
```
parisiale: Configuration "initialized" was ambiguous; "valid" is more clear, since what may lead to the "unconfigured" path is the validate() call.

exe/main.cc
```diff
@@ -40,90 +53,118 @@ int startAgent(std::vector<std::string> arguments) {
...
+        // pxp-agent will execute in uncofigured mode
+        loopIdly();
+    } else {
+        try {
+            Agent agent { Configuration::Instance().getAgentConfiguration() };
+            agent.start();
+        } catch (const Agent::WebSocketConfigurationError& e) {
```
parisiale: New. The underlying WebSocket TLS init handle will throw an exception in case it fails to load the cert / key pair, which  the Agent instance will propagate in that way.

exe/main.cc
```diff
@@ -40,90 +53,118 @@ int startAgent(std::vector<std::string> arguments) {
...
+    Configuration::Instance().initialize(startAgent);
+
+    // Parse options
 
     HW::ParseResult parse_result { HW::ParseResult::OK };
+    std::string err_msg {};
 
     try {
-        parse_result = Configuration::Instance().initialize(argc, argv);
+        parse_result = Configuration::Instance().parseOptions(argc, argv);
```
parisiale: Separating the 1) parse step from 2) setup logging and 3) validate.
That helps understanding why the logging can be set up (useful for PCP-134).

lib/inc/pxp-agent/agent.hpp
```diff
@@ -19,6 +19,14 @@ class Agent {
         explicit Error(std::string const& msg) : std::runtime_error(msg) {}
     };
 
+    struct WebSocketConfigurationError : public Error {
```
parisiale: Needed because we want to continue in "unconfigured" mode in that case.

lib/inc/pxp-agent/configuration.hpp
```diff
@@ -133,31 +133,37 @@ class Configuration {
         std::string client_type;
     };
 
-    /// Set the configuration entries to their default values.
-    /// Unset the initialized_ flag.
-    void reset();
```
parisiale: Not needed anymore. The unit tests will use initialize() for that.

lib/src/configuration.cc
```diff
@@ -88,14 +94,12 @@ static const std::string AGENT_CLIENT_TYPE { "agent" };
...
 
-void Configuration::reset() {
+void Configuration::initialize(
+        std::function<int(std::vector<std::string>)> start_function) {
+    // Ensure the state is reset (useful for testing)
     HW::Reset();
-    setDefaultValues();
-    initialized_ = false;
-}
+    valid_ = false;
```
parisiale: We want to make sure this state var is reset that for the unit tests.

exe/main.cc
```diff
@@ -40,90 +53,118 @@ int startAgent(std::vector<std::string> arguments) {
...
-        Agent agent { agent_configuration };
-        agent.start();
-        success = true;
-    } catch (const Agent::Error& e) {
-        LOG_ERROR("Fatal error: %1%", e.what());
-    } catch (const std::exception& e) {
-        LOG_ERROR("Unexpected error: %1%", e.what());
-    } catch (...) {
-        LOG_ERROR("Unexpected error");
+    bool success { true };
```
ploubser: Instead of using a bool, would it be simpler if we just used an integer called exit_status, and then assign it to EXIT_SUCCESS and EXIT_FAILURE as needed? Saves as the ternary operation at the end of the function.

lib/src/util/posix/pid_file.cc
```diff
@@ -27,21 +27,6 @@ PIDFile::PIDFile(const std::string& file_path_)
           file_path { file_path_ },
           pidfile_fd {},
           cleanup_when_done { false } {
-    if (fs::exists(file_path) && !(fs::is_regular_file(file_path))) {
```
parisiale: Moved to the Configuration::validate() method, since we don't check this in the unit tests anymore.
It is better to break sooner in case of invalid PID dir.

lib/src/configuration.cc
```diff
@@ -127,105 +141,134 @@ HW::ParseResult Configuration::initialize(int argc, char *argv[],
...
-
-        if (enable_logging) {
-            setupLogging();
-        }
-
-        validateAndNormalizeConfiguration();
-        setAgentConfiguration();
-
-        initialized_ = true;
+    } else {
```
ploubser: Drop the else, put the comment after `}`

lib/src/configuration.cc
```diff
@@ -127,105 +141,134 @@ HW::ParseResult Configuration::initialize(int argc, char *argv[],
...
 
-void Configuration::validateAndNormalizeConfiguration() {
-    // determine which of your values must be initalised
-    if (HW::GetFlag<std::string>("broker-ws-uri").empty()) {
-        throw Configuration::UnconfiguredError{ "broker-ws-uri value must be defined" };
-    } else if (HW::GetFlag<std::string>("broker-ws-uri").find("wss://") != 0) {
-        throw Configuration::UnconfiguredError { "broker-ws-uri value must start with wss://" };
+    if (fs::exists(logdir_path)) {
+        if (!fs::is_directory(logdir_path)) {
+            throw Configuration::Error { "log directory is not a directory" };
```
ploubser: Can we make this error look more like the next one? "cannot write to x; parent directory is not a directory" or something like that.

exe/main.cc
parisiale: Ok.

lib/src/configuration.cc
```diff
@@ -127,105 +141,134 @@ HW::ParseResult Configuration::initialize(int argc, char *argv[],
...
+        logfile_ = lth_file::tilde_expand(logfile_);
+
+        // NOTE(ale): we must validate the logifle path since we set
+        // up logging before calling validateAndNormalizeConfiguration
+        validateLogDirPath(logfile_);
+
+        logfile_fstream_.open(logfile_.c_str(), std::ios_base::app);
+        stream = &logfile_fstream_;
+    } else {
+        // Log on stdout by default
```
ploubser: We can drop this comment. We're not really doing anything here by default, we're logging to stdout because it was configured that way.

lib/src/configuration.cc
parisiale: Ok.

PR #211, opened 2015-11-02
(PCP-134) Return 2 on parsing failure

parisiale:


exe/main.cc
```diff
@@ -143,11 +159,11 @@ int main(int argc, char *argv[]) {
     try {
         Configuration::Instance().setupLogging();
         LOG_INFO("pxp-agent logging has been initialized");
-    } catch(const Configuration::Error& e) {
+    }  catch(const Configuration::Error& e) {
```
ploubser: Indentation

PR #212, opened 2015-11-04
(PCP-129) New PID-base status module

parisiale:
- internal modules (ping, echo, and status) must now be invoked only by blocking request
- add processExist() function for Windows
- write PID of non-blocking action jobs (separate process) on file
- rewrite status query action

### DIFF
--- a/lib/src/util/windows/process.cc
+++ b/lib/src/util/windows/process.cc
@@ -1,0 +1,27 @@
+#include <pxp-agent/util/process.hpp>
+
+#include <leatherman/windows/windows.hpp>
+#include <leatherman/windows/system_error.hpp>
+
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.util.windows.process"
+#include <leatherman/logging/logging.hpp>
+
+namespace PXPAgent {
+namespace Util {
+
+namespace lth_win = leatherman::windows;
+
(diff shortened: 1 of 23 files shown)


lib/src/util/windows/process.cc
```diff
@@ -0,0 +1,31 @@
...
+
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.util.windows.process"
+#include <leatherman/logging/logging.hpp>
+
+namespace PXPAgent {
+namespace Util {
+
+namespace lth_win = leatherman::windows;
+
+bool processExists(int pid) {
```
MikaelSmith: Can you file a ticket to move this to Leatherman? I'm alright with expedience now as long as we give ourselves work to clean it up later.

joshcooper:
If this is required for puppet-agent 1.3.0, please retarget for stable.

lib/src/util/windows/process.cc
parisiale: Will do.
I didn't put that in leatherman in the first place because, as you know, we need to speed up this.

parisiale:
@joshcooper will do.

parisiale:
Closing.
Will make another PR against the stable branch that will include the changes needed to write stdout / stderr directly to file while executing non-blocking actions.

PR #214, opened 2015-11-05
(PCP-125) Acceptance tests for pxp-agent with invalid SSL config

james-stocks:


joshcooper:
should we retarget for stable?

james-stocks:
Closing, needs targeted to stable instead

PR #222, opened 2015-11-09
(maint) Fix compilation on older versions of Clang

MikaelSmith:
Mac OS X 10.9 and 10.10 builds use older versions of Clang that had
initializer list issues. Fix compilation for those compilers.


MikaelSmith:
Verifying on OS X 10.10.

peterhuene:
:+1:

PR #223, opened 2015-11-10
(PA-46) Don't '--foreground' on startup on Solaris

geoffnichols:
Currently, once a Solaris pxp-agent successfully connects
to a PCP broker, the startup does not return. After 60 seconds,
SMF kills the service and places it into maintenance mode.

This commit removes `--foreground` from puppet-agent's SMF
SMF manifest, which allows the start method to return.

### DIFF
--- a/ext/solaris/smf/pxp-agent.xml
+++ b/ext/solaris/smf/pxp-agent.xml
@@ -23,7 +23,7 @@
       <service_fmri value="svc:/system/filesystem/local"/>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


underscorgan:
:+1:

joshcooper:
Please retarget stable if this is required for puppet-agent 1.3.0

geoffnichols:
Retargeting against stable. Please see https://github.com/puppetlabs/pxp-agent/pull/224.

PR #229, opened 2015-11-10
(maint) Add acceptance test for enable service

johnduarte:
This commit adds an acceptance test to ensure that the pxp-agent
service can be enabled and started.

[skip ci]

### DIFF
--- a/acceptance/tests/service_enable_start.rb
+++ b/acceptance/tests/service_enable_start.rb
@@ -1,0 +1,12 @@
+test_name '(C94777) Validate that pxp-agent service can be enabled and started'
+
+agents.each do |agent|
+  step 'start service'
+  on(agent, puppet('resource service pxp-agent ensure=running enable=true'))
+
+  step 'validate that service is enabled and running'
+  on(agent, puppet('resource service pxp-agent')) do |result|
+    assert_match(/ensure\s+=>\s+'running'/, result.stdout, 'pxp-agent failed to start')
+    assert_match(/enable\s+=>\s+'true'/, result.stdout, 'pxp-agent failed to enable')
+  end
+end


acceptance/tests/service_enable_start.rb
```diff
@@ -0,0 +1,12 @@
+test_name 'Validate that pxp-agent service can be enabled and started'
+
+agents.each do |agent|
+  step 'start service'
+  on(agent, puppet('resource service pxp-agent ensure=running enable=true'))
+
+  step 'validate that service is enabled and running'
+  on(agent, puppet('resource service pxp-agent')) do |result|
```
james-stocks: Lines 7-12 are perhaps redundant, as during line 5 Puppet will check that the property values it tried to set were set.

The only thing we get from repeating the check explicitly is perhaps that there is a slight delay before we go back and check again.

I'd say keep the test as-is; because even if the asserts are technically redundant, readers of this code will see what it is that we're interested in testing. If the entire test was just line 5 it might be confusing to understand how its even a test!

acceptance/tests/service_enable_start.rb
```diff
@@ -0,0 +1,12 @@
+test_name 'Validate that pxp-agent service can be enabled and started'
```
james-stocks: Not critical; but for the other tests I've been putting the Test Rail Case ID (C12345) at the start of the test_name

acceptance/tests/service_enable_start.rb
johnduarte: (shrug) when I tested line 5 on Soi 11, puppet returned an exit code of 0 even though the service state returned 'ensure=stopped' and 'enable=false'. So I put in the addition check to get back the actual state after the 60 second time out.

acceptance/tests/service_enable_start.rb
johnduarte: Thanks. I will add a case to TR and amend the test title.

johnduarte:
closing PR in order to re-target against stable

PR #232, opened 2015-11-11
(maint) Add acceptance test for enable service

johnduarte:
This commit adds an acceptance test to ensure that the pxp-agent
service can be enabled and started.

[skip ci]

### DIFF
--- a/acceptance/tests/service_stop_start.rb
+++ b/acceptance/tests/service_stop_start.rb
@@ -39,6 +39,18 @@ assert_stopped
 start_service
 assert_running
 
+step 'C94777 - Service enabled and started' do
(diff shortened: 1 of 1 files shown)


richardc:
We already have tests that test a service can be started as part of https://github.com/puppetlabs/pxp-agent/blob/stable/acceptance/tests/service_stop_start.rb so should this test only test for enabling?

Should the existing service_stop_start.rb tests do less? Or should this test case be added to service_stop_start.rb?  It seems odd to have service starting testing in multiple places.

johnduarte:
Thanks @richardc . I have incorporated the enable test into the existing start/stop test.

PR #233, opened 2015-11-11
(maint) Remove ssh-agent call

joshcooper:
Previously, acceptance was only failing in CI, but not when run locally,
and the failure was that the ca_crt.pem was never rsync'ed to the agent
node:

    WebSocket configuration error (ssl-ca-cert file '/root/test-resources/ssl/ca/ca_crt.pem' not found

I believe the root cause is due to using ssh-agent on the coordinator
node. This step was only necessary so we could git clone the private
pcp-broker repo. Now that the repo is public, the step is not necessary,
and it eliminates having to interact with ssh-agent on the shared CI
coordinator.

This commit also increases the beaker log level so we can see what
commands beaker is executing.

### DIFF
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -1,4 +1,5 @@
 {
+  :log_level     => 'debug',
   :type          => 'packages',
   :forge_host    => 'forge-aio01-petest.puppetlabs.com',
   :load_path     => './lib/',
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -8,19 +8,6 @@ namespace :ci do
     end
 
     task :aio do
-      # Start ssh-agent, parse the env variables from its output and add them to ENV
-      # The usual `eval $(ssh-agent)` does not work because
-      # the exported values will not persist after the statement completes
-      # TODO (JS) - QA-2108 - This no longer works for local execution and must be skipped.
-      if not ENV['LOCAL_EXECUTION']
-        `ssh-agent -t 24h -s`
-          .split.map{|e| e.tr(';', '')}
-          .map{|e| e.split('=') }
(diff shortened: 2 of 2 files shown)


joshcooper:
I executed this from a redhat-7 coordinator in the pool to ensure my local ssh-agent config wasn't corrupting the test:

```
# env TEST_TARGET=redhat-7-x86_64 SUITE_COMMIT=c1603e64d67ddc29e62d4ac280ecb59f56d6ac15 SUITE_VERSION=1.2.7.425.gc1603e6 bundle exec rake ci:test:aio SHA=d14f37f86e206ad724c3453db7c9ac1a0bc25c51 CONFIG=config/nodes/redhat-7-x86_64.yaml
..
      Test Suite: tests @ 2015-11-11 15:27:32 -0800

      - Host Configuration Summary -


       Total Suite Time: 32.10 seconds
      Average Test Time: 8.02 seconds
              Attempted: 4
                 Passed: 4
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 4
```

joshcooper:
I also noticed that we're still defaulting the acceptance type to [`packages`](https://github.com/puppetlabs/pxp-agent/blob/master/acceptance/config/packages/options.rb#L2) instead of `aio`. It didn't seem to hurt anything the way it is, but would be good to make that consistent in a separate maint commit.

acceptance/config/aio/options.rb
```diff
@@ -1,4 +1,5 @@
 {
+  :log_level     => 'debug',
```
MikaelSmith: Do you mean to have debug here?

kylog:
:+1:

kylog:
Merging. I'll kick CI also.

PR #234, opened 2015-11-12
(PCP-164) Quote puppet_bin property

MikaelSmith:
If puppet_bin is configured, and contains a space, executing it will
fail attempting to find an executable with the path preceding the space.
Quote unquoted paths with spaces to fix this.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -27,6 +27,15 @@ def is_win?
   return !!File::ALT_SEPARATOR
 end
 
+def quote(s)
+  quoted = (s[0] == '"' || s[0] == "'")
+  if !quoted && s.include?(' ')
+    '"'+s+'"'
+  else
+    s
+  end
+end
+
 def last_run_result(exitcode)
   return {"kind"             => "unknown",
           "time"             => "unknown",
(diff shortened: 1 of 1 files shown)


modules/pxp-module-puppet
```diff
@@ -62,7 +71,7 @@ def make_environment_hash(params)
 end
 
 def make_command_array(config, params)
-  cmd_array = [config["puppet_bin"], "agent"]
+  cmd_array = [quote(config["puppet_bin"]), "agent"]
```
richardc: This one should not be needed, as it's an array.

modules/pxp-module-puppet
```diff
@@ -38,7 +47,7 @@ def last_run_result(exitcode)
 end
 
 def check_config_print(cli_arg, config)
-  command = "#{config["puppet_bin"]} agent --configprint #{cli_arg}"
+  command = "#{quote(config["puppet_bin"])} agent --configprint #{cli_arg}"
```
richardc: This could be converted to the array form, save needing to quote.

Iristyle: Yeah, thanks for pointing that out.  I was digging in a bit on this PR yesterday as it didn't quite look right... and then when I next looked it got merged.

PR #236, opened 2015-11-12
(PCP-164) Quote puppet_bin property

ploubser:
The previous commit added quotes to command created by
 #make_command_array. Passing the quoted result to
Puppet::Util::Execution::execute would cause the result to be quoted
twice.

Here we remove the additional quoting and rely on passing an array to
Puppet::Util::Execution::execute to always take care of executing
correctly on Windows.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -27,15 +27,6 @@ def is_win?
   return !!File::ALT_SEPARATOR
 end
 
-def quote(s)
-  quoted = (s[0] == '"' || s[0] == "'")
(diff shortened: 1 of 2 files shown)


ploubser:
@joshcooper @kylog @MikaelSmith I have been unable to successfully run pxp-module-puppet on Windows after the previous commit, due to quoting happening twice. This commit resolves the issue.

richardc:
:+1: this addresses my concerns from #234

MikaelSmith:
:+1:

PR #238, opened 2015-11-12
(maint) Travis uses cpp-pcp-client#stable on stable

parisiale:


.travis_target.sh
```diff
@@ -24,6 +24,7 @@ fi
 
 git clone https://github.com/puppetlabs/cpp-pcp-client
 cd cpp-pcp-client
+git checkout stable
```
richardc: I think this should be a tag.  This is the travis setup for pxp-agent rather than an attempt to test-integrate on the fly.  Now we have cpp-pcp-client 1.0.0 we can just make this checkout of stable be a checkout of 1.0.0

parisiale:
Closing, as a proper fix would require knowing on what branch we are when executing travis / appveyor scripts; otherwise we would need to keep different scripts in different branches.

PR #240, opened 2015-11-12
(PCP-143) Association acceptance test asserts the correct log entry.

james-stocks:
The test previously only checked for websocket success. It now checks for association succeeding.
The assertion on websocket success has been left in so that in the event of a failure we have more information on when it failed.


james-stocks:
Closing - should not go into stable

PR #241, opened 2015-11-12
(PCP-143) Update association acceptance test to actually check for success

james-stocks:
The test previously only checked for websocket success. It now checks for association succeeding.
The assertion on websocket success has been left in so that in the event of a failure we have more information on when it failed.

### DIFF
--- a/acceptance/tests/pxp_agent_associate.rb
+++ b/acceptance/tests/pxp_agent_associate.rb
@@ -19,10 +19,16 @@ on agent1, puppet('resource service pxp-agent ensure=running')
 step 'Allow 10 seconds after service start-up for association to complete'
 sleep(10)
 
-success = /INFO.*Successfully established a WebSocket connection with the PCP broker/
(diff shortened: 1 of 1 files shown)


acceptance/tests/pxp_agent_associate.rb
```diff
@@ -19,10 +19,16 @@
...
   log_contents = result.stdout
-  assert_match(success, log_contents,
-               "Did not match expected association message '#{success.to_s}' " \
+  step 'Check pxp-agent.log for websocket connection'
+  assert_match(websocket_success, log_contents,
+               "Did not match expected websocket connection message '#{websocket_success.to_s}' " \
+               "in actual pxp-agent.log contents '#{log_contents}'")
+  step 'Check pxp-agent.log for successful association response'
+  assert_match(association_success, log_contents,
+               "Did not match expected association success message '#{association_success.to_s}' " \
```
parisiale: If you use the "Session Association" term instead of "association", I think you make it clear in the log message that what went wrong was the actual PCP connection.

PR #242, opened 2015-11-13
(maint) Acceptance tests should accept SHA as environment variable

james-stocks:
Our main build pipeline uses environment variable SUITE_COMMIT to store the SHA of Puppet Agent to fetch and test.
To re-use our tests in the Puppet Agent nightly pipeline, the environment variable SHA should be used instead.
This commit supports either during acceptance test setup.


james-stocks:
Closing, will re-open against master

PR #244, opened 2015-11-13
(maint) Add log message to blocking response function

parisiale:


acceptance/tests/pxp_agent_associate.rb
```diff
@@ -19,7 +19,7 @@
 step 'Allow 10 seconds after service start-up for association to complete'
 sleep(10)
 
-websocket_success = /INFO.*Successfully established a WebSocket connection with the PCP broker/
+websocket_success = /INFO.*Successfully established a WebSocket connection with the PCP broker.*/
```
parisiale: Necessary since the WS onOpen handler in cpp-pcp-client now shows the broker URL.

PR #245, opened 2015-11-14
(PCP-172) Synchronize status metadata writing

parisiale:
With this commit we synchronize reading and writing operations on
transaction status metadata files by non-blocking tasks and status
requests handlers.

The status query action now reads the PID file before accessing the
metadata file. In case the PID is related to an existent process, the
status handler will sleep for 0.1 s, to ensure that the non-blocking
task have the time to acquire the lock that enables writing the metadata.

Both the non-blocking task and status handler use a boost named_mutex to
synchronize the access to the metadata file. Such mutex will be named
after the first 8 letters of the transaction id of the interested
non-blocking task.

### DIFF
--- a/lib/src/modules/status.cc
+++ b/lib/src/modules/status.cc
@@ -2,10 +2,16 @@
 #include <pxp-agent/util/process.hpp>
 #include <pxp-agent/external_module.hpp>    // readNonBlockingOutcome
 
+// #include <cpp-pcp-client/util/thread.hpp>   // this_thread::sleep_for
+// #include <cpp-pcp-client/util/chrono.hpp>
+
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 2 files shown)


parisiale:
Note that this is not a fix for the issues reported in PCP-168; this prevents the agent from wrongly reporting a transaction status as "unknown", due to a race between the PID check and metadata access.

parisiale:
boost::interprocess seems to clash with PCPClient::Utill imports of boost:thread; I'll need to fix this.

parisiale:
I'm incurring in https://svn.boost.org/trac/boost/ticket/6145; boost::thread and boost::interprocess are incompatible in windows / mingw / gcc.
I will close this and use a different approach, as it doesn't feel safe to try solving this with a building hack.

PR #246, opened 2015-11-15
(PCP-172) Synchronize non-blocking tasks and status handlers

parisiale:
With this commit we synchronize reading and writing operations on
transaction status metadata files by non-blocking tasks and status
requests handlers.

The status query action now reads the PID file before accessing the
metadata file. In case the PID is related to a non-existent process, and
its mutex is still in the cache, the status handler will sleep for 0.1 s,
to ensure that the non-blocking task has the time to lock the mutex
and update the metadata file.

The synchronization relies on the ResultsMutex singleton that caches
mutexes named against transaction IDs.

### DIFF
--- a/lib/inc/pxp-agent/results_mutex.hpp
+++ b/lib/inc/pxp-agent/results_mutex.hpp
@@ -33,6 +33,11 @@ class ResultsMutex {
     using Lock = PCPClient::Util::unique_lock<PCPClient::Util::mutex>;
     using LockGuard = PCPClient::Util::lock_guard<PCPClient::Util::mutex>;
 
+    // TODO(ale): update this as soon as we have tagged cpp-pcp-client
+    static constexpr boost::defer_lock_t defer_lock() {
+        return boost::defer_lock_t();
+    };
+
     // Singleton users should lock this class mutex before accessing
     // the cache.
     // Note that the singleton caches shared_ptrs to mutexes; because
--- a/lib/inc/pxp-agent/results_mutex.hpp
+++ b/lib/inc/pxp-agent/results_mutex.hpp
@@ -1,0 +1,69 @@
+#ifndef SRC_AGENT_RESULTS_MUTEX_HPP_
+#define SRC_AGENT_RESULTS_MUTEX_HPP_
+
+#include <cpp-pcp-client/util/thread.hpp>
+
+#include <map>
+#include <string>
+#include <memory>
+#include <stdexcept>
+
+namespace PXPAgent {
+
+// NOTE(ale): this class does not provide general synchronizaion
+// functions; it provides a cache for named mutexes where it is
+// assumed that a single actor (the RequestProcessor instance, in
+// practice) is allowed to add and remove mutexes, whereas all actors
+// (including instances of the Transaction Status request handler) can
+// lock the cached mutexes
+
+class ResultsMutex {
+  public:
+    struct Error : public std::runtime_error {
+        explicit Error(std::string const& msg) : std::runtime_error(msg) {}
+    };
+
+    static ResultsMutex& Instance() {
+        static ResultsMutex instance {};
+        return instance;
+    }
+
+    using Mutex = PCPClient::Util::mutex;
+    using Mutex_Ptr = std::shared_ptr<PCPClient::Util::mutex>;
+    using Lock = PCPClient::Util::unique_lock<PCPClient::Util::mutex>;
+    using LockGuard = PCPClient::Util::lock_guard<PCPClient::Util::mutex>;
+
+    // Singleton users should lock this class mutex before accessing
+    // the cache.
+    // Note that the singleton caches shared_ptrs to mutexes; because
+    // of that, it is safe to use the mutex pointer returned by get()
+    // after releasing the access_mtx lock, even if a concurrent
+    // thread is removing such mutex from the cache.
+    Mutex access_mtx;
+
+    // Useful for testing
+    void reset();
+
+    // Whether the specified transaction mutex exists
+    bool exists(std::string const& transaction_id);
+
+    // Throw Error if transaction_id does not exist
+    Mutex_Ptr get(std::string const& transaction_id);
+
+    // Throw Error if transaction_id exists
+    void add(std::string const& transaction_id);
+
+    // Throw Error if transaction_id does not exist
+    void remove(std::string const& transaction_id);
+
+  private:
+    // Cache
+    std::map<std::string, Mutex_Ptr> mutexes_;
+
+    // Private ctor
+    ResultsMutex();
+};
+
+}  // namespace PXPAgent
+
+#endif  // SRC_AGENT_RESULTS_MUTEX_HPP_
--- a/lib/src/modules/status.cc
+++ b/lib/src/modules/status.cc
@@ -1,6 +1,10 @@
 #include <pxp-agent/modules/status.hpp>
 #include <pxp-agent/util/process.hpp>
 #include <pxp-agent/external_module.hpp>    // readNonBlockingOutcome
+#include <pxp-agent/results_mutex.hpp>
+
+#include <cpp-pcp-client/util/thread.hpp>   // this_thread::sleep_for
+#include <cpp-pcp-client/util/chrono.hpp>
 
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/operations.hpp>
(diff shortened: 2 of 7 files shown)


parisiale:
This requires https://github.com/puppetlabs/cpp-pcp-client/pull/117.

parisiale:
This is related to the previously closed PR https://github.com/puppetlabs/pxp-agent/pull/245.

lib/src/request_processor.cc
```diff
@@ -124,32 +189,34 @@ void nonBlockingActionTask(std::shared_ptr<Module> module_ptr,
...
         results_storage.writeMetadata(exit_code, exec_error, duration);
-    } catch (const std::exception& e) {
+    } catch (const ResultsStorage::Error& e) {
         LOG_ERROR("Failed to write metadata of non blocking request %1%: %2%",
-                  job_id, e.what());
+                  transaction_id, e.what());
     }
-
-    // Flag end of processing
-    *done = true;
```
parisiale: Moved to the above scope_exit instance.

lib/src/modules/status.cc
```diff
@@ -134,63 +143,102 @@ ActionOutcome Status::callAction(const ActionRequest& request) {
     LOG_DEBUG("Retrieving results for job %1% from %2%",
               t_id, results_dir_path.string());
 
-    auto metadata_file = (results_dir_path / "metadata").string();
+    bool running_by_pid { false };
+    bool not_running_by_pid { false };
+    auto pid_file = (results_dir_path / "pid").string();
```
ploubser: Could we rename this to action_pid_file or something similar?

lib/src/modules/status.cc
```diff
@@ -134,63 +143,102 @@ ActionOutcome Status::callAction(const ActionRequest& request) {
...
+    std::string pid_txt {};
```
ploubser: Same as above

lib/src/modules/status.cc
```diff
@@ -134,63 +143,102 @@ ActionOutcome Status::callAction(const ActionRequest& request) {
...
+    int pid {};
+
+    // Read the PID file before the metadata, to avoid a race in the
+    // non-blocking task thread, due to the elapsed time between the
+    // end of the job and the metadata write
+    if (!fs::exists(pid_file)) {
+        LOG_DEBUG("PID file '%1%' does not exist", pid_file);
```
ploubser: And again, maybe log it as Action PID file.

lib/src/modules/status.cc
parisiale: You're inside the ActionOutcome namespace, is it really confusing?
I find it ugly to have such redundancy ("action" - "action") in the instance variables names: ActionOutcome::action_pid_file.
Besides, that's exactly the name of the files of the directory: metadata, pid, stdout, stderr.

I'm happy to change if you really think it's confusing.

ploubser: Nope, your point is valid.

ploubser:
:+1: 

@MikaelSmith @kylog Would appreciate some more eyes here if possible.

lib/src/modules/status.cc
```diff
@@ -21,8 +26,10 @@ namespace Modules {
 namespace fs = boost::filesystem;
 namespace lth_file = leatherman::file_util;
 namespace HW = HorseWhisperer;
+namespace pcp_util = PCPClient::Util;
 
 static const std::string QUERY { "query" };
+static const uint32_t METADATA_RACE_MS { 100 };
```
MikaelSmith: Where did this value come from?

lib/src/modules/status.cc
```diff
@@ -134,63 +143,102 @@ ActionOutcome Status::callAction(const ActionRequest& request) {
...
+                mtx_ptr = ResultsMutex::Instance().get(t_id);
+            }
+        }
+        if (mtx_ptr != nullptr) {
+            LOG_TRACE("Non-blocking transaction %1% mutex is cached; "
+                      "locking it", t_id);
+            ResultsMutex::LockGuard r_l { *mtx_ptr };  // RAII
+            metadata = ActionMetadata(metadata_file);
+        } else {
+            LOG_TRACE("Non-blocking transaction %1% mutex is not cached", t_id);
```
MikaelSmith: This is safe because no child holds a mutex, and the current thread is the only one that would cause one to start and claim the mutex?

lib/inc/pxp-agent/results_mutex.hpp
```diff
@@ -0,0 +1,74 @@
...
+        static ResultsMutex instance {};
+        return instance;
+    }
+
+    using Mutex = PCPClient::Util::mutex;
+    using Mutex_Ptr = std::shared_ptr<PCPClient::Util::mutex>;
+    using Lock = PCPClient::Util::unique_lock<PCPClient::Util::mutex>;
+    using LockGuard = PCPClient::Util::lock_guard<PCPClient::Util::mutex>;
+
+    // TODO(ale): update this as soon as we have tagged cpp-pcp-client
```
MikaelSmith: What pcp change is this waiting on?

MikaelSmith:
I think it generally looks ok. I don't have the mental energy right now to be sure I've thought through everything.

lib/inc/pxp-agent/results_mutex.hpp
parisiale: Adding the boost::defer_lock const and tagging it; Travis and AppVeyor use cpp-pcp-client 1.0.1, so we currently use the function defined in ResultsMutex, but it's not nice for breaking the PCPClient abstraction.

lib/src/modules/status.cc
parisiale: The race @nicklewis talks about in PCP-168 is due to a very short time window, where the non-blocking process terminates, leatherman's execute() returns, and the non-blocking action task updates the metadata file. If a status request is processed during such interval, the PID exists() function will return false, but metadata.completed would be (wrongly) false. In that case (PID does not exist and metadata still not updated) the above delay will allow the non-blocking task to acquire the transaction mutex and update the metadata before the status handler reads it. The duration value was chosen by considering that 1) the race window is very tight and 2) 100 ms is an order of magnitude less compared to the status polling period of the clj-pxp-puppet function (it shouldn't impact the overall transaction).

lib/src/modules/status.cc
parisiale: There's no child - parent relationship here. We have a non-blocking action task that is created by a non-blocking request (pxp module puppet) and a blocking action handler (transaction status); they're both different threads of execution compared to the main pxp-agent one.

This is safe due to what is assumed in https://github.com/puppetlabs/pxp-agent/pull/246/files#diff-4940fea1e12e8b651312b2025f483274R13; if the mutex is not cached (the action task already wrote it or pxp-agent previously terminated while the action was still executing), we know that nobody will write its metadata file.

MikaelSmith: Is there a chance a new asynchronous task with the same pid could be started and claim the same mutex during this operation?

parisiale: I'm not sure I understand your question, because the non-blocking task that executes external modules (say, pxp module puppet) and uses ResultsMutex is https://github.com/puppetlabs/pxp-agent/blob/master/lib/src/request_processor.cc#L106; it's a different thread within the pxp-agent process - it doesn't have a different PID.

The non-blocking task is the one the writes the PID of the external module process to file, via leatherman's execution. But that PID file is stored in the spool directory named after the PXP transaction ID; so, if we assume that such UUID is unique, the writing operation is always safe, regardless of the PID value. But I guess this not is what you're concerned of...

As for accessing the metadata file, if the mutex named after the transaction ID is not cached in the ResultsMutex singleton, it means that the metadata will not be updated. But, regardless of that, the non-blocking action writes the metadata file with atomic_write_to_file(), so reading should always be safe. What we enforce with this commit is that the transaction status handler will wait for the non-blocking until the metadata is updated, when necessary - the transaction status handler only reads the metadata; concurrent transaction status handlers can operate safely.

Also, note that in ResultsMutex we have double locking with the singleton's access_mutex (that must be locked to access the cache) and the mutexes of the transaction currently being processed (that must be locked, when existent, to access the metadata file).

Hope this clarifies a bit.

MikaelSmith: I think I follow. I was probably concerned about how PID files were named, and using a UUID makes sense. :+1:

PR #247, opened 2015-11-16
(PCP-169) Acceptance tests should not use sed

james-stocks:
Tests fail on some supported platforms due to the use of 'sed -i'.
This PR adds a helper lib file to provide constants and helper methods for pxp-agent. This includes helper methods to generate pxp-agent config JSON.
The sed statements that altered pxp-agent.conf have been replaced with beaker helper methods that replace the config file's content with generated JSON.

### DIFF
--- a/acceptance/lib/pxp-agent/config_helper.rb
+++ b/acceptance/lib/pxp-agent/config_helper.rb
@@ -1,0 +1,151 @@
+# This file contains common strings and functions for pxp-agent acceptance tests to use
+
+PXP_CONFIG_DIR_CYGPATH = '/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/etc/'
+PXP_CONFIG_DIR_POSIX = '/etc/puppetlabs/pxp-agent/'
+
+PCP_BROKER_PORT = 8142
+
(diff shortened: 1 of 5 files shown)


acceptance/lib/pxp-agent/config_helper.rb
```diff
@@ -0,0 +1,243 @@
...
+# @return pxp-agent config, with ssl ca cert intentionally not matching its ssl cert and key, as a JSON object
+def pxp_invalid_config_wrong_ca_json(broker, agent, client_number)
+  pxp_invalid_config_wrong_ca_hash(broker, agent, client_number).to_json
+end
+
+# @param broker the host that runs pcp-broker
+# @param agent the agent machine that this config is for
+# @param client_number which client cert 01-05 to use
+# @return pxp-agent config, with ssl ca cert intentionally not matching its ssl cert and key, as a hash
+def pxp_invalid_config_wrong_ca_hash(broker, agent, client_number)
```
johnduarte: The creation of the values here seems too tightly coupled to static files supplied. Can we just have strings passed in for these here to allow for user supplied values? When this is tested against an running puppetserver, it would be preferable to use the servers ca infrastructure.

james-stocks:
Hi @johnduarte - I updated this PR so that the pxp_config_json method is totally configurable and not tied to our test certs. I renamed the existing methods to reflect that they use our test certs.

PR #251, opened 2015-11-20
Use lastrunreport setting instead of guessed value

vinzent:
Better use the configured value instead of joining statedir and a fixed string.


ploubser:
Awesome, thanks.

PR #253, opened 2015-11-20
Make agent run more robust

vinzent:
Check if the agent disabled lockfile already exists before running the agent

Before running create a disabled lockfile containing a message that pxp run is
in progress. Ensure this file is unlinked afterwards.

Use a pxp specific agent disabled lockfile to prevent further runs by the
daemon, cronjobs or manually. The others will get a nice message that a pxp run
is in progress.

If a puppet run is already in progress, try to wait some time. max waittime is
60s now. this should be configurable.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -115,25 +115,59 @@ def start_run(config, params)
   env = make_environment_hash(params)
   start_time = Time.now
 
-  run_result = Puppet::Util::Execution.execute(cmd, {:failonfail => false,
-                                                     :custom_environment => env})
+  agent_disabled_lockfile = check_config_print("agent_disabled_lockfile",config)
+  agent_disabled_lockfile_pxp = agent_disabled_lockfile + '.pxp'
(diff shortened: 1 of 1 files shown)


ploubser:
Thanks @vinzent, really appreciate the attention you're giving to the project. 

The behaviour you're seeing here is intented, as can be seen here https://tickets.puppetlabs.com/browse/PCP-117

Since the project is still fairly new and very much intertwined with future work, I recommend that you open a ticket at https://tickets.puppetlabs.com/browse/PCP before investing time in making any big changes.

vinzent:
The description of PCP-117 is valid for the old code. there was really some time between the checks of the files and will lead to troubles.

IMHO this is not valid for my patch: 

a) the agent_catalog_run_lockfile is only considered to wait some time, it won't fail the run, the agent run will execute and the running? function is still called afterwards. it only tries to wait some time until the run finished. 

b) IMHO if the agent is disabled  it really don't makes any sense to run the agent. 

c) there is no time between checking the lockfile and creating it. I'm using this pattern successfully at a customers site in a pre/post part of the monthly patch script.

PR #254, opened 2015-11-21
(PCP-174) Acceptance - expand install utils

johnduarte:
This commit expands the acceptance install_utils in order to
support installation on the full range of platforms for
acceptance testing.

It also incorporates the Rakefile structure from puppet/facter/hiera
in order to allow Jenkins pipeline support for AIO. The Rakefile
retains usage and checks for SUITE_COMMIT and TEST_TARGET to ensure
that it continues to work with the existing Jenkins job parameters.

[skip ci]


acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -1,24 +1,10 @@
-require 'puppet/acceptance/puppet_helpers'
-extend Puppet::Acceptance::PuppetHelpers
-
 module Puppet
   module Acceptance
-    module CAUtils
```
richardc: This seems like it was weird before, and continues to be weird after this change.

Before `lib/puppet/acceptance/common_utils.rb` was declaring the namespace `Puppet::Acceptance::CAUtils`, and now it's declaring `Puppet::Acceptance::CommandUtils`.  I'd say that neither of those is correct, and that `lib/puppet/acceptance/common_utils.rb` should be declaring the namespace `Puppet::Acceptance::CommonUtils`

johnduarte: @richardc yes, it certainly seems like we should be naming these consistently.

@kylog I copied the `lib/puppet/acceptance/common_utils.rb` file from facter. Should we re-namespace this here and in the other projects for consistency?

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -0,0 +1,39 @@
+require 'puppet/acceptance/common_utils'
```
james-stocks: If this file is included; then we should completely replace acceptance/setup/aio/010_Install.rb
There's some code at the bottom of acceptance/setup/aio/010_Install.rb that needs retained. It can perhaps move to setup/common/020_Configure_Pxp_Agents.rb ?

acceptance/Rakefile
```diff
@@ -1,30 +1,343 @@
...
-    task :aio do
-      if (ENV['SHA']) && (not ENV['SUITE_COMMIT'])
-        ENV['SUITE_COMMIT'] = ENV['SHA']
-      end
-      unless ENV['SUITE_COMMIT'] then
-        fail('SUITE_COMMIT or SHA environment variable must be set to the SHA of the puppet-agent package to test')
+module HarnessOptions
+
+  DEFAULTS = {
+    :type => 'git',
```
james-stocks: Should the default be 'aio' ?
Possibly we should strip out anything referring to a git setup; as we don't support such a thing. Or possibly we keep it in, to keep uniform with other projects?

acceptance/Rakefile
```diff
@@ -1,30 +1,343 @@
...
+    :helper => ['lib/helper.rb'],
+    :tests  => ['tests'],
+    :log_level => 'debug',
```
james-stocks: I think this should be 'info'. Unless Jenkins always specifies a logging level; and debug is a useful default, assuming that someone running is someone troubleshooting?

acceptance/lib/puppet/acceptance/common_utils.rb
richardc: So if we're copying in artefacts from other projects, it surprises me a little that facter is the project of record for the namespace `Puppet::Acceptance`.  Is this Puppet in the sense of 'Puppet the company', or was it just a convenient place to grab it from as it should be the same across puppet and facter repos?

Actually.  This is fairly icky too:  https://github.com/puppetlabs/puppet/blob/master/acceptance/lib/puppet/acceptance/common_utils.rb in puppet's tree that one file declares `Puppet::Acceptance::{CronUtils,CAUtils, CommandUtils}`, which though it explains 'common' in `common_utils.rb` to mean  'random grab bag', is still fairly poor mapping from namespace to file system space.

If we're confident that this PR is self-consistent I'm happy to merge and then continue the discussion about shared test library hygene and lifecycle in some other avene.

johnduarte: @richardc I agree that the name-spacing is not ideal. We are working on pulling this stuff out into an external gem, but it is not of high priority at the moment.

This PR is a carbon copy of `lib/puppet/acceptance/common_utils.rb` as it exists currently in `facter`.

acceptance/setup/aio/pre-suite/010_Install.rb
johnduarte: @james-stocks I added the needed install code to `setup/aio/pre-suite/010_Install.rb` and removed `/setup/aio/010_Install.rb`. Seemed undesirable to add it to `setup/common/020_Configure_Pxp_Agents.rb` since it is AIO specific.

james-stocks:
Looks Good To Me. @richardc - any further comments on the namespace item?

PR #256, opened 2015-11-23
(PCP-176) Support 32bit Puppet on 64bit Windows

james-stocks:
Puppet and pxp-agent will be in the Program Files x86 folder if Puppet is 32bit on 64bit Windows.
To support this, I've added a helper method to get the path to pxp-agent.
Removed the static strings in the pxp_agent_version test and replaced with a call to the helper.

### DIFF
--- a/acceptance/lib/pxp-agent/config_helper.rb
+++ b/acceptance/lib/pxp-agent/config_helper.rb
@@ -7,10 +7,6 @@ PXP_CONFIG_DIR_POSIX = '/etc/puppetlabs/pxp-agent/'
 PXP_LOG_FILE_CYGPATH = '/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/var/log/pxp-agent.log'
 PXP_LOG_FILE_POSIX = '/var/log/puppetlabs/pxp-agent/pxp-agent.log'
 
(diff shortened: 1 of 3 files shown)


acceptance/lib/pxp-agent/config_helper.rb
```diff
@@ -7,6 +7,10 @@
 PXP_LOG_FILE_CYGPATH = '/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/var/log/pxp-agent.log'
 PXP_LOG_FILE_POSIX = '/var/log/puppetlabs/pxp-agent/pxp-agent.log'
 
+PXP_EXE_DIR_CYGPATH = "/cygdrive/c/Program\\ Files/Puppet\\ Labs/Puppet/pxp-agent/bin"
+PXP_EXE_DIR_CYGPATH_x86 = "/cygdrive/c/Program\\ Files\\ \\(x86\\)/Puppet\\ Labs/Puppet/pxp-agent/bin"
+PXP_EXE_DIR_POSIX = "/opt/puppetlabs/puppet/bin"
+
```
richardc: Premature declaration of a constant, they're here so pxp_agent_dir has something to return, and only so pxp_agent_dir has something to return.  They should just be string literals in the pxp_agent_dir function.

PR #259, opened 2015-11-24
(maint) Acceptance pre-suite improvements

james-stocks:
One commit to implement Ethan's improvements from PR 189
A second commit to ensure each beaker step in the pre-suite has a do block

### DIFF
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -13,23 +13,25 @@ step "Install repositories on target machines..." do
   end
 end
 
-PACKAGES = {
-  :redhat => [
-    'puppet-agent',
-  ],
-  :debian => [
-    'puppet-agent',
-  ],
-}
+step "Install puppet-agent on hosts" do
+  PACKAGES = {
+    :redhat => [
+      'puppet-agent',
+    ],
+    :debian => [
+      'puppet-agent',
+    ],
+  }
 
-install_packages_on(hosts, PACKAGES)
+  install_packages_on(hosts, PACKAGES)
 
-# make sure install is sane, beaker has already added puppet and ruby
-# to PATH in ~/.ssh/environment
-agents.each do |agent|
-  on agent, puppet('--version')
-  ruby = Puppet::Acceptance::CommandUtils.ruby_command(agent)
-  on agent, "#{ruby} --version"
+  # make sure install is sane, beaker has already added puppet and ruby
+  # to PATH in ~/.ssh/environment
+  agents.each do |agent|
+    on agent, puppet('--version')
+    ruby = Puppet::Acceptance::CommandUtils.ruby_command(agent)
+    on agent, "#{ruby} --version"
+  end
 end
 
 # This step is unique to pxp-agent
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -44,3 +44,25 @@ step 'Add path for pxp-agent.exe on Windows' do
     end
   end
 end
+
+step 'Install build dependencies on master' do
+
+ MASTER_PACKAGES = {
+    :redhat => [
+      'git',
+      'java-1.8.0-openjdk-devel',
+    ],
+  }
+  install_packages_on(master, MASTER_PACKAGES, :check_if_exists => true)
+end
+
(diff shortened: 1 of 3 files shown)


acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -44,3 +46,25 @@
...
+end
+
+step 'Confirm SSH Forwarding Enabled on master' do
+  ssh_auth_sock = on(master, 'echo $SSH_AUTH_SOCK').stdout.chomp
+  if ssh_auth_sock.nil? || ssh_auth_sock.empty?
+    fail_test('SSH forwarding not configured properly - check ~/.ssh/config for ForwardAgent settings')
+  end
+end
+
+step 'Allow git clones via ssh to "unknown" host at github.com; required to clone private repos.' do
```
richardc: What private repos are we cloning?

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -44,3 +46,25 @@
...
+ MASTER_PACKAGES = {
+    :redhat => [
+      'git',
+      'java-1.8.0-openjdk-devel',
+    ],
+  }
+  install_packages_on(master, MASTER_PACKAGES, :check_if_exists => true)
+end
+
+step 'Confirm SSH Forwarding Enabled on master' do
```
richardc: What is this used for?

acceptance/setup/aio/pre-suite/010_Install.rb
james-stocks: None. I think it refers to clj-pxp-puppet , but I removed the checkout of that repo as it's not currently used. I'll correct this comment.

acceptance/setup/aio/pre-suite/010_Install.rb
james-stocks: Do you mean the ssh forwarding check starting at line 61? I believe it makes the ssh auth available to the user executing beaker also available to the user on the test machine when beaker has ssh'ed onto it.

Looking at it, I'm not sure if we should be testing it every run; or if we do test it, it should be as early as possible in the setup so we fail faster

richardc: I mean what is beaker, or this particular test suite, using ssh forwarding for?

acceptance/setup/aio/pre-suite/010_Install.rb
richardc: Rather than correcting the comment, I think we just don't need to have the step.

acceptance/setup/aio/pre-suite/010_Install.rb
james-stocks: The git clone of pcp-broker via git (rather than http) depends on it. The previous code (before this PR) was doing the git clone over http and got away without having the github key set properly.

acceptance/setup/aio/pre-suite/010_Install.rb
james-stocks: Git clone of pcp-broker using the beaker clone_git_repo_on method fails without it.

acceptance/setup/aio/pre-suite/010_Install.rb
richardc: Then you mean it's a clone via ssh, not via git.  git's own git:// wire protocol is distinct and doesn't need ssh setup.

I'm not convinced of the need for this extra ssh set-up.  Beaker's helper support https urls, https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L42-L50 so we can continue to use that even with using the standard git helper.

acceptance/setup/aio/pre-suite/010_Install.rb
richardc: It can be pointed at a https uri, so what else?

johnduarte:
:+1: lgtm

Iristyle:
@james-stocks it looks like there might have been a few things that got muddled in this PR.  I repushed my original PR, retargeted at stable (and removed the ssh-agent / clj-pxp-puppet bits).

So you might want to take a look at / potentially merge that.

richardc:
@james-stocks please stage a new PR just with this second commit in it, so we're not losing the history and intent from the Ethan's original set of commits.

james-stocks:
Closing in favour of #261 

Will do a separate PR with do blocks for every step

PR #260, opened 2015-11-24
(maint) Acceptance - refactor alt helper methods

johnduarte:
This commit removes the acceptance helper methods for building
the alt file paths for pxp-agent certs. It replaces these
methods with an additional boolean parameter on the default
file path methods. This parameter will instruct the method
whether to return the standard or alt paths based on its value.

[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/config_helper.rb
+++ b/acceptance/lib/pxp-agent/config_helper.rb
@@ -13,9 +13,8 @@ PCP_BROKER_PORT = 8142
 SSL_BASE_DIR_WINDOWS = "C:\\\\cygwin64\\\\home\\\\Administrator\\\\test-resources\\\\ssl\\\\"
 SSL_BASE_DIR_POSIX = "/root/test-resources/ssl/"
 SSL_BASE_DIR_OSX = "/var/root/test-resources/ssl/"
-ALT_SSL_BASE_DIR_WINDOWS = "C:\\\\cygwin64\\\\home\\\\Administrator\\\\test-resources\\\\ssl\\\\alternative\\\\"
-ALT_SSL_BASE_DIR_POSIX = "/root/test-resources/ssl/alternative/"
-ALT_SSL_BASE_DIR_OSX = "/var/root/test-resources/ssl/alternative/"
+ALT_SUFFIX_WINDOWS = "alternative\\\\"
+ALT_SUFFIX_POSIX = "alternative/"
 SSL_KEY_FILE_DIR_WINDOWS = "private_keys\\\\"
 SSL_KEY_FILE_DIR_POSIX = "private_keys/"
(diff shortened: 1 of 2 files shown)


acceptance/lib/pxp-agent/config_helper.rb
```diff
@@ -119,89 +118,55 @@ def pxp_agent_config_file(host)
 
 # @param host a beaker host (to determine the correct path for the OS)
 # @param client_number which client 01-05 you want the key file for
+# @param [boolean] use_alt_path use alt file path?
```
james-stocks: I think if someone new comes onto the project; this doc comment might be confusing. Perhaps

# @param [boolean] use_alt_path Use the repo's alternative set of ssl certs? Default false, use the standard set of ssl certs

(Too wordy?)

johnduarte: Yard will introspect the default from the method definition. Each method only defines a single path, not a set of certs. Maybe the following is more accurate?

    # @param [boolean] use_alt_path Use alternative test ssl file if true, otherwise use standard test ssl file

james-stocks: Perfect!

james-stocks:
Looks good to me.
If this merges, https://github.com/puppetlabs/pxp-agent/pull/262 may need updated, as the PRs are in the same area

acceptance/tests/invalid_ssl_config.rb
```diff
@@ -21,7 +21,7 @@
 invalid_config_mismatching_keys = {:broker_ws_uri => broker_ws_uri(master),
                                    :ssl_key => ssl_key_file(agent1, 1),
                                    :ssl_ca_cert => ssl_ca_file(agent1),
-                                   :ssl_cert => alt_ssl_cert_file(agent1, 1)}
+                                   :ssl_cert => ssl_cert_file(agent1, 1, use_alt_path=true)}
```
richardc: This isn't python, so we don't have semi-named arguments here.  All this did was introduce a temporary variable we ignored.

```
ssl_cert_file(agent1, 1, eat_at_joes=true)
```

has the same effect.

richardc: Actually to correct myself, ruby 2.0 did introduce named keyword parameters, so If the intent was to use those the code would need to look more like this:

```
def ssl_cert_file(host, client_number, use_alt_path: false)
  # code here
end

ssl_cert_file(agent1, 1, use_alt_path: true)
```

In that pattern use_alt_path is the name of the parameter.

richardc: @james-stocks has raised #264 to correct this.

richardc: @james-stocks also raised #265 as a non-ruby 2.0 requiring version to correct this.

PR #261, opened 2015-11-24
(maint) Acceptance cleanup

Iristyle:
This is the code from my prior PR, rebased against stable... compare to #259

### DIFF
--- a/acceptance/setup/aio/010_Install.rb
+++ b/acceptance/setup/aio/010_Install.rb
@@ -11,7 +11,9 @@ step "Install repositories on target machines..." do
   repo_configs_dir = 'repo_configs'
   logger.debug('about to install repo for puppet-agent from ' + sha.to_s + ' ' + repo_configs_dir.to_s)
   hosts.each do |host|
-    install_repos_on(host, 'puppet-agent', sha, repo_configs_dir)
+    if !host.platform.start_with?('windows')
+      install_puppetlabs_dev_repo(host, 'puppet-agent', sha, repo_configs_dir, { :dev_builds_repos => 'PC1'})
+    end
   end
 end
 
--- a/acceptance/setup/aio/010_Install.rb
+++ b/acceptance/setup/aio/010_Install.rb
@@ -27,6 +27,16 @@ PACKAGES = {
 # Install puppet on all hosts including master
 install_packages_on(hosts, PACKAGES)
 
+step 'Install build dependencies on master'
+
+MASTER_PACKAGES = {
+  :redhat => [
+    'git',
+    'java-1.8.0-openjdk-devel',
+  ],
+}
+install_packages_on(master, MASTER_PACKAGES, :check_if_exists => true)
+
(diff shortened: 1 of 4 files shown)


james-stocks:
Can this be closed and re-opened against master please? Nothing further should go into stable; and the changes in this PR are behind a number of changes to acceptance pre-suite on master.

acceptance/tests/pxp_agent_version.rb
```diff
@@ -1,12 +1,17 @@
 test_name 'C93805 - pxp-agent - Versioning test'
 
-agent1 = agents[0]
+def pxp_version_cmd(platform)
+  if platform.start_with?('windows')
+    # simple hack to ensure OpenSSL is found w/out modifying PATH
+    "cd /cygdrive/c/Program\\ Files/Puppet\\ Labs/Puppet/sys/ruby/bin && ../../../pxp-agent/bin/pxp-agent --version"
```
james-stocks: The case of 32bit Ruby where Puppet is in C:\Program Files (x86) needs to be supported here

ploubser:
Can we target this at master?

james-stocks:
Comments not addressed after over a month, and some of the changes in the PR have been implemented in the meantime. Created ticket PCP-235 to ensure any changes that were in this PR that are still applicable get done.

acceptance/tests/pxp_agent_version.rb
```diff
@@ -1,12 +1,17 @@
...
+  else
+    'cd /opt/puppetlabs/puppet/bin && ./pxp-agent --version'
+  end
```
joshcooper: Puppet acceptance has a helper for executing other private commands: https://github.com/puppetlabs/puppet/blob/master/acceptance/lib/puppet/acceptance/common_utils.rb#L128-L147. Since privatebindir on Windows includes both "Program Files" and "Program Files (x86)", it works for both 64 and 32-bit puppet.

acceptance/setup/aio/010_Install.rb
```diff
@@ -11,7 +11,9 @@
   repo_configs_dir = 'repo_configs'
   logger.debug('about to install repo for puppet-agent from ' + sha.to_s + ' ' + repo_configs_dir.to_s)
   hosts.each do |host|
-    install_repos_on(host, 'puppet-agent', sha, repo_configs_dir)
+    if !host.platform.start_with?('windows')
+      install_puppetlabs_dev_repo(host, 'puppet-agent', sha, repo_configs_dir, { :dev_builds_repos => 'PC1'})
+    end
```
joshcooper: We should be able to use `install_puppet_agent_dev_repo` on all hosts (in beaker 2.23.0 and up), and it will do the right thing for all agent platforms. There's some additional complexity if you are installing a master though: https://github.com/puppetlabs/puppet/blob/master/acceptance/setup/aio/pre-suite/010_Install.rb#L9-L18

PR #262, opened 2015-11-25
 (PCP-174) Acceptance - refactor cert paths

johnduarte:
This commit refactors the pxp-agent acceptance helper and tests
to use beaker's `tmpdir` helper method to define the location
to copy and use the test certificate files from.

This allows all operating systems to use a consistent file pathing.

[skip ci]


acceptance/tests/invalid_ssl_config.rb
```diff
@@ -59,8 +59,8 @@
 on agent1, puppet('resource service pxp-agent ensure=running')
 retry_on(agent1, "grep 'TLS handshake failed' #{logfile(agent1)}", {:max_retries => 30,
                                                              :retry_interval => 1})
-assert(on(agent1, "grep 'retrying in' #{logfile(agent1)}"),
-       "pxp-agent should log that it will retry connection")
+retry_on(agent1, "grep 'retrying in' #{logfile(agent1)}", {:max_retries => 10,
```
james-stocks: Not sure why we would need to retry here; as the test already waits for the prior log entry to appear. Retrying can only be helpful here though; to avoid spurious fails.
One comment on this is that retry_on means we lose the clearly defined assertion; we could consider a try/rescue with an explicit fail if retry_on fails

acceptance/lib/pxp-agent/config_helper.rb
```diff
@@ -201,6 +213,8 @@ def alt_ssl_cert_file(host, client_number)
     "#{ALT_SSL_BASE_DIR_WINDOWS}#{SSL_CERT_FILE_DIR_WINDOWS}client#{client_number}.alt.example.com.pem"
   when /osx/
     "#{ALT_SSL_BASE_DIR_OSX}#{SSL_CERT_FILE_DIR_POSIX}client#{client_number}.alt.example.com.pem"
+  when /solaris-10/
+    "#{ALT_SSL_BASE_DIR_SOL10}#{SSL_CERT_FILE_DIR_POSIX}client#{client_number}.alt.example.com.pem"
```
james-stocks: I suspect we can tidy up the code a bit; we have 5 repetitive case statements in this file now.
Maybe by adding a method base_dir(host, use_alt_path=false) ?

richardc:
Instead of adjusting this because of ~root (the home directory of the root user) moving from / to /root/ depending on distribution, can't we just push the 'POSIX' install of files to always explicitly just move the test certificates to a known path, so always `/test-resources`

james-stocks:
+1 on Richard's suggestion.

johnduarte:
@richardc I agree with the premise of your suggestion. After giving it more thought, I would suggest using beaker's `tmpdir` method to create a directory at run time as the base location for putting the `test-resources` files. This would remove the duplication and prevent us from leaving the `test-resources` files in the $HOME of the root user.

The `tmpdir` function also works on Windows.

acceptance/tests/invalid_ssl_config.rb
johnduarte: @james-stocks I ran into a race condition on AIX where this assertion would fail because the check had happened a couple of seconds before the expected line was written to the log.

richardc:
That sounds like a far better generalisation, :+1:

johnduarte:
@james-stocks , @richardc I have updated this PR with the refactor to use beaker's `tmpdir` method.

acceptance/tests/invalid_ssl_config.rb
```diff
@@ -4,10 +4,19 @@
 
 agent1 = agents[0]
 
+step 'Setup - Add base certs and config file'
+test_ssl_dir = agent1.tmpdir('test-ssl')
+scp_to(agent1, '../test-resources/ssl', test_ssl_dir)
+test_ssl_dir = File.join(test_ssl_dir, 'ssl')
+create_remote_file(agent1, pxp_agent_config_file(agent1), pxp_config_json_using_test_certs(master, agent1, 1, test_ssl_dir).to_s)
+if agent1['platform'] =~ /windows/
```
james-stocks: config_helper.rb has a windows? method to test if a host is Windows; it could/should be re-used here

acceptance/tests/invalid_ssl_config.rb
```diff
@@ -4,10 +4,19 @@
 
 agent1 = agents[0]
 
+step 'Setup - Add base certs and config file'
+test_ssl_dir = agent1.tmpdir('test-ssl')
```
james-stocks: Lines 8-14 are repeated in each test case now - can this be moved to a helper method? Perhaps configure_host_with_tmpdir_certs(host, cert_number)

james-stocks:
yo dawg i heard u like test code
so we put a PR in your PR
so u can test while u test

https://github.com/johnduarte/pxp-agent/pull/1

PR #264, opened 2015-11-25
(maint) Fix acceptance test not using keyword argument correctly

james-stocks:
The use_alt_file option for SSL config methods was being called like a named argument, but it was actually creating a local variable with the argument name rather than passing by keyword.
This commit changes the helper methods and the test to use an actual keyword argument.


richardc:
This ties us to a ruby 2.0 feature.  Is it a fair assumption that the tests in our pipelines will always use ruby 2.0 or newer?

james-stocks:
Closing - at least some CI Jenkins slaves are on Ruby 1.9.3. Not using Ruby 2.0 code until I can check with QE about Ruby versioning in CI.

PR #268, opened 2015-12-02
(PCP-191) Acceptance test for pxp-agent reconnecting after pcp-broker outage

james-stocks:


acceptance/tests/reconnect_after_broker_unavailable.rb
```diff
@@ -0,0 +1,38 @@
...
+    create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_test_certs(master, agent, i + 1, cert_dir).to_s)
+    on(agent, "rm -rf #{logfile(agent)}")
+    on agent, puppet('resource service pxp-agent ensure=running')
+    expect_file_on_host_to_contain(agent, logfile(agent), PXP_AGENT_LOG_ENTRY_ASSOCIATION_SUCCESS, 60)
+  end
+end
+
+step 'On master, stop the broker for at least 5 seconds' do
+  kill_pcp_broker(master)
+  sleep(5)
```
johnduarte: Can the kill method be modified to make sure the process is killed rather than putting a random sleep here?

`kill -0 <pid>` or setting a sleep loop in the method with a timeout threshold?

acceptance/tests/reconnect_after_broker_unavailable.rb
```diff
@@ -0,0 +1,38 @@
...
+    on agent, puppet('resource service pxp-agent ensure=stopped')
+    cert_dir = configure_std_certs_on_host(agent)
+    create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_test_certs(master, agent, i + 1, cert_dir).to_s)
+    on(agent, "rm -rf #{logfile(agent)}")
+    on agent, puppet('resource service pxp-agent ensure=running')
+    expect_file_on_host_to_contain(agent, logfile(agent), PXP_AGENT_LOG_ENTRY_ASSOCIATION_SUCCESS, 60)
+  end
+end
+
+step 'On master, stop the broker for at least 5 seconds' do
```
johnduarte: Is there something significant about the 5 second shutdown?

acceptance/tests/reconnect_after_broker_unavailable.rb
james-stocks: The sleep is to simulate the broker being unavailable for some amount of time. I guess there *shouldn't* be any difference between testing a 5 second outage and a 5ms outage. I'll remove the 5 seconds and retest

johnduarte:
:+1: lgtm validated on AIX

PR #269, opened 2015-12-03
(PCP-177) Add connection-timeout configuration option

ploubser:
Here we add the "connection-timeout" config parameter. This allows us to
set the timeout that cpp-pcp-client passes on to websocketpp so that
users can tweak on slow networks.

This requires a cpp-pcp-client update to work.


ploubser:
Dependent on https://github.com/puppetlabs/cpp-pcp-client/pull/122

PR #273, opened 2015-12-10
(maint) Use cpp-pcp-client master branch in CI

parisiale:
Updating Travis and AppVeyor scripts to use master instead of
cpp-pcp-client 1.0.1

### DIFF
--- a/.travis_target.sh
+++ b/.travis_target.sh
@@ -24,7 +24,6 @@ fi
 
 git clone https://github.com/puppetlabs/cpp-pcp-client
 cd cpp-pcp-client
-git checkout 1.0.1
 git submodule update --init --recursive
 cmake -DCMAKE_INSTALL_PREFIX=$USERDIR .
 make install -j2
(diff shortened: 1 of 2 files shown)


ploubser:
@richardc @james-stocks Are you happy with this change in testing behavior?

parisiale:
@richardc @james-stocks do you agree with this? It's meant to enable using cpp-pcp-client#master and merging #269 and #274.
We'll change the building configuration anyway, for PCP-209, so we prevent the hassle of tagging each time we need a change in cpp-pcp-client.

richardc:
We specifically pinned to 1.0.1 so that we wouldn't accidentally break pxp-agent by moving cpp-pcp-client's master branch underneath it.  If that's no longer a concern then it's fine, but I'd prefer that this component continued to use a released version of any dependency (openssl, cmake, cpp-pcp-client) in its testing.

If we want to test against both a release and a master branch of a dependency we could add it via an axis in a travis matrix.

PR #275, opened 2015-12-10
(PCP-213) Add test for file specs

johnduarte:
This commit adds an acceptance test to validate that the pxp-agent
installation conforms to the defined file layout specifications
as declared at
https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md

### DIFF
--- a/acceptance/tests/validate_file_paths.rb
+++ b/acceptance/tests/validate_file_paths.rb
@@ -1,0 +1,104 @@
+test_name 'C93828 - pxp-agent - File paths from spec'
+
+# verify that the installed file paths match those of the
+# puppet specification.
+# https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md
+#
+
+# {{{ helper methods
+def config_options(host)
+  platform = host[:platform]
+  ruby_arch = host[:ruby_arch]
+  type =  @options[:type]
+
+  case platform
+  when /windows/
+    if platform =~ /2003/
+      common_app_data = 'C:/Documents and Settings/All Users/Application Data'
+    else
+      common_app_data = 'C:/ProgramData'
+    end
+    if ruby_arch == 'x64'
+      ruby_arch = /-64/
+    else
+      ruby_arch = /-32/
+    end
+    if platform =~ ruby_arch
+      pupdir = 'C:/Program Files/Puppet Labs/Puppet'
+    else
+      pupdir = 'C:/Program Files (x86)/Puppet Labs/Puppet'
+    end
+
+    exdir = "#{pupdir}/pxp-agent"
+    binfile = "#{exdir}/bin/pxp-agent.exe"
+    moduledir = "#{exdir}/modules"
+    puppetlabs_data = "#{common_app_data}/PuppetLabs"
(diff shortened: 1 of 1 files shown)


acceptance/tests/validate_file_paths.rb
```diff
@@ -0,0 +1,106 @@
...
+      common_app_data = 'C:/Documents and Settings/All Users/Application Data'
+    else
+      common_app_data = 'C:/ProgramData'
+    end
+    if ruby_arch == 'x64'
+      ruby_arch = /-64/
+    else
+      ruby_arch = /-32/
+    end
+    if platform =~ ruby_arch
```
james-stocks: Does this work OK for the 32bit case? It looks like line 24 will set ruby_arch to /-32/ and then line 26 will evaluate to true (platform contains -32) and the 64bit Program Files folder will be used

acceptance/tests/validate_file_paths.rb
```diff
@@ -0,0 +1,106 @@
...
+agents.each do |agent|
+  config_options(agent).select {|v| !v[:not_path] }.each do |config_option|
+    path = config_option[:expected]
+    case config_option[:installed]
+    when :dir
+      on(agent, "test -d \"#{path}\"", :acceptable_exit_codes => [0, 1])  do |result|
+        if result.exit_code == 0 then
+          puts "Found expected directory '#{path}' on platform '#{agent[:platform]}' agent '#{agent}'"
+        else
+          puts "Failed to find  expected directory '#{path}' on platform '#{agent[:platform]}' agent '#{agent}'"
```
james-stocks: Need to fail at this point

acceptance/tests/validate_file_paths.rb
```diff
@@ -0,0 +1,106 @@
...
+        end
+      end
+    when :file
+      on(agent, "test -f \"#{path}\"", :acceptable_exit_codes => [0, 1])  do |result|
+        if result.exit_code == 0 then
+          puts "Found expected file '#{path}' on platform '#{agent[:platform]}' agent '#{agent}'"
+        else
+          puts "Failed to find  expected file '#{path}' on platform '#{agent[:platform]}' agent '#{agent}'"
```
james-stocks: Need to fail at this point

acceptance/tests/validate_file_paths.rb
johnduarte: (doh) left this in for debugging

acceptance/tests/validate_file_paths.rb
johnduarte: (doh) left this in for debugging

acceptance/tests/validate_file_paths.rb
johnduarte: My understanding is that if you are running a 32bit Windows, you are only going to have "Program Files" since you cannot install 64bit software on it. So, the "Program Files (x86)" directory is used when the ruby is 32bit, but the OS is 64bit.

james-stocks: You are correct. I wrongly thought that "platform" contained the rubyxXX string and you'd always match true against it. My bad, sorry.

PR #277, opened 2015-12-17
(PCP-196) Remove full path for systemctl in logrotate

underscorgan:
systemctl is either in /bin or /usr/bin depending on whether or not
you're running on deb-based systems. Those directories should always be
in your path so there's no need to specify the full path for systemctl.


kylog:
:+1:

PR #278, opened 2015-12-18
(maint) Fix brittle acceptance test regex

kylog:


kylog:
I'm putting this up blind, but this might fix the acceptance test failure on pxp-agent. /cc @parisiale or @james-stocks.

Btw, I believe this CI failure resulted from pxp-agent#stable bullding against cpp-pcp-client#master. Seems like we don't want that, but IIRC there was some discussion elsewhere on this and I don't know where that discussion landed. Thoughts?

kylog:
And the appveyor failure needs #279, or something along those lines.

PR #283, opened 2015-12-30
(PCP-198) Acceptance rakefile should support beaker-hostgenerator

james-stocks:
If TEST_TARGET ends with a, consider it a beaker-hostgenerator string.
If it does not contain an el7 master, add one (the Puppet component pipelines may ommit master).
Support for static host config files is still provided, they can be deleted once pipeline changes are complete.


acceptance/Rakefile
```diff
@@ -232,7 +235,20 @@ end
 
 def config
   if (ENV['TEST_TARGET']) && (not ENV['CONFIG'])
-    ENV['CONFIG'] = "config/nodes/#{ENV['TEST_TARGET']}.yaml"
+    if ENV['TEST_TARGET'][-1,1] == 'a'
```
kylog: I only did this in puppet as a super temporary change (because the b-hg changes weren't quite in place yet), and it's removed in https://github.com/puppetlabs/puppet/pull/4531.

I'd rather we had only one format for specifying test_targets to reduce confusion. Since pxp-agent already got switched to b-hg style test_targets, can we drop this and just assume it's b-hg style?

PR #284, opened 2015-12-31
(PCP-198) Acceptance should use TEST_TARGET ahead of CONFIG

james-stocks:
If Jenkins defines both TEST_TARGET and CONFIG for beaker host configuration then CONFIG should be ignored and TEST_TARGET should be used.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -233,16 +233,19 @@ def sha
   ENV['SHA']
 end
 
+# Return the host config file for beaker to use
+# If TEST_TARGET is set, pass it to beaker-hostgenerator to generate a file
+# else, assume CONFIG is set to the path of an existing host config file
 def config
-  if (ENV['TEST_TARGET']) && (not ENV['CONFIG'])
+  if (ENV['TEST_TARGET'])
     if !ENV['TEST_TARGET'].start_with?('redhat7-64m')
       ENV['TEST_TARGET'] = "redhat7-64m-#{ENV['TEST_TARGET']}"
-      puts "Adding EL7 master to ENV['TEST_TARGET']. Value is now '#{ENV['TEST_TARGET']}'"
+      puts "Adding Redhat 7 master to ENV['TEST_TARGET']. Value is now '#{ENV['TEST_TARGET']}'"
     end
     cli = BeakerHostGenerator::CLI.new([ENV['TEST_TARGET'], '--disable-default-role'])
     ENV['CONFIG'] = "tmp/#{ENV['TEST_TARGET']}-#{SecureRandom.uuid}.yaml"
     FileUtils.mkdir_p('tmp')
-    File.open(config, 'w') do |fh|
+    File.open(ENV['CONFIG'], 'w') do |fh|
       fh.print(cli.execute)
     end
   end


acceptance/Rakefile
```diff
@@ -233,16 +233,19 @@ def sha
...
+  if (ENV['TEST_TARGET'])
     if !ENV['TEST_TARGET'].start_with?('redhat7-64m')
       ENV['TEST_TARGET'] = "redhat7-64m-#{ENV['TEST_TARGET']}"
-      puts "Adding EL7 master to ENV['TEST_TARGET']. Value is now '#{ENV['TEST_TARGET']}'"
+      puts "Adding Redhat 7 master to ENV['TEST_TARGET']. Value is now '#{ENV['TEST_TARGET']}'"
     end
     cli = BeakerHostGenerator::CLI.new([ENV['TEST_TARGET'], '--disable-default-role'])
     ENV['CONFIG'] = "tmp/#{ENV['TEST_TARGET']}-#{SecureRandom.uuid}.yaml"
     FileUtils.mkdir_p('tmp')
-    File.open(config, 'w') do |fh|
```
rick: Was this originally doing recursion? What am I missing here?

Regardless, the new version makes more sense.

rick:
I don't have 100% context, but textually, from a ruby perspective, I'm :+1:.

acceptance/Rakefile
```diff
@@ -233,16 +233,19 @@ def sha
...
+    File.open(ENV['CONFIG'], 'w') do |fh|
```
er0ck: what happens when this ENV is empty?  or the file doesn't exist? or someone sets it to /dev/zero?

kylog:
Yeah, agreed, this is cleaner and gives priority to TEST_TARGET over CONFIG, which is the pattern in the other puppet-agent components.

:+1:

acceptance/Rakefile
kylog: In this case, the ENV is set two lines above, and this is the logic to write that desired file out. So the intent is that the file doesn't exist.

But to riff on that question: there's a good concern about error handling in general, e.g. when a developer specifies CONFIG on the invocation line for local testing. I assume beaker will fail dramatically in any of those cases. It would be slick if, perhaps as part of refactoring/commonizing this logic across projects/branches (rototiller is my new dream for a brighter future :)), if we had graceful handling of such scenarios with helpful error messages.

acceptance/Rakefile
```diff
@@ -233,16 +233,19 @@ def sha
   ENV['SHA']
 end
 
+# Return the host config file for beaker to use
+# If TEST_TARGET is set, pass it to beaker-hostgenerator to generate a file
+# else, assume CONFIG is set to the path of an existing host config file
 def config
-  if (ENV['TEST_TARGET']) && (not ENV['CONFIG'])
+  if (ENV['TEST_TARGET'])
```
er0ck: ok.  so it turns out these jobs are sending both CONFIG and TEST_TARGET.
this should probably be the same as facter's current work around:
```
def hosts
  if ENV['HOSTS']
    ENV['HOSTS']
  else
    puts 'Warning: environment variable CONFIG deprecated. Please use HOSTS to match beaker options.'
    ENV['CONFIG']
  end
end
[...]

 # The 'if ...' suffix conditional below is a temporary measure which should go
  # away once the pipeline changes are made so that TEST_TARGET is always
  # intended for use with beaker-hostgenerator. So please remove that
  # conditional (and this comment) once that pipeline change is made.
  target = ENV['TEST_TARGET'] if ENV['TEST_TARGET'] && ENV['TEST_TARGET'][-1,1] == 'a'
  if target
    cli = BeakerHostGenerator::CLI.new([target, '--disable-default-role', '--osinfo-version', '1'])
    ENV['HOSTS'] = "tmp/#{target}-#{SecureRandom.uuid}.yaml"
    FileUtils.mkdir_p('tmp')
    File.open(hosts, 'w') do |fh|
      fh.print(cli.execute)
    end
    config_opt = "--hosts=#{hosts}"
  elsif hosts
    config_opt = "--hosts=#{hosts}"
  end
```

acceptance/Rakefile
er0ck: yeah.  took me a while to catchup on this logic but i see it now.
re: rototiller and ENV checking:
https://confluence.puppetlabs.com/display/QA/Rototiller#Rototiller-asataskauthor

PR #285, opened 2016-01-06
(PCP-233) Add Arista support

LuvCurves:


waynr:
:+1:

joshcooper:
Do we need to install the netdev module on the master like we do for puppet? I'm not sure if that's needed to exercise pxp-agent on arista? or if that is even supposed to work?

LuvCurves:
@joshcooper No need to install module on master unless we are doing something Arista specific.

PR #289, opened 2016-01-12
(PCP-237) Expand modules-dir and modules-config-dir paths

parisiale:


lib/src/configuration.cc
```diff
@@ -613,42 +613,58 @@ void Configuration::checkUnconfiguredMode() {
...
-            if (fs::create_directories(tmp_path)) {
-                is_writable = true;
-                fs::remove_all(tmp_path);
-            }
-        } catch (const std::exception& e) {
-            LOG_DEBUG("Failed to create the test dir in spool path during"
-                      "configuration validation: %1%", e.what());
+        if (!fs::exists(val_path)) {
+            throw Configuration::Error { option + " '" + val_path.string()
+                + "'' does not exists" };
```
MikaelSmith: Should probably say "does not exist".

MikaelSmith:
:+1: There's a spelling choice I don't agree with, but it existed before. We can fix it up in a separate PR if you agree.

PR #290, opened 2016-01-12
(PCP-238) Fix suffix used to match module config

parisiale:
As indicated in the docs and in puppet-specifications, the suffix for
module configuration files must be ".conf", not ".cfg".


MikaelSmith:
Do we have any plans for testing module configuration files?

parisiale:
We validate the contents of a module config file by using the JSON schema provided by the module itself. I'd say that this calls for system tests. Yeah, this should be ticketed.

PR #294, opened 2016-01-15
(PCP-240) Formalize and extend external modules' interface

parisiale:
Here we formalize and extend the operational requirements for I/O,
introduce the metadata schema, and rename a few metadata entries.

The interface for external modules is here extended in order to allow
pxp-agent to put the modules in charge of writing their output to file.

[skip ci]

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -1,58 +1,168 @@
+# Module requirements
 
-# PXP agent architecture
+### Metadata
 
-map of name:module, possibly factory style from peg
+The `metadata` of a module is a JSON object that contains a number of JSON
+schemas for specifying configuration options and actions. It contains:
 
-Module abstraction
+ - **configuration**: (optional) schema that describes the module configuration format;
+ - **actions**: an array where each item is an object that describes an action implemented by the module (please, refer to the below schema).
 
-    class Module {
-        std::string name;
-        bool invoke(const CthunMessage& request, CthunMessage& response);
-        Schema actions;
+The `metadata` schema is:
+
+    {
+        "type" : "object",
+        "description" : "Metadata format",
+        "properties" : {
+            "actions" : {
+                "type" : "array",
+                "description" : "List of actions supported by this module",
+                "items" : {
+                    "type" : "object",
+                    "properties" : {
+                        "description" : {
+                            "type" : "string",
+                            "description" : "Describes an action",
+                        },
+                        "name" : {
+                            "type" : "string",
+                            "description" : "Name of the action",
+                        },
+                        "input" : {
+                            "type" : "object",
+                            "description" : "Schema for the PXP request arguments",
+                        },
+                        "results" : {
+                            "type" : "object",
+                            "description" : "Schema for the action results",
+                        },
+                    },
+                    "required" : ["description", "name", "input", "results"],
+                }
+            },
+            "configuration" : {
+                "type" : "object",
+                "description" : "Schema for the module configuration"
+            },
+        },
+        "required" : ["actions"],
+        "additionalProperties" : false,
     }
 
+Example of `metadata`:
 
-## External modules
-
-Load from libdir
-
-Will have 1 required action, metadata, takes no parameters, yields a
-json description of the actions it has.  Parameters subkey should be a
-json-schema document the deamon will validate:
-
-    ./package metadata
     {
-        "description": "Manage packages",
-        "actions": [
-            {
-                "name": "install",
-                "input": {
-                    "type": "object",
-                    "properties": {
-                        "name": {
-                            "type": "string"
-                        },
-                        "version": {
-                            "type": "string"
-                        },
-                    },
-                    "required": [ "name" ]
+        "description" : "Metadata of the reverse module, used for testing",
+        "configuration" : {
+            "type" : "object",
+            "properties" : { "spam_dir" : { "type" : "string" } },
+            "required" : ["spam_dir"],
+            "additionalProperties" : false,
+        },
+        "actions" : [
+          { "name" : "string",
+            "description" : "Reverses a string",
+            "input" : {
+                "type" : "object",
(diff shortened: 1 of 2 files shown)


modules/README.md
```diff
@@ -1,58 +1,164 @@
...
+                        "input" : {
+                            "type" : "object",
+                            "description" : "Schema for the PXP request arguments",
+                        },
+                        "results" : {
+                            "type" : "object",
+                            "description" : "Schema for the action results",
+                        },
+                    },
+                    "required" : ["description", name", "input", "results"],
```
mruzicka: Either the ```description``` doesn't belong here, ore there must be a ```description``` property.

modules/README.md
```diff
@@ -1,58 +1,164 @@
...
+                        "description" : "File for storing the output on stderr"
                     },
-                    "required": [ "success" ]
-                }
+                    "exitcode" : {
+                        "type" : "string",
+                        "description" : "File for storing the action's exitcode"
+                    },
+                },
+                "required" : ["out", "err", "exitcode"],
```
mruzicka: out -> stdout, err -> stderr

modules/README.md
mruzicka: Missing double quote before ```name```.

modules/README.md
```diff
@@ -1,58 +1,164 @@
...
 
+Each action should write its processing results on stdout. Error messages should
+go on stderr.
 
-Invocation of an action will supply parameters as a document on stdin
+If the input object parsed from stdin contains the `output_files` object, the
+action must ALSO write its results, error messages, and exit code in the
+specified files (respectively: `stdout`, `stderr`, and `exitcode`). In this
+case, pxp-agent guarantees that those files will be read only after the
+`exitcode` file is created; the existence of `exitcode` will be used to indicate
```
mruzicka: Since the {{exitcode}} file is also used for storing the actual exit code of the module, there is a race condition between creating the file and writing the exit code to it.
To avoid it the each module needs to write its exit status to a file with a different name and then rename it to the name mandated by the ```exitcode``` parameter.
Alternatively we could specify yet another file (other than the ```exitcode``` one) to be used as the indication of the completion of the module execution.

modules/README.md
parisiale: Right, I thought it may get confusing.
This JSON object is meant to be a schema for `metadata`, os it's basically a schema containing other schemas ("configuration and, for each action, "input" and "results" are all JSON schemas).
So, on one hand, we're not specifying the structure of each required schema; we're just stating that we need those entries. On the other, we're using a schema for the whole metadata structure - from http://json-schema.org/example1.html:
'''
the title and description keywords are descriptive only, in that they do not add constraints to the data being validated. 
'''

It may be the best way to specify the `metadata` structure, but in this case I believe that using `description` is ok.

modules/README.md
parisiale: Yep, that's formally correct, but that's not the only indication of the action execution that pxp-agent uses. Consider that:

- pxp-agent will normally get the outcome of the action run directly by the leatherman::execution interface, since it must validate the results output and, if requested, send a final non-blocking response;
- only in case pxp-agent stops executing during an action run it will later rely on the existence of the exitcode file to determine if the action has made available its outcome;
- besides the pid file check and the current approach based on named mutexes to prevent (bigger) race conditions between leatherman::execution and onMessage handlers, pxp-agent can easily prevent concurrency issues use the existence of the file to determine the action execution state, then parse / validate stdout, process stderr, and finally get the exitcode int.

I believe this is purely an implementation issue; the above requirements (write on files) and guarantees (pxp-agent will check exitcode's existence before reading the rest) are enough to enable a safe workflow. Anyway, we have PCP-208 in program to improve the whole mechanism; things will change and it's not worth mandating a fourth file now.

modules/README.md
mruzicka: Maybe I'm still not getting it, but it feels like you should have the ```description``` property along the ```name```, ```input``` & ```output```:
```JSON
                    "properties" : {
                         "name" : {
                             "type" : "string",
                             "description" : "Name of the action",
                         },
                         "description" : {
                             "type" : "string",
                             "description" : "Description of the action",
                         },
                         "input" : {
                             "type" : "object",
                             "description" : "Schema for the PXP request arguments",
                         },
                         "results" : {
                             "type" : "object",
                             "description" : "Schema for the action results",
                         },
                     },
```

modules/README.md
mruzicka: This is really a corner case, so if we plan to improve it ... it's OK.

modules/README.md
mruzicka: and as far as I can tell, it's what the current ```pxp-module-puppet``` produces:
```JSON
        {
            "description": "Start a Puppet run",
            "input": {
                "properties": {
                    "env": {
                        "type": "array"
                    },
                    "flags": {
                        "type": "array"
                    }
                },
                "required": [
                    "env",
                    "flags"
                ],
                "type": "object"
            },
            "name": "run",
            "output": {
                "properties": {
                    "environment": {
                        "type": "string"
                    },
                    "error": {
                        "type": "string"
                    },
                    "error_type": {
                        "type": "string"
                    },
                    "exitcode": {
                        "type": "number"
                    },
                    "kind": {
                        "type": "string"
                    },
                    "status": {
                        "type": "string"
                    },
                    "time": {
                        "type": "string"
                    },
                    "transaction_uuid": {
                        "type": "string"
                    },
                    "version": {
                        "type": "number"
                    }
                },
                "required": [
                    "kind",
                    "time",
                    "transaction_uuid",
                    "environment",
                    "status",
                    "exitcode",
                    "version"
                ],
                "type": "object"
            }
        }
```

parisiale: :+1:

mruzicka:
LGTM

PR #295, opened 2016-01-15
(PCP-188) Acceptance test of pxp-module-puppet run

james-stocks:


james-stocks:
There's a shortcoming in the rpc_blocking_request helper method - it accepts multiple targets to send the RPC to, but returns when one response is received. It should either only accept one target; or wait for and return all responses; or yardoc the current limitation

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -50,3 +51,99 @@ def kill_pcp_broker(host)
...
+      loop do
+        mutex.synchronize do
+          have_response.wait(mutex)
+          done = have_all_rpc_responses?(targets, responses)
+        end
+        break if done
+      end
+    end
+  rescue Timeout::Error
+    if !have_all_responses?
```
parisiale: To check have_all_responses() again, the mutex should be acquired.

james-stocks: check_all_responses? is just a quick comparison between a Ruby array and hash, and is a separate method (rather than inline code here) just for readability. Does it really need to participate in the mutex?

parisiale: You need to acquire the mutex to call check_all_responses() because, after the timeout expiration, you may have concurrent message handler that modifies the `responses` hash.

james-stocks: I believe the calls to synchronize at lines 84 and 122 will lock the mutex and prevent on_message from executing when have_all_rpc_responses? is reading the responses hash

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -50,3 +51,102 @@ def kill_pcp_broker(host)
...
+# @param broker the hostname or beaker host object of the machine running PCP broker
+# @param targets array of PCP identities to send request to
+#                e.g. ["pcp://client01.example.com/agent","pcp://client02.example.com/agent"]
+# @param pxp_module which PXP module to call, default pxp-module-puppet
+# @param action which action in the PXP module to call, default run
+# @param params params to send to the module. e.g for pxp-module-puppet:
+#               {:env => [], :flags => ['--noop', '--onetime', '--no-daemonize']}
+# @return hash of responses, with target identities as keys, and value being the response message for that identity as a hash
+# @raise String indicating something went wrong, test case can use to fail
+def rpc_blocking_request(broker, targets,
```
er0ck: couldn't you just use:
https://github.com/puppetlabs/ruby-pcp-client/blob/fdd6eaa9a37a0919dea1fd24c556daf5d22209a1/lib/pcp/client.rb#L44

instead of replicating a bunch of that functionality here?
i might be missing some functionality you need that isn't in that client.

acceptance/tests/pxp-module-puppet/run_puppet.rb
```diff
@@ -0,0 +1,49 @@
+require 'pcp/client'
+require 'pxp-agent/test_helper.rb'
+
+test_name 'C95972 - pxp-module-puppet run'
```
er0ck: test_name blocks should have do..end pairs now as well.

acceptance/tests/pxp-module-puppet/run_puppet.rb
```diff
@@ -0,0 +1,49 @@
...
+
+step 'Ensure each agent host has pxp-agent running and associated' do
+  agents.each_with_index do |agent, i|
+    on agent, puppet('resource service pxp-agent ensure=stopped')
+    cert_dir = configure_std_certs_on_host(agent)
+    create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_test_certs(master, agent, i + 1, cert_dir).to_s)
+    on agent, puppet('resource service pxp-agent ensure=running')
+    expect_file_on_host_to_contain(agent, logfile(agent), PXP_AGENT_LOG_ENTRY_ASSOCIATION_SUCCESS, 60)
+  end
```
er0ck: 60, magic number, coule be in explanatory variable name

acceptance/tests/pxp-module-puppet/run_puppet.rb
```diff
@@ -0,0 +1,49 @@
+require 'pcp/client'
```
johnduarte: This require is not needed, since it is in and used by the helper.

acceptance/lib/pxp-agent/test_helper.rb
james-stocks: The PCP client.rb you linked to does not provide what is implemented in this method.

Potentially there could be a PXP client gem that implements this method plus other PXP operations, but that doesn't exist currently.

james-stocks:
Closing - made a new branch to squash the commits for the fixes in this PR, will open a new PR for that branch...

PR #297, opened 2016-01-19
(PCP-198) Acceptance pre-suite improvements

james-stocks:
Remove references to "package" test setup in Rakefile.
Use beaker to install puppet-agent instead of lib code in this repo.
Have agent nodes sign themselves against puppet-server on the master node.

### DIFF
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -1,9 +1,14 @@
 {
   :type => 'aio',
+  :is_puppetserver => true,
+  :puppetservice => 'puppetserver',
+  :'puppetserver-confdir' => '/etc/puppetlabs/puppetserver/conf.d',
   :pre_suite => [
     'setup/common/000-delete-puppet-when-none.rb',
     'setup/aio/pre-suite/010_Install.rb',
     'setup/aio/021_InstallAristaModule.rb',
     'setup/common/010_Setup_Broker.rb',
+    'setup/common/035_StartPuppetServer.rb',
+    'setup/common/040_ValidateSignCert.rb',
   ],
 }
--- a/acceptance/lib/puppet/acceptance/common_utils.rb
+++ b/acceptance/lib/puppet/acceptance/common_utils.rb
@@ -6,5 +6,30 @@ module Puppet
       end
       module_function :ruby_command
     end
-  end
-end
+
+    module CAUtils
+      def initialize_ssl
+        agents.each do |agent|
+          next if agent == master
+
+          ssldir = on(agent, puppet('agent --configprint ssldir')).stdout.chomp
+          on(agent, "rm -rf '#{ssldir}'")
+
+          step "Agents: Run agent --test first time to gen CSR"
+          on agent, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [1]
+        end
+
+        step "Master: sign all certs"
+        on master, puppet("cert --sign --all"), :acceptable_exit_codes => [0,24]
+
+        agents.each do |agent|
+          next if agent == master
+
+          step "Agents: Run agent --test second time to obtain signed cert"
+          on agent, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [0,2]
+        end
+      end # initialize_ssl
+
+    end # Module CAUtils
+  end # Module Acceptance
+end # Module Puppet
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -72,7 +72,7 @@ module HarnessOptions
   end
 end
 
-def beaker_test(mode = :packages, options = {})
+def beaker_test(mode = :aio, options = {})
   final_options = HarnessOptions.options(mode, options)
 
   if mode == :git
@@ -222,7 +222,7 @@ def print_preserved(preserved)
 end
 
 def beaker_run_type
-  type = ENV['TYPE'] || :packages
+  type = ENV['TYPE'] || :aio
   type = type.to_sym
 end
 
@@ -273,14 +273,6 @@ If there is a Beaker options hash in a ./local_options.rb, it will be included. 
 EOS
 
     desc <<-EOS
-Run the acceptance tests through Beaker and install packages on the configuration targets.
-#{USAGE}
-EOS
-    task :packages => 'ci:check_env' do
-      beaker_test
-    end
-
-    desc <<-EOS
 Run the acceptance tests through Beaker and install packages as part of the AIO puppet-agent installation.
 #{USAGE}
 EOS
@@ -304,7 +296,7 @@ EOS
 
   desc <<-EOS
 Run an acceptance test for a given node configuration and preserve the hosts.
-Defaults to a packages run, but you can set it to 'git' with TYPE='git'.
+Defaults to an AIO run, but you can set it to 'git' with TYPE='git'.
 #{USAGE}
   EOS
   task :test_and_preserve_hosts => 'ci:check_env'  do
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -6,23 +6,40 @@ extend Beaker::DSL::InstallUtils
 
 test_name "Install Packages"
 
-step "Install repositories on target machines..." do
-  sha = ENV['SHA']
-  hosts.each do |host|
-    install_repos_on(host, 'puppet-agent', sha, 'repo-configs')
+step "Install puppet-agent..." do
+  opts = {
+    :puppet_collection    => 'PC1',
+    :puppet_agent_sha     => ENV['SHA'],
+    :puppet_agent_version => ENV['SUITE_VERSION'] || ENV['SHA']
+  }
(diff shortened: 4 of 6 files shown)


joshcooper:
:+1:

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,111 @@ def ruby_command(host)
...
     end
-  end
-end
+
+    module CAUtils
+      def initialize_ssl
+        hostname = on(master, 'facter hostname').stdout.strip
+        fqdn = on(master, 'facter fqdn').stdout.strip
+
+        if master.use_service_scripts?
```
joshcooper: You may need to specify to use service scripts like we do in puppet's presuite: https://github.com/puppetlabs/puppet/blob/master/acceptance/setup/aio/pre-suite/015_PackageHostsPresets.rb

acceptance/config/aio/options.rb
```diff
@@ -1,9 +1,13 @@
...
   :type => 'aio',
+  :is_puppetserver => true,
+  :puppetservice => 'puppetserver',
+  :'puppetserver-confdir' => '/etc/puppetlabs/puppetserver/conf.d',
   :pre_suite => [
     'setup/common/000-delete-puppet-when-none.rb',
     'setup/aio/pre-suite/010_Install.rb',
     'setup/aio/021_InstallAristaModule.rb',
     'setup/common/010_Setup_Broker.rb',
+    'setup/common/040_ValidateSignCert.rb'
```
joshcooper: I don't see this step in the PR, do you need to `git add` it?

acceptance/lib/puppet/acceptance/common_utils.rb
james-stocks: Hi @joshcooper - I've copied the Puppet pre-suite file here; but I'm unsure what the context of the phrase "use service scripts" is? What does it mean, and what are the cases for using it or not using it? Can I simplify by not having 015_PackageHostsPresets.rb and making the code here unconditional (i.e. always stop puppetserver at this point) or does that cause a problem in some cases?

acceptance/config/aio/options.rb
james-stocks: Ouch, thanks for catching that. I've added it then squashed the commit so no-one will ever know

acceptance/lib/puppet/acceptance/common_utils.rb
joshcooper: :+1: to making the code unconditional.

The term "use service scripts" refers to whether there are service scripts for start/stopping the puppet server, and is implemented in beaker: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host.rb#L176. Based on the return value, beaker will start/stop the puppet master differently, e.g. https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/puppet_helpers.rb#L165, such as if you're using `with_puppet_running_on`. But yeah, I wouldn't worry about that.

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,111 @@ def ruby_command(host)
...
+          step "Ensure puppet is stopped"
+          # Passenger, in particular, must be shutdown for the cert setup steps to work,
+          # but any running puppet master will interfere with webrick starting up and
+          # potentially ignore the puppet.conf changes.
+          on(master, puppet('resource', 'service', master['puppetservice'], "ensure=stopped"))
```
joshcooper: Since you're always running against `puppetserver` you can change `master['puppetservice']` to just be `puppetserver`. Puppet had to account that the service name was sometimes `puppet` (webrick), `puppetserver` (JVM puppetserver), `apache2` or `httpd` (passenger-based puppet master).

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,111 @@ def ruby_command(host)
...
+
+            step "Agents: Run agent --test second time to obtain signed cert"
+            on agents, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [0,2]
+          end
+        end
+      end
+
+      def clean_cert(host, cn, check = true)
+        if host == master && master[:is_puppetserver]
+            on master, puppet_resource("service", master['puppetservice'], "ensure=stopped")
```
joshcooper: since you're always running against `puppetserver` you can drop the `is_puppetserver` business, and hardcode the service name as `puppetserver`.

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,111 @@ def ruby_command(host)
...
+      def reset_agent_ssl(resign = true)
+        return if master.is_pe?
+        clear_agent_ssl
+
+        hostname = master.execute('facter hostname')
+        fqdn = master.execute('facter fqdn')
+
+        step "Clear old agent certificates from master" do
+          agents.each do |agent|
+            next if agent == master && agent.is_using_passenger?
```
joshcooper: you can remove this, passenger is so 2015.

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,111 @@ def ruby_command(host)
...
+          step "Master: Ensure the master is listening and autosigning"
+          with_puppet_running_on(master,
+                                  :master => {
+                                    :dns_alt_names => "puppet,#{hostname},#{fqdn}",
+                                    :autosign => true,
+                                  }
+                                ) do
+
+            agents.each do |agent|
+              next if agent == master && agent.is_using_passenger?
```
joshcooper: same here.

james-stocks:
@joshcooper - thanks very much for the guidance - I pushed up the changes you suggested. Since it falls under master/agent setup, I fixup'd into the previous commit

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,108 @@ def ruby_command(host)
...
+            step "Master: sign all certs"
+            on master, puppet("cert --sign --all"), :acceptable_exit_codes => [0,24]
+
+            step "Agents: Run agent --test second time to obtain signed cert"
+            on agents, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [0,2]
+          end
+        end
+      end
+
+      def clean_cert(host, cn, check = true)
```
joshcooper: You should be able to delete this method, along with `clear_agent_ssl` and `reset_agent_ssl` below. Those methods are needed in puppet, because we do things like change the agent's ssl client cert during the test run, and then restore the original cert. But since pxp-agent just needs to create the initial ssl certs at the beginning and never touch them, these methods can be deleted.

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,5 +6,108 @@ def ruby_command(host)
...
+              step "Agents: Run agent --test first time to gen CSR"
+              on host, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [1]
+            end
+
+            # Sign all waiting certs
+            step "Master: sign all certs"
+            on master, puppet("cert --sign --all"), :acceptable_exit_codes => [0,24]
+
+            step "Agents: Run agent --test second time to obtain signed cert"
+            on agents, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [0,2]
```
joshcooper: I think if you move `050_StartPuppetServer.rb` to be `035_StartPuppetServer.rb` so that it comes before `040_ValidateSignCert.rb`, then the puppetserver will generate a cert/key of its own as part of its normal installation process. Then you shouldn't need to wipe out the SSL directory on the master host and start webrick (via `with_puppet_running_on`) to regenerate the master's SSL cert.

So then this method can be a shorter version of what you have:

```ruby
agents.each do |agent|
  next if agent == master

  ssldir = on(agent, puppet('agent --configprint ssldir')).stdout.chomp
  on(agent, "rm -rf '#{ssldir}'")
  
  step "Agents: Run agent --test first time to gen CSR"
  on agent, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [1]
end

step "Master: sign all certs"
on master, puppet("cert --sign --all"), :acceptable_exit_codes => [0,24]

agents.each do |agent|
  next if agent == master

  step "Agents: Run agent --test second time to obtain signed cert"
  on agent, puppet("agent --test --server #{master}"), :acceptable_exit_codes => [0,2]
end
```

james-stocks:
@joshcooper I pushed up further changes according to your last comment; can you review again please?

joshcooper:
One unrelated issue, pxp-agent/acceptance/Gemfile sets the [default GEM_SOURCE to our internal mirror](https://github.com/puppetlabs/pxp-agent/blob/master/acceptance/Gemfile#L3), which is different than our other components that default to [rubygems.org](https://github.com/puppetlabs/puppet/blob/master/acceptance/Gemfile#L3).

joshcooper:
:+1:

PR #303, opened 2016-01-22
(maint) Improve usage of ActionRequest in RequestProcessor

parisiale:
- adding results_dir_ to ActionRequest;
- tidying up RequestProcessor and ResultsStorage.

### DIFF
--- a/lib/src/request_processor.cc
+++ b/lib/src/request_processor.cc
@@ -47,12 +47,13 @@ class ResultsStorage {
 
     // Throw a ResultsStorage::Error in case of failure while writing
     // to any of result files
-    ResultsStorage(const ActionRequest& request, const std::string& results_dir)
+    ResultsStorage(const ActionRequest& request)
             : module { request.module() },
               action { request.action() },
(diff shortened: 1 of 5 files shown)


lib/src/request_processor.cc
```diff
@@ -133,12 +133,12 @@ void nonBlockingActionTask(std::shared_ptr<Module> module_ptr,
 
     try {
         ResultsMutex::LockGuard a_l { ResultsMutex::Instance().access_mtx };
-        mtx_ptr = ResultsMutex::Instance().get(transaction_id);
+        mtx_ptr = ResultsMutex::Instance().get(request.transactionId());
```
mruzicka: should the change from ```transaction_id``` to ```request.transactionId()``` go into a separate commit?

mruzicka:
@parisiale except for the single comment this looks good to me. If you think another commit is not necessary, I'm happy to merge as is.

lib/src/request_processor.cc
parisiale: No. This is about "Rely on ActionRequest".

PR #305, opened 2016-01-25
(PCP-252) Let external modules write their output

parisiale:
With this PR, we rely on external modules for writing action output to file,
in case of non-blocking requests.

This is achieved by using the new interface for external modules.

ExternalModule now uses the leatherman::execution::execute variant
that only processes PID, without processing stdout/stderr.

### DIFF
--- a/lib/tests/resources/modules/reverse_valid
+++ b/lib/tests/resources/modules/reverse_valid
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
 require 'json'
+require_relative './check_output'
 
 def action_metadata
    metadata = {
@@ -47,7 +48,7 @@ def action_metadata
         :input => {
           :type => "object",
           :properties => {
-            :input => {
+            :the_input => {
               :type => 'string',
             },
           },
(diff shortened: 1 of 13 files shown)


lib/tests/resources/modules/reverse_valid
```diff
@@ -73,14 +74,18 @@ def action_metadata
...
+  puts results.to_json
 end
 
 def action_hash
-  hash = JSON.load($stdin)["params"]
-  hash['output'] = hash['input'].reverse
+  args = JSON.load($stdin)
+  check_output_files(args)
+  input_args = args["input"]
+  hash['output'] = input_args['the_input'].reverse
```
mruzicka: ```hash``` is undefined at this point

mruzicka: on top of that the ```resutls``` schema mandates ```output``` *and* ```input``` fields in the result: https://github.com/parisiale/pxp-agent/blob/352ff9f3d2dff47b5f6b18ea05be7434ebee5d49/lib/tests/resources/modules/reverse_valid#L60-L62

parisiale: Thanks, updated.

mruzicka:
Other then the problems in the ruby test helper this LGTM. Is the hash action in the helper used during the tests at all?

parisiale:
Not currently, as we were not getting any error, but I may use it for PCP-239.

mruzicka:
LGTM

PR #308, opened 2016-01-26
(maint) Acceptance - Workaround for PCP-253 (PCP client.connect)

james-stocks:
When ruby-pcp-client has client.connect() called with the default timeout (0), it might fail due to thread logic.
The workaround (this commit) is to specify a non-zero timeout value.
Also add checking of the result of connect and raise an exception if it doesn't return true.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -93,7 +93,10 @@ def rpc_blocking_request(broker, targets,
     end
   end
 
-  client.connect()
+  if !client.connect(5)
+    raise "Controller PCP client failed to connect with pcp-broker on #{broker}"
+  end
+  
   if !client.associated?
     raise "Controller PCP client failed to associate with pcp-broker on #{broker}"
   end


parisiale:
You can append a "[skip ci]" line to the commit message (the last commit, in case you're making a PR for multiple commits) to avoid triggering Travis and Appveyor.

parisiale:
It would prefer to have an indication of the changes you're submitting in the commit title instead of a ticket reference, which can got in the lines below.

er0ck:
FYI, i have a prepare-commit-msg git hook which appends [skip ci] if it thinks it's an acceptance PR:
https://github.com/er0ck/dotfiles/blob/master/git/git_template/hooks/prepare-commit-msg

(it does some other things like tries to prepare the title from the branch name)

er0ck:
works for me.  :+1: 
looks like acceptance bundle requires ruby 2+, unrelated to this change.

er0ck:
@james-stocks  ugh... this is merged and works, but i just noticed your branch is named stable.  did you intend this for stable?  if so, should probably back-port this to stable.

richardc:
It was mistargeted.  Luckily though we've been merging stable into master fairly regularly recently, so this didn't bring across anything too unexpected.

```
*   079b7fc (HEAD -> master, origin/master, origin/HEAD) Merge pull request #308 from james-stocks/stable_fix
|\
| * fb6c207 (maint) Acceptance - Workaround for PCP-253 (PCP client.connect)
* | d2eca4b (packaging) Bump 'master' version to 1.1.0
* |   4f11d49 (maint) Merge branch 'stable' into 'master'
|\ \
| |/
| * 2b73085 (origin/stable) (packaging) Bump version to 1.0.3
```

I raised #310 to merge the commit to stable, where it should have been aimed.

er0ck:
:hand: :five:

PR #311, opened 2016-01-27
(maint) Acceptance - Workaround for PCP-253 (PCP client.connect)

parisiale:
When ruby-pcp-client has client.connect() called with the default
timeout (0), it might fail due to thread logic.
The workaround (this commit) is to specify a non-zero timeout value.
Also add checking of the result of connect and raise an exception
if it doesn't return true.


parisiale:
Closing as duplicate

PR #312, opened 2016-01-28
(PCP-251) support for saving stdout/stderr/exitcode to files

mruzicka:
The pxp-module-puppet's run action can now save its standard output / standard error / exit code to files specified in the action's input.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -51,9 +51,9 @@ def disabled?(run_result)
   !!(run_result =~ /disabled(.*?)Use 'puppet agent --enable' to re-enable/m)
 end
 
-def make_environment_hash(params)
+def make_environment_hash(action_input)
   env_hash = {}
-  params["env"].each do |entry|
+  action_input["env"].each do |entry|
     key, value = entry.split('=')
     env_hash[key] = value
   end
@@ -61,9 +61,9 @@ def make_environment_hash(params)
   return env_hash
 end
 
-def make_command_array(config, params)
+def make_command_array(config, action_input)
   cmd_array = [config["puppet_bin"], "agent"]
-  cmd_array +=  params["flags"]
+  cmd_array +=  action_input["flags"]
   return cmd_array
 end
 
@@ -76,6 +76,10 @@ end
 
 def get_result_from_report(exitcode, config, start_time)
   last_run_report = check_config_print("lastrunreport",config)
+  if last_run_report.empty?
+    return make_error_result(exitcode, Errors::NoLastRunReport,
+                             "could not find the location of the last run report")
+  end
 
   if !File.exist?(last_run_report)
     return make_error_result(exitcode, Errors::NoLastRunReport,
@@ -91,9 +95,9 @@ def get_result_from_report(exitcode, config, start_time)
                              "#{last_run_report} could not be loaded: #{e}")
   end
 
-  run_result = last_run_result(exitcode)
-
-  if exitcode != 0
+  if exitcode == 0
+    run_result = last_run_result(exitcode)
+  else
     run_result = make_error_result(exitcode, Errors::NonZeroExit,
                                    "Puppet agent exited with a non 0 exitcode")
   end
@@ -109,10 +113,10 @@ def get_result_from_report(exitcode, config, start_time)
   return run_result
 end
 
-def start_run(config, params)
+def start_run(config, action_input)
   exitcode = DEFAULT_EXITCODE
-  cmd = make_command_array(config, params)
-  env = make_environment_hash(params)
+  cmd = make_command_array(config, action_input)
+  env = make_environment_hash(action_input)
   start_time = Time.now
 
   run_result = Puppet::Util::Execution.execute(cmd, {:failonfail => false,
@@ -124,11 +128,12 @@ def start_run(config, params)
   end
 
   exitcode = run_result.exitstatus
+  puppet_output = run_result.to_s
 
-  if disabled?(run_result.to_s)
+  if disabled?(puppet_output)
     return make_error_result(exitcode, Errors::Disabled,
                              "Puppet agent is disabled")
-  elsif running?(run_result.to_s)
+  elsif running?(puppet_output)
     return make_error_result(exitcode, Errors::AlreadyRunning,
                              "Puppet agent is already performing a run")
   end
@@ -154,7 +159,7 @@ def metadata()
           },
           :required => [:env, :flags]
         },
-        :output => {
+        :results => {
           :type => "object",
           :properties => {
             :kind => {
@@ -201,15 +206,17 @@ def metadata()
   }
 end
 
-def run(params_and_config)
-  if !params_and_config
-    return make_error_result(DEFAULT_EXITCODE, Errors::InvalidJson,
-                             "Invalid json parsed on STDIN. Cannot start run action")
+def run(config, action_input)
+  if !File.exist?(config["puppet_bin"])
+    return make_error_result(DEFAULT_EXITCODE, Errors::NoPuppetBin,
+                             "Puppet executable '#{config["puppet_bin"]}' does not exist")
   end
 
-  params = params_and_config["params"]
-  config = params_and_config["config"]
-  params["flags"] = params["flags"] | ["--onetime", "--no-daemonize", "--verbose"]
+  return start_run(config, action_input)
+end
+
+def get_configuration(args)
+  config = args["configuration"] || {}
 
   if config["puppet_bin"].nil? || config["puppet_bin"].empty?
     if !is_win?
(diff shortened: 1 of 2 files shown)


modules/pxp-module-puppet
```diff
@@ -235,13 +284,14 @@ if __FILE__ == $0
   if action == 'metadata'
     puts metadata.to_json
   else
-    params_and_config = JSON.parse($stdin.read.chomp) rescue nil
+    args = JSON.parse($stdin.read.chomp) rescue nil
 
-    run_result = run(params_and_config)
-    puts run_result.to_json
+    # TODO(ale): retrieve action_output instead of action_results,
```
parisiale: This can be removed; it's fine as it is.

modules/pxp-module-puppet
```diff
@@ -124,11 +125,12 @@ def start_run(config, params)
   end
 
   exitcode = run_result.exitstatus
+  output = run_result.to_s
```
parisiale: Do you mind calling this one `puppet_output`?

parisiale:
:+1: 
Tested with https://github.com/puppetlabs/pxp-agent/pull/305; we're able to restart pxp-agent in the middle of a run without losing the results.

Perhaps you could squash the last three commits once were done with this.
Make sure there's an empty "new line" at the end of the files :-P

@richardc please have a look at the changes.

mruzicka:
I've force pushed this PR with minor changes discussed with @parisiale in HipChat.
AFAIKT all the files modified in the frame of this PR end in a new line character:
```
[mruzicka@michalruzickambp pxp-agent]$ od -ta modules/pxp-module-puppet | tail -2
0017440   sp   e   n   d  nl   e   n   d  nl                            
0017451
[mruzicka@michalruzickambp pxp-agent]$ od -ta modules/spec/unit/modules/puppet_spec.rb | tail -2
0037500   sp  sp   e   n   d  nl   e   n   d  nl                        
0037512
[mruzicka@michalruzickambp pxp-agent]$ git status
On branch PCP-251_simple_output_files_support
Your branch is up-to-date with 'origin/PCP-251_simple_output_files_support'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	input.json
	pxp-agent.conf

nothing added to commit but untracked files present (use "git add" to track)
```

I'm happy to squash the commits into one once this passes review.

parisiale:
I've closed the PR by mistake.

modules/pxp-module-puppet
```diff
@@ -221,12 +225,57 @@ def run(params_and_config)
...
+      else
+        1
+      end
+
+      begin
+        File.open(output_files["exitcode"], 'w') do |f|
+          f.print(status)
+        end
+      rescue
+        exit 5 # this exit code is reserved for problems with opening
```
parisiale: You should do as above (print make_error_result()).

mruzicka: At this point the result should have been written - mind you that this is an ```at_exit``` block. The problem is only with creating/writing the ```exitcode``` file.

parisiale: It doesn't matter.

If pxp-agent is running, it will consider this a failure because it will get `exitcode == 5`.
If pxp-agent is not running, once it restarts, it will consider this as a failure, since the "exitcode" file is not in the results directory.
This will change anyway when we deal with exception handling; the module has to report its own errors on `stderr` or on the "stderr" file, not on the results object.
Also, consider that this is an edge case; pxp-module-puppet would need to successfully write to "stdout" and "stderr" but fail to write on "exitcode".

modules/pxp-module-puppet
```diff
@@ -91,12 +95,9 @@ def get_result_from_report(exitcode, config, start_time)
...
   end
 
-  run_result = last_run_result(exitcode)
-
-  if exitcode != 0
-    run_result = make_error_result(exitcode, Errors::NonZeroExit,
-                                   "Puppet agent exited with a non 0 exitcode")
-  end
+  run_result = exitcode == 0 ?  last_run_result(exitcode)
+    : make_error_result(exitcode, Errors::NonZeroExit,
```
richardc: Use of the ternary here over multiple lines is really hard to read.  If we're going to touch this for clarity I'd suggest a more explicit structure.

```
if exitcode == 0
  run_result = last_run_result(exitcode)
else
  run_result = make_error_result(exitcode, Errors::NonZeroExit,  "Puppet agent exited with a non 0 exitcode")
end
```

modules/pxp-module-puppet
```diff
@@ -109,10 +110,10 @@ def get_result_from_report(exitcode, config, start_time)
   return run_result
 end
 
-def start_run(config, params)
+def start_run(config, action_input)
```
richardc: If it's appropriate to rename `params` -> `action_input` here, isn't it also appropriate to rename them at the point of use?  https://github.com/mruzicka/pxp-agent/blob/PCP-251_simple_output_files_support/modules/pxp-module-puppet#L64

I don't think the rename adds much, but if done it should be consistent.

modules/pxp-module-puppet
```diff
@@ -154,7 +156,7 @@ def metadata()
           },
           :required => [:env, :flags]
         },
-        :output => {
+        :results => {
```
richardc: This seems unrelated to the change at hand.

mruzicka: @parisiale Do you depend on this in your changes?

modules/pxp-module-puppet
```diff
@@ -235,13 +286,11 @@ if __FILE__ == $0
...
 
-    run_result = run(params_and_config)
-    puts run_result.to_json
+    action_results = action_run(args)
+    print action_results.to_json
 
-    if !run_result["error"].nil?
-      exit 1
-    end
+    exit 1 unless action_results["error"].nil?
```
richardc: Preferred the original expression of this (explicit if vs adverbial unless).  Adverbial unless shifts the eye to the action, but I think when following this code the test is more important to encounter first.

modules/pxp-module-puppet
parisiale: PCP-251 is about both that schema change and how to convey the output.

I think the point here is whether or not this should go in a separate commit or PR. I don't have any strong opinion on that.

modules/pxp-module-puppet
```diff
@@ -221,12 +225,59 @@ def run(params_and_config)
...
+end
+
+def action_run(args)
+  if !args
+    return make_error_result(DEFAULT_EXITCODE, Errors::InvalidJson,
+                             "Invalid json parsed on STDIN. Cannot start run action")
   end
 
-  return start_run(config, params)
+  output_files = args["output_files"]
```
richardc: Oh, this seems like a terrible place to pull this from.

richardc:
Hang on.  From the actions *input*?  As in the paths will be specifiable in the pxp command message?  That seems like an odd place to put the responsibility of picking a sane location for a spool directory.

Also if it is there, then shouldn't the action's input schema also change to reflect this?

parisiale:
I think you missed a lot of what was going on here.

For Ankeny, pxp-agent was passing PXP request params + config to the module. And the module was required to write its output on stdout.

Now that's formalized.
On top of that, pxp-agent may put the module in charge of writing its output to file. That is done via the optional "output_files" entry. In that case the module is required to write output on file.

modules/pxp-module-puppet
richardc: Well this is changing output (from the old way of running a module?) to results (from the new way I assume) so I'd expect this change to be part of the 'port to the new api' change, rather than the 'add this new behaviour' change.

richardc:
OK, feels weird but I see the code does that.

Generally seems it does what it should, but there are style nits I observed.

mruzicka:
I've added another commit addressing the style issues.
@richardc please ping me if you are happy with it.

modules/pxp-module-puppet
```diff
@@ -221,12 +228,59 @@ def run(params_and_config)
...
+  if output_files
+    begin
+      $stdout.reopen(File.open(output_files["stdout"], 'w'))
+      $stderr.reopen(File.open(output_files["stderr"], 'w'))
+    rescue => e
+      print make_error_result(DEFAULT_EXITCODE, Errors::InvalidJson,
+                              "Could not open output files: #{e.message}").to_json
```
richardc: This should probably use a new Error, rather than being 'invalid json'

modules/pxp-module-puppet
```diff
@@ -221,12 +228,59 @@ def run(params_and_config)
...
+      end
+
+      begin
+        File.open(output_files["exitcode"], 'w') do |f|
+          f.puts(status)
+        end
+      rescue => e
+        print make_error_result(DEFAULT_EXITCODE, Errors::InvalidJson,
+                                "Could not open exit code file: #{e.message}").to_json
+        exit 5 # this exit code is reserved for problems with opening
```
richardc: This is also not Json was invalid.  Some form of file handling error instead?

parisiale:
@richardc changing the error types would imply changing pxp-module-puppet's interface: https://github.com/puppetlabs/pxp-agent/blob/master/modules/pxp-module-puppet.md.
We don't want to do that here. There's PCP-219 for that. It's ok to use the errors as they are.

mruzicka:
I've squashed the commit into one.

PR #313, opened 2016-01-28
(PCP-207) Ext modules interface: rely only on files

parisiale:
We're simplifying the interface for external modules by requiring that
the output must be written only to file in case the `output_files`
entry is included in the action input.

This change implies that pxp-agent will need to read the outcom of the
action run from file.

On the other hand, modules will have a simpler output requirement, as
they won't need to guarantee the same output on files and stream.

[skip ci]

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -3,7 +3,7 @@
 ### Metadata
 
 The `metadata` of a module is a JSON object that contains a number of JSON
-schemas for specifying configuration options and actions. It contains:
+schemas for specifying its configuration options and actions. It contains:
 
  - **configuration**: (optional) schema that describes the module configuration format;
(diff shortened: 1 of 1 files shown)


modules/README.md
```diff
@@ -153,16 +153,40 @@ that the `configuration` schema is optional; in case it's not included in the
...
+In this case, pxp-agent guarantees that those files will be read only after the
 `exitcode` file is created; the existence of `exitcode` will be used to indicate
-the action completion.
+the processing completion.
 
-Once an action completes its execution, its results will be validated by
-pxp-agent by using the relative `results` schema contained in the `metadata`.
-The output on stderr will be treated as free format text.
+Also, in this case, pxp-agent will discard the output on stdout and stderr
+streams .
```
mruzicka: typo: space before period

mruzicka:
LGTM

PR #314, opened 2016-02-01
(maint) ExternalModule deals with empty module's output

parisiale:
As external modules can now return "5" in case they fail to write their
output, JsonContainer must be correctly instantiated in case there's no
output.

Also, external modules can return no output in case of error.

Ensuring JsonContainer is instantiated with "null" string in that case.
The validation is performed after that, by the parent Module class.


mruzicka:
LGTM

PR #315, opened 2016-02-02
(maint) Set cmake's -DTEST_VIRTUAL for Travis/Appveyor

parisiale:
Flagging the build option that enables virtual functions of PXPConnector
and related unit tests that rely on mocking the PCP Connector logic.


mruzicka:
LGTM

PR #316, opened 2016-02-02
(PCP-239) CMakeLists fixes

parisiale:


parisiale:
Closing as this will need a number of related fixes that will be part of PCP-239.

PR #319, opened 2016-02-03
(maint) Don't use MockConnector without TEST_VIRTUAL

parisiale:
Here we fix the request_processor_test.cc tests that were relying on
MockConnector without checking the TEST_VIRTUAL macro. They now safely
use the PXPConnector class as they don't rely on any WebSocket
connection.


mruzicka:
LGTM

PR #320, opened 2016-02-03
(PCP-255) Better logging

parisiale:


lib/src/module.cc
```diff
@@ -24,32 +24,31 @@ ActionOutcome Module::executeAction(const ActionRequest& request) {
...
-            if (!outcome.std_err.empty()) {
-                err_msg += " - stderr: " + outcome.std_err;
-            }
-            throw Module::ProcessingError { err_msg + outcome.std_err };
+            throw Module::ProcessingError {
+                (boost::format("The task executed for the %1% returned %2% "
+                               "results on stdout%3% - stderr:%4%")
+                    % request.prettyLabel()
+                    % (outcome.std_out.empty() ? "no" : "invalid")
+                    % (outcome.std_out.empty() ? "" : std::string(": ") + outcome.std_out)
```
mruzicka: is the ```std::string``` necessary here?

parisiale: No. As long as one argument of operator+ is a string.
Updated.

lib/src/external_module.cc
```diff
@@ -273,24 +273,21 @@ ActionOutcome ExternalModule::processRequestOutcome(const ActionRequest& request
...
 
     if (exit_code != EXIT_SUCCESS) {
-        if (!err_txt.empty()) {
-            LOG_ERROR("'%1% %2%' failure, returned %3%; error: %4%",
-                      module_name, action_name, exit_code, err_txt);
-        } else {
-            LOG_ERROR("'%1% %2%' failure, returned %3%",
-                      module_name, action_name, exit_code);
-        }
+        LOG_TRACE("Execution failure (exit code %1%) for the %2%%3%",
```
mruzicka: Should there be a separator (a colon?) between ```%2%``` and ```%3%```?

mruzicka: Oh, I see, its part of the parameter: ```"; stderr:\n" + err_txt```

lib/src/external_module.cc
```diff
@@ -299,14 +296,14 @@ ActionOutcome ExternalModule::processRequestOutcome(const ActionRequest& request
...
-        LOG_ERROR("'%1% %2%' output is not valid JSON: %3%",
-                  module_name, action_name, e.what());
-        std::string err_msg { "'" + module_name + " " + action_name + "' "
-                              "returned invalid JSON"  };
-        if (!err_txt.empty()) {
-            err_msg += " - stderr: " + err_txt;
-        }
-        throw Module::ProcessingError { err_msg };
+        LOG_DEBUG("Obtained invalid JSON on stdout for the %1%: %2%",
+                  request.prettyLabel(), e.what());
```
mruzicka: Should we log the malformed stdout too?
We log only the ```lth_jc::data_parse_error``` exception's message, and then we include the stderr in the ```Module::ProcessingError``` exception we throw. But we don't log the offending stdout anywhere.

lib/src/external_module.cc
```diff
@@ -362,12 +355,13 @@ ActionOutcome ExternalModule::callNonBlockingAction(const ActionRequest& request
...
-                  module_name, action_name, request.transactionId(),
-                  (exec.output.empty() ? "" : std::string("; output: ") + exec.output),
-                  (exec.output.empty() ? "" : std::string("; error: ") + exec.error));
-        throw Module::ProcessingError { "failed to write output to files" };
+        // This is unexpected. The outcome of the task will not be
+        // available for future transaction status requests
+        LOG_WARNING("The task process failed to write output on file for the %1%%2%%3%",
+                    request.prettyLabel(),
+                    (exec.output.empty() ? "" : "; stdout: " + exec.output),
+                    (exec.error.empty() ? "" : "; stderr: " + exec.error));
```
mruzicka: I'd prefer always including the ```stdout: ``` & ```stderr: ``` even if they are empty - in which case I'd just say ```(empty)``` similarly as above.

lib/src/external_module.cc
parisiale: :+1:

lib/src/external_module.cc
parisiale: :+1:

mruzicka:
LGTM :+1:

PR #322, opened 2016-02-04
(PCP-208) ThreadContainer associates names to threads

parisiale:


lib/inc/pxp-agent/thread_container.hpp
```diff
@@ -36,39 +40,58 @@ struct ManagedThread {
...
+
     uint32_t check_interval;  // [ms]
     uint32_t threads_threshold;  // number of stored thread objects
 
+    ThreadContainer() = delete;
     ThreadContainer(const std::string& name = "",
                     uint32_t _check_interval = THREADS_MONITORING_INTERVAL_MS,
                     uint32_t _threads_threshold = THREADS_THRESHOLD);
+    ThreadContainer(const ThreadContainer&) = delete;
+    ThreadContainer& operator=(const ThreadContainer&) = delete;
```
mruzicka: Why are you disabling the constructors (and the ```=``` operator)?

parisiale: For the rule of 3 plus threads are not copiable.

lib/src/thread_container.cc
```diff
@@ -64,8 +66,8 @@ ThreadContainer::~ThreadContainer() {
 
     // Detach the completed threads
     bool all_detached { true };
-    for (auto thread_ptr : threads_) {
-        if (!detachIfCompleted(thread_ptr)) {
+    for (auto itr = threads_.begin(); itr != threads_.end(); ++itr) {
```
mruzicka: why can't you use the foreach syntax on the map?

mruzicka:
LGTM

PR #323, opened 2016-02-05
(PCP-167) Change acceptance tests to not rely on logs to assert association

james-stocks:
Check pcp-broker's inventory to see if pxp-agent is included; instead of grepping pxp-agent.log, which is subject to change at any time
[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -52,9 +52,87 @@ def kill_pcp_broker(host)
   end
 end
 
+# Query pcp-broker's inventory of associated clients
+# Reference: https://github.com/puppetlabs/pcp-specifications/blob/master/pcp/inventory.md
+# @param broker hostname or beaker host object of the pcp-broker host
+# @param query pattern of client identities that you want to check e.g. pcp://*/agent for all agents, or pcp://*/* for everything
+# @return Ruby array of strings that are the client identities that match the query pattern
+# @raise Exception if pcp-client cannot connect to make the inventory request, does not receive a response, or the response is missing
+#        the required [:data]["uris"] property
+def pcp_broker_inventory(broker, query)
+  # Event machine is required by the ruby-pcp-client gem
+  # https://github.com/puppetlabs/ruby-pcp-client
+  em_thread = Thread.new { EventMachine.run }
+  Thread.pass until EventMachine.reactor_running?
+
+  mutex = Mutex.new
+  have_response = ConditionVariable.new
+  response = nil
+
+  client = connect_pcp_client(broker)
+
+  client.on_message = proc do |message|
+    mutex.synchronize do
+      resp = {
+        :envelope => message.envelope,
+        :data     => JSON.load(message.data)
+      }
+      response = resp
+      have_response.signal
+    end
+  end
+
+  message = PCP::Message.new({
+    :message_type => 'http://puppetlabs.com/inventory_request',
+    :targets => ["pcp:///server"]
+  })
+
+  message.data = {
+    :query => [query]
+  }.to_json
+
+  message_expiry = 10 # Seconds for the PCP message to be considered failed
+  inventory_expiry = 60 # Seconds for the inventory request to be considered failed
+  message.expires(message_expiry)
+
+  client.send(message)
+
+  begin
+    Timeout::timeout(inventory_expiry) do
+      loop do
+        mutex.synchronize do
+          have_response.wait(mutex)
+        end
+        break if response
+      end
+    end
+  rescue Timeout::Error
+    raise "Didn't receive a response for PCP inventory request"
+  end # wait for message
+
+  EventMachine.stop_event_loop
+  em_thread.join
+
+  if(!response.has_key?(:data))
+    raise 'Response to PCP inventory request is missing :data'
+  end
+  if(!response[:data].has_key?("uris"))
+    raise 'Response to PCP inventory request is missing an array of uri\'s'
+  end
+  response[:data]["uris"]
+end
+
+# Check if a PCP identity is in the pcp-broker's inventory
+# @param broker hostname or beaker host object of the machine running PCP broker
+# @param identity PCP identity of the client/agent to check e.g. "pcp://client01.example.com/agent"
+# @return true if PCP identity is included in the broker's inventory, false otherwise
+def is_associated?(broker, identity)
+  return pcp_broker_inventory(master, identity).include?(identity)
+end
+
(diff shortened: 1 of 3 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -132,8 +195,8 @@ def rpc_blocking_request(broker, targets,
     end
   rescue Timeout::Error
     mutex.synchronize do
-      if !have_all_rpc_responses?
-        raise "Didn't receive all PCP responses when requesting puppet run on #{targets}. Responses received were: #{responses.to_s}"
+      if !have_all_rpc_responses?(targets, responses)
+        raise "Didn't receive all PCP responses when requesting #{pxp_module} #{action} on #{targets}. Responses received were: #{responses.to_s}"
```
richardc: This changes the arguments passed to `have_all_rpc_responses?` but doesn't seem to change the definition of this function.  Was this just broken before?

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -52,9 +49,87 @@ def kill_pcp_broker(host)
...
+          have_response.wait(mutex)
+        end
+        break if response
+      end
+    end
+  rescue Timeout::Error
+    raise "Didn't receive a response for PCP inventory request"
+  end # wait for message
+
+  em.exit
```
richardc: This isn't particularly safe.  Best to call EM.stop and let the thread exit normally.

acceptance/lib/pxp-agent/test_helper.rb
james-stocks: Yes it was broken. Since it's in the rescue block, it usually never executes, so the mistake was overlooked.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -52,9 +49,87 @@ def kill_pcp_broker(host)
...
+# Reference: https://github.com/puppetlabs/pcp-specifications/blob/master/pcp/inventory.md
+# @param broker hostname or beaker host object of the pcp-broker host
+# @param query pattern of client identities that you want to check e.g. pcp://*/agent for all agents, or pcp://*/* for everything
+# @return Ruby array of strings that are the client identities that match the query pattern
+# @raise Exception if pcp-client cannot connect to make the inventory request, does not receive a response, or the response is missing
+#        the required [:data]["uris"] property
+def pcp_broker_inventory(broker, query)
+  # Event machine is required by the ruby-pcp-client gem
+  # https://github.com/puppetlabs/ruby-pcp-client
+  em = Thread.new { EM.run }
```
richardc: Needs to pause until EM is started to get the benefit of this:

https://github.com/faye/faye-websocket-ruby/blob/master/lib/faye/websocket.rb#L37

Will be updating the example usage in ruby-pcp-client as part of the rework I'm doing.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -52,9 +49,87 @@ def kill_pcp_broker(host)
...
+
+  message_expiry = 10 # Seconds for the PCP message to be considered failed
+  inventory_expiry = 60 # Seconds for the inventory request to be considered failed
+  message.expires(message_expiry)
+
+  client.send(message)
+
+  begin
+    Timeout::timeout(inventory_expiry) do
+      done = false
```
richardc: What is done being used for here?

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -52,9 +49,87 @@ def kill_pcp_broker(host)
...
+  response = nil
+
+  client = connect_pcp_client(broker)
+
+  client.on_message = proc do |message|
+    mutex.synchronize do
+      resp = {
+        :envelope => message.envelope,
+        :data     => JSON.load(message.data),
+        :debug    => JSON.load(message.debug)
```
parisiale: You don't need :debug

acceptance/tests/pxp-module-puppet/run_puppet.rb
```diff
@@ -8,7 +8,8 @@
       cert_dir = configure_std_certs_on_host(agent)
       create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_test_certs(master, agent, i + 1, cert_dir).to_s)
       on agent, puppet('resource service pxp-agent ensure=running')
-      expect_file_on_host_to_contain(agent, logfile(agent), PXP_AGENT_LOG_ENTRY_ASSOCIATION_SUCCESS, seconds_to_wait = 60)
+      assert(is_associated?(master, "pcp://client0#{i+1}.example.com/agent"),
+             "Agent #{agent} with PCP identity pcp://client0#{i+1}.example.com/agent is not associated with pcp-broker")
```
richardc: Shouldn't the message here describe not the failure, but what is being tested for?

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -132,16 +194,43 @@ def rpc_blocking_request(broker, targets,
...
-      if !have_all_rpc_responses?
-        raise "Didn't receive all PCP responses when requesting puppet run on #{targets}. Responses received were: #{responses.to_s}"
+      if !have_all_rpc_responses?(targets, responses)
+        raise "Didn't receive all PCP responses when requesting #{pxp_module} #{action} on #{targets}. Responses received were: #{responses.to_s}"
       end
     end
   end # wait for message
 
-  em.exit
+  em.stop unless em.stop?
```
richardc: You should be stopping eventmachine here, not the thread it's runloop is in.  `EM.stop`

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -69,23 +143,19 @@ def rpc_blocking_request(broker, targets,
   # Event machine is required by the ruby-pcp-client gem
   # https://github.com/puppetlabs/ruby-pcp-client
   em = Thread.new { EM.run }
```
richardc: Keeping a handle to this thread isn't really needed, unless you want to .join it after calling EM.stop.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -69,23 +143,19 @@ def rpc_blocking_request(broker, targets,
...
+  Thread.pass until EventMachine.reactor_running?
```
richardc: `EM` is an alias to `EventMachine` so this code works, but for consistency make this EM.reactor_running? or change the preceding line to be `EventMachine.run`

PR #324, opened 2016-02-05
(maint) Add dependency on librt

MikaelSmith:
pxp-agent uses Boost.Thread directly, which on some platforms has a
dependency on librt. Previously it picked up the dependency as a
side-effect of using Leatherman, which is fragile when using Leatherman
shared libraries and Boost static libraries.


MikaelSmith:
Required before we can use https://github.com/puppetlabs/leatherman/pull/131.

parisiale:
:+1:

PR #325, opened 2016-02-06
(PE-13918) Solaris pxp-agent stopped after 2015.2.3 to 2015.3.x upgrade

mwbutcher:
Prior to this change the pxp-agent service was configured with a dependancy
on pxp-agent.conf which doesn't exist imediately following an upgrade of PE.
This state results in the service not starting as expected until the first
puppet run following the upgrade.

This commit removes the dependancy on pxp-agent.conf which allows the service
to start.


parisiale:
:+1:

PR #326, opened 2016-02-08
(maint) Remove job_id entry from RPC Error schema

parisiale:
Removing that entry as is not in the specs anymore.

[skip ci]


mruzicka:
LGTM

PR #327, opened 2016-02-08
(PCP-202) Module's interface: success when results are valid

parisiale:
Making it clear in the modules interface specs that pxp-agent will not
mark an action as failed by considering the action's exit code.

Instead, an action will succeed whenever its results on stdout are
validated by using the "results" JSON schema provided by the action's
metadata.

[skip ci]

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -2,6 +2,9 @@
 
 ### Metadata
 
+When a module is invoked with no arguments, it must write its `metadata` to
+stdout.
+
 The `metadata` of a module is a JSON object that contains a number of JSON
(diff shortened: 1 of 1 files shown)


mruzicka:
All our existing modules (the only real one and the testing ones) support the ```metadata``` action - i.e. executing the module with an argument of ```metadata``` which has the same effect as executing the module without any argument.
Do we want to mandate this behavior or should we rather strive to reimplement the modules?

parisiale:
Yeah, we can do that.
I'll make a separate PR.

PR #328, opened 2016-02-08
(maint) Fix dependencies on AIX

MikaelSmith:
librt and Boost.Chrono are dependencies of pxp-agent, which need to be
included explicitly on AIX.


parisiale:
:+1:

PR #329, opened 2016-02-09
(PCP-206) clearer structure of pxp-module-puppet output

mruzicka:
Information obtained from a puppet run is now separated from messaging by the ```pxp-module-puppet``` in the module's output.


parisiale:
:+1:

PR #330, opened 2016-02-09
(PCP-227) Acceptance test of pxp-module-puppet run with 'changed' result

james-stocks:
Use pxp-module-puppet to run puppet with the expected result being 'changed'

[skip ci]

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
@@ -1,0 +1,59 @@
+require 'pxp-agent/test_helper.rb'
+
+ENVIRONMENT_NAME = 'changed_result'
+
+test_name 'C93062 - Run puppet and expect \'changed\' result' do
+
+  step 'On master, create a new environment that will result in Changed' do
+    environmentpath = master.puppet['environmentpath']
+    site_manifest = "#{environmentpath}/#{ENVIRONMENT_NAME}/manifests/site.pp"
+    on(master, "cp -r #{environmentpath}/production #{environmentpath}/#{ENVIRONMENT_NAME}")
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  notify {'Notify resources cause a Puppet run to have a \\'changed\\' outcome':}
+}
+SITEPP
+    on(master, "chmod 644 #{site_manifest}")
+  end
+
+  step 'Ensure each agent host has pxp-agent running and associated' do
+    agents.each_with_index do |agent, i|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      cert_dir = configure_std_certs_on_host(agent)
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_test_certs(master, agent, i + 1, cert_dir).to_s)
+      on agent, puppet('resource service pxp-agent ensure=running')
+      assert(is_associated?(master, "pcp://client0#{i+1}.example.com/agent"),
+             "Agent #{agent} with PCP identity pcp://client0#{i+1}.example.com/agent should be associated with pcp-broker")
+    end
+  end
+
+  step "Send an rpc_blocking_request to all agents" do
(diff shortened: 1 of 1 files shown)


acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
```diff
@@ -0,0 +1,65 @@
...
+      target_identities << "pcp://client0#{i+1}.example.com/agent"
+    end
+    responses = nil # Declare here so not local to begin/rescue below
+    begin
+      responses = rpc_blocking_request(master, target_identities,
+                                      'pxp-module-puppet', 'run',
+                                      {:env => [], :flags => ['--server', "'#{master}'",
+                                                              '--onetime',
+                                                              '--no-daemonize',
+                                                              "--environment", "'#{ENVIRONMENT_NAME}'"]
```
richardc: `"'#{ENVIRONMENT_NAME}'"` is a lot of overquoting.  Just ENVIRONMENT_NAME should work.

acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
```diff
@@ -0,0 +1,65 @@
...
+  step "Send an rpc_blocking_request to all agents" do
+    target_identities = []
+    agents.each_with_index do |agent, i|
+      target_identities << "pcp://client0#{i+1}.example.com/agent"
+    end
+    responses = nil # Declare here so not local to begin/rescue below
+    begin
+      responses = rpc_blocking_request(master, target_identities,
+                                      'pxp-module-puppet', 'run',
+                                      {:env => [], :flags => ['--server', "'#{master}'",
```
richardc: Same overquoting of master here.

acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
```diff
@@ -0,0 +1,65 @@
...
+                                                              '--onetime',
+                                                              '--no-daemonize',
```
richardc: I believe these flags are there by default.  https://github.com/puppetlabs/pxp-agent/blob/master/modules/pxp-module-puppet#L242

acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
```diff
@@ -0,0 +1,65 @@
...
+
+test_name 'C93062 - Run puppet and expect \'changed\' result' do
+
+  step 'On master, create a new environment that will result in Changed' do
+    environmentpath = master.puppet['environmentpath']
+    site_manifest = "#{environmentpath}/#{ENVIRONMENT_NAME}/manifests/site.pp"
+    on(master, "cp -r #{environmentpath}/production #{environmentpath}/#{ENVIRONMENT_NAME}")
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  include #{MODULE_NAME}
```
richardc: This use of a module seems overkill.  The entire site.pp could be just the notify resource.

Andrew-Hayes:
Looks good, however is a few i+1 around. perhaps a variable would be better?

agent_index = i+1

james-stocks:
@Andrew-Hayes : agreed - but that pattern is used throughout the repo; so should be improved everywhere and not just this test.
At the same time; making that improvement is perhaps low-value; as the acceptance suite needs to be updated to use Puppet certs (i.e. the agent certname) instead of the test certs that these i + 1's are referring to; so they'll be removed in the near future anyway.

acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
```diff
@@ -0,0 +1,60 @@
+require 'pxp-agent/test_helper.rb'
+
+ENVIRONMENT_NAME = 'changed_result'
+MODULE_NAME = 'changed'
```
richardc: I think MODULE_NAME is leftover from a previous swing?   Nothing is using it now.

PR #335, opened 2016-02-11
(PCP-297) On POSIX, create /var/run/puppetlabs if needed

parisiale:
Solaris cleans up /var/run at reboot. Daemonization relies on the
existence of the PID file directory. Here we create such directory when
running on Solaris, if necessary.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -683,6 +683,21 @@ void Configuration::validateAndNormalizeOtherSettings() {
         }
 
         auto pid_dir = fs::path(pid_file).parent_path().string();
+
+        // We know that, on Solaris, the /var/run when gets cleaned up
+        // at reboot; daemonization assumes that the PID file
+        // directory exists. So, if we're using the default PID
+        // directory, ensure it exists (PCP-297)
+        if (pid_dir == "/var/run/puppetlabs" && !fs::exists(pid_dir)) {
+            try {
+                LOG_DEBUG("Creating the PID file directory '%1%'", pid_dir);
+                fs::create_directories(pid_dir);
+            } catch (const std::exception& e) {
+                LOG_DEBUG("Failed to create '%1%' directory: %2%", pid_dir, e.what());
+                throw Configuration::Error { "failed to create the PID file directory" };
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


lib/src/configuration.cc
```diff
@@ -683,6 +683,22 @@ void Configuration::validateAndNormalizeOtherSettings() {
         }
 
         auto pid_dir = fs::path(pid_file).parent_path().string();
+
+#ifdef __sun
```
kylog: Should we enable this logic everywhere? E.g. someone might configure a linux distro like this as well.

lib/src/configuration.cc
```diff
@@ -683,6 +683,22 @@ void Configuration::validateAndNormalizeOtherSettings() {
...
+        // Solaris cleans up /var/run when rebooting; daemonization
+        // assumes that the PID file directory exists. If we're using
+        // the default directory, make sure it exists (PCP-297)
+        if (pid_dir == "/var/run/puppetlabs" && !fs::exists(pid_dir)) {
```
kylog: Should we create the directory if it doesn't exist even if it's not the default directory? The counter-argument to that idea would be that if someone has specified the pidfile manually but typo'd the path, they wouldn't want the directory created.

If we do restrict this logic to the default directory, can we DRY out occurences of `/var/run/puppetlabs` and use `DEFAULT_PID_FILE` above?

parisiale: @kylog yeah, I thought about that. But pxp-agent does not create any directories, apart from the spool ones; we followed such principle in the `Configuration` class. So, what I've tried to do here was to limit the case where we create the PID directory (Solaris platform AND default PID dir).

Also, @richardc was considering  a way to ensure the existence of the PID directory that would not involve pxp-agent directly.

kylog: Yeah, on sleeping on it, I agree with restricting this to just when `pid_dir == "/var/run/puppetlabs"`.

I still like the idea about having this fix in place for non-solaris os's or installs which mount /var/run on a temp filesystem (i.e. my comment above, not this comment :))

mruzicka:
LGTM

kylog:
:+1:

PR #336, opened 2016-02-12
(maint) Improve PCP broker and client usage in acceptance

james-stocks:
A couple of minor improvements:
 * pcp-client should be set to debug loglevel if beaker is using debug logging
 * The pre-suite code should not have its own code for starting pcp-broker; especially now that the helper lib has code to test pcp-broker on startup

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -276,7 +276,8 @@ def connect_pcp_client(broker)
   client = PCP::Client.new({
     :server => broker_ws_uri(broker),
     :ssl_cert => "../test-resources/ssl/certs/controller01.example.com.pem",
-    :ssl_key => "../test-resources/ssl/private_keys/controller01.example.com.pem"
+    :ssl_key => "../test-resources/ssl/private_keys/controller01.example.com.pem",
+    :loglevel => logger.is_debug? ? Logger::DEBUG : Logger::WARN
   })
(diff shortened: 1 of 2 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -6,6 +6,9 @@
 
 # This file contains general test helper methods for pxp-agent acceptance tests
 
+# The standard path for git checkouts is where pcp-broker will be
+GIT_CLONE_FOLDER = '/opt/puppet-git-repos'
```
james-stocks: I'm not happy with this being defined here. I've created PCP-298 to refactor this lib - possibly at this point pcp-broker deserves its own helper file

richardc: The comment says it's standard, so doesn't the place where it's standardised export a constant?

richardc: I was able to answer my own question.  This is the place where it's now being standardised, so it can be used in this helper and also in acceptance/setup/common/010_Setup_Broker.rb

richardc: On checking out the branch and poking around I now follow - this is the standard place, as it's now common to the call to `clone_git_repo_on` in acceptance/setup/common/010_Setup_Broker.rb and the `run_pcp_broker` helper here.  Will do for now.

PR #338, opened 2016-02-12
(maint) Acceptance - restructure retry logic for PCP client

james-stocks:
An intermittent failure in test_helper/connect_pcp_client in CI might be due to the retry logic re-using the same PCP::Client instance.
This commit changes the retry logic so that a new PCP::Client instance is used for each attempt.
[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -276,17 +276,17 @@ end
 # @return instance of pcp-client associated with the broker
 # @raise exception if could not connect or client is not associated with broker after connecting
 def connect_pcp_client(broker)
-  client = PCP::Client.new({
-    :server => broker_ws_uri(broker),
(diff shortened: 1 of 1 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -276,19 +276,18 @@ def rpc_blocking_request(broker, targets,
...
   connected = false
   until (connected || retries == max_retries) do
+    client = PCP::Client.new({
+      :server => broker_ws_uri(broker),
+      :ssl_cert => "../test-resources/ssl/certs/controller01.example.com.pem",
+      :ssl_key => "../test-resources/ssl/private_keys/controller01.example.com.pem",
+      :loglevel => logger.is_debug? ? Logger::DEBUG : Logger::WARN
+    })
     connected = client.connect(5)
-    retries += 1
```
richardc: I think this was important.

PR #341, opened 2016-02-19
(PCP-208) Improve action processing

parisiale:
With these commits we introduce new data structures to allow processing
the results of actions when they are retrieved from file, in the same
way as when they are obtained directly at the end of an action run. This
will enable pxp-agent to be fully resilient to restarts occurred during
an action run.

Also, the changes introduced here enable a better encapsulation of the
logic for formatting messages and provide new metadata information about
the processing of tasks. All that will ease the implementation of the
PXP changes suggested in PCP-202.

- removing the ActionOutput data structure; the results and outcome of
  an action run are now contained in the new ActionResponse that
  includes 1) ActionOutput (the action's output on stdout, stderr, and
  exit code) and 2) an action_metadata JSON object (which schema is
  enforced);
- refactoring the logic for the validation of the action output in order
  to make it accessible to the 'status query' implementation;
- creating a new ResultsStorage class that provides an interface to
  access and update the data stored in the spool directory;
- refactoring the module and request types.

We are removing the status module as the logic for implementing the 'status
query' action needs to have access to the stored modules to perform the
validation of actions' output, so it will be hosted by the
RequestProcessor.

Adding the logic to process a 'status query' request to the
RequestProcessor. The new logic provides new functionalities compared to
the old status module:

- validates the action's output obtained from the spool directory, in
  case it was not previously validated;
- updates the metadata in case it wrongly reports that the task is
  running;
- relies on the existence of the exitcode file to determine the task
  completion;
- supports new metadata entries and status values, as per PCP-202.


lib/inc/pxp-agent/action_request.hpp
```diff
@@ -11,13 +13,6 @@
 
 namespace PXPAgent {
 
-namespace lth_jc = leatherman::json_container;
```
mruzicka: Why removing this alias?

lib/inc/pxp-agent/action_response.hpp
```diff
@@ -0,0 +1,82 @@
...
+    void setStatus(ActionStatus status);
+
+    // Set results_are_valid to true in the action metadata and update
+    // the specified entries.
+    void setValidResultsAndEnd(leatherman::json_container::JsonContainer&& results,
+                               const std::string& execution_error = "");
+
+    // Set results_are_valid to false in the action metadata and
+    // update the execution_error entry.
+    void setBadResultsAndEnd(const std::string& execution_error);
```
mruzicka: On the first reading the ```End``` in the names of the two methods above felt like a verb - _to end_, for that reason I'd prefer the method names to be: ```setExecutionErrorAndEndTimestamp``` and ```setExecutionErrorAndEndTimestamp``` respectively.

lib/inc/pxp-agent/action_request.hpp
parisiale: because it's a header

lib/inc/pxp-agent/action_response.hpp
parisiale: no, too long; the entry is called "end"

lib/src/module.cc
```diff
@@ -1,55 +1,106 @@
...
+            response.action_metadata.get<std::string>("action"));
+        LOG_TRACE("Successfully validated the results for the %1%",
+                  response.prettyRequestLabel());
+    } catch (PCPClient::validation_error) {
+        // Modify the response metadata to indicate the failure
+        if (type() == ModuleType::Internal) {
+            // This is unexpected
+            err_msg += "The task executed for the ";
+            err_msg += response.prettyRequestLabel();
+            err_msg += " returned invalid results.";
```
mruzicka: Should we include the invalid results in the message?

parisiale: no; you already have the output available

lib/src/request_processor.cc
```diff
@@ -412,7 +391,312 @@ void RequestProcessor::processNonBlockingRequest(const ActionRequest& request) {
...
+                    ResultsMutex::LockGuard a_l { ResultsMutex::Instance().access_mtx };
+                    cached = ResultsMutex::Instance().exists(t_id);
+                }
+                if (cached)
+                    // The process doe not exist anymore, but its
+                    // mutex is still cached - wait a bit before
+                    // reading the metadata to allow the non-blocking
+                    // to lock its mutex and update the metadata
+                    pcp_util::this_thread::sleep_for(
+                        pcp_util::chrono::milliseconds(METADATA_RACE_MS));
```
mruzicka: It would be great to know which thread is handling the action and just wait for it to finish.

parisiale: Not waiting for it is basically the purpose of the 'status query' action.

mruzicka:
Phew ... lengthy but looking good to me.

lib/src/module.cc
mruzicka: Where, I don't see it being logged?

PR #343, opened 2016-02-22
(maint) Restore previous behaviour of 'status query'

parisiale:
With the changes introduced for PCP-208 we modified the behaviour of the
status query action, by setting the "status" of a completed action in
relation to the validation of the output on stdout.

Here we restore the old behaviour that consists of setting the "status"
in relation to the exitcode value.

The reason for restoring the old  behaviour is that we want to preserve
compatibility against old versions of clj-pxp-puppet.


mruzicka:
LGTM

PR #345, opened 2016-02-22
(maint) Fix the status query response

parisiale:


lib/src/action_response.cc
```diff
@@ -186,7 +186,13 @@ bool ActionResponse::valid(R_T response_type) const
             is_valid = action_metadata.includes(RESULTS);
             break;
         case (R_T::StatusOutput):
-            is_valid = !status_query_transaction.empty();
+            if (action_metadata.includes("status")) {
+                auto found = NAMES_OF_ACTION_STATUS.find(
```
mruzicka: Why don't you use ```std::map::at``` here?

parisiale: Because valid() should return a boolean, not throwing.

mruzicka:
LGTM

PR #346, opened 2016-02-22
(PCP-275) Purge spool directory

parisiale:


lib/src/request_processor.cc
```diff
@@ -185,6 +191,28 @@ RequestProcessor::RequestProcessor(std::shared_ptr<PXPConnector> connector_ptr,
     }
 
     logLoadedModules();
+
```
MikaelSmith: I'm not sure what this check is doing. Why minutes?

lib/src/request_processor.cc
```diff
@@ -819,4 +847,37 @@ void RequestProcessor::logLoadedModules() const
...
+// Spool directory purge task (private interface)
+//
+
+void RequestProcessor::spoolDirPurgeTask()
+{
+    auto num_minutes = Timestamp::getMinutes(spool_dir_purge_ttl_);
+    num_minutes *= 1.2;
+    LOG_DEBUG("Starting the task for purging the spool directory every %1% "
+              "minute%2%; thread id %3%",
+              num_minutes, lth_util::plural(num_minutes),
```
MikaelSmith: Uses of `plural` are something we'll want to look for when doing i18n work. We'll want to use Boost.Locale facilities to denote pluralization instead. Nothing to do yet though.

lib/src/request_processor.cc
```diff
@@ -819,4 +847,37 @@ void RequestProcessor::logLoadedModules() const
...
+            spool_dir_purge_cond_var_.wait_until(
+                the_lock,
+                now + pcp_util::chrono::minutes(num_minutes));
+
+        if (is_destructing_)
+            return;
+
+        auto num = storage_ptr_->purge(spool_dir_purge_ttl_,
+                                       thread_container_.getThreadNames());
+        LOG_INFO("Removed %1% directorie%2% from %3%",
```
MikaelSmith: Is `directorie` a singular somewhere? Seems like a typo.

lib/src/results_storage.cc
```diff
@@ -186,4 +191,55 @@ ActionOutput ResultsStorage::getOutput(const std::string& transaction_id,
...
+            if (!ongoing_transactions.empty()
+                    && std::find(ongoing_transactions.begin(),
+                                 ongoing_transactions.end(),
+                                 transaction_id) != ongoing_transactions.end())
+                return true;
+
+            try {
+                auto md = getActionMetadata(transaction_id);
+                if (md.get<std::string>("status") != "running"
+                        && !ts.isOlderThan(md.get<std::string>("start"))) {
```
MikaelSmith: My mind likes isNewerThan rather than !isOlderThan.

MikaelSmith:
Makes sense to me.

lib/src/request_processor.cc
parisiale: We provide a way to prevent purging that consists of setting the TTL value to 0, no matter what suffix is used in the TTL string (NB: TTL format = <value><suffix>).

So, here, the getMinutes() method allows to check if such 'value' was set to 0.
I think that having a getMinutes() method, that gives you complete information in any case,  could be more useful than having a value extractor, which would provide only a part of the TTL data.

lib/src/request_processor.cc
parisiale: True. I'll add a note to PCP-257.

lib/src/request_processor.cc
parisiale: Nope, just a mistake.

lib/src/results_storage.cc
parisiale: Mine too, but when I actually used Timestamp and got that feeling, my lazy side prevailed...

mruzicka:
LGTM.
The only general comment I have is that I'd like to see the time period suffix (```m```, ```h```, ```d```) in the ```spool-dir-purge-ttl``` option to be optional and default to one of the three (my preference would probably be the shortest: ```m```).

PR #347, opened 2016-02-23
Add helpers to wrap eventmachine use

richardc:
Here we extract methods to start and stop the eventmachine thread that ruby-pcp-client needs, then add some additional exception handling and state checks to investigate current issues with non-running reactors.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -84,7 +84,21 @@ end
 # Start an eventmachine reactor in a thread
 # @return Thread object where the EM reactor should be running
 def start_eventmachine_thread
-  em_thread = Thread.new { EventMachine.run }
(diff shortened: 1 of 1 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -81,6 +81,40 @@ def get_pcp_broker_status(host)
   end
 end
 
+# Start an eventmachine reactor in a thread
+# @return Thread object where the EM reactor should be running
+def start_eventmachine_thread
```
james-stocks: Another method name like get_eventmachine_thread would inform readers that this method returns a thread. Then again it wouldn't inform them that that starts a thread...

richardc: Yeah, I think 'start' or 'create' are the verbs here.

PR #349, opened 2016-02-24
(PCP-315) pxp-module-puppet input validation

mruzicka:
The input for the pxp-module-puppet's run action is not validated thoroughly enough. This commit improves it just a little bit.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -15,6 +15,15 @@ module Errors
   NonZeroExit = "agent_exit_non_zero"
 end
 
+class ProcessingError < StandardError
+  attr_reader :error_type
+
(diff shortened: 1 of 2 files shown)


parisiale:
:+1: 
This should cover what's reported in PCP-315.

modules/pxp-module-puppet
```diff
@@ -233,6 +242,9 @@ end
 
 def get_action_input(args)
   action_input = args["input"]
+  unless action_input.is_a?(Hash)
+    raise ProcessingError.new(Errors::InvalidJson, "The json received on STDIN did not contain a valid 'input' key.")
```
kylog: Maybe add `args.to_s` to the error message so we could get some hint about what is wrong with args?

PR #352, opened 2016-02-26
(PCP-317) flush pxp-module-puppet output explicitly

mruzicka:
There was a race condition between creating the exitcode file (which pxp-agent uses to determine if a pxp-module-puppet execution finished) and the actual exit of the pxp-module-puppet process (which implicitly flushes its output). This could lead (and led) to pxp-agent reading incomplete output of the pxp-module-puppet.
To prevent this we flush the output of pxp-module-puppet explicitly before writing the exitcode file.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -282,6 +282,10 @@ def action_run(input)
         1
       end
 
+      # flush the stdout/stderr before writing the exitcode
+      # file to avoid pxp-agent reading incomplete output
+      $stdout.fsync
+      $stderr.fsync
       begin
         File.open(output_files["exitcode"], 'w') do |f|
           f.puts(status)


modules/pxp-module-puppet
```diff
@@ -282,6 +282,10 @@ def action_run(input)
         1
       end
 
+      # flush the stdout/stderr before writing the exitcode
+      # file to avoid pxp-agent reading incomplete output
+      $stdout.fsync
```
parisiale: Perhaps, we should set `$stdout.sync = true` as you initially suggested, to include the print call of the rescue block.

mruzicka: I agree

mruzicka: After a second thought, I actually don't agree, since:

1. we can't guarantee valid JSON on stdout in the rescue block anyway - as by the time we print the response, there might have been something already written to the stdout
2. the print in the rescue block is followed by exit which flushes the buffers implicitly

parisiale: You're passing 5, so pxp-agent knows what to do and will use what's on stdout.
I think you should configure the un-buffered behaviour anyway for consistency.

mruzicka: I don't think it makes much sense, as it doesn't provide any string guarantees, it only guarantees ruby internal buffers are flushed on each write, but it doesn't guarantee the same for OS level buffers - so we'd still need the ```fsync``` call (which does call the system fsync, thus ensuring any buffered data is written to disk).

parisiale: No, and there's leatherman in between as well; so, as said, it's for consistency.

mruzicka: I think it's just clutter then, adding the ```sync = true``` doesn't really help anything - worse it may trick someone to believe that there is something like a working autoflush in ruby.

parisiale: I really don't see that as issue.

PR #356, opened 2016-02-26
(PCP-317) Pause before retrieving output; better logs

parisiale:


lib/src/request_processor.cc
```diff
@@ -496,7 +498,7 @@ void RequestProcessor::processStatusRequest(const ActionRequest& request)
                     // The process doe not exist anymore, but its
```
mruzicka: Typo: doe -> does

mruzicka:
LGTM

PR #357, opened 2016-02-29
(PCP-232) Acceptance - ensure pxp-agent is usable after restarting host

james-stocks:
[skip ci]


james-stocks:
Closing PR as the test will currently fail due to not using puppet ssl certs

PR #358, opened 2016-03-01
(PCP-296) Acceptance tests should use Puppet generated SSL certs

james-stocks:
Change the acceptance tests to use Puppet generated SSL certs instead of our test fixture certs.
For pcp-broker; the puppet certs are copied onto the standard test cert paths so that no change to pcp-broker code is needed.
Adds a pre-suite class to generate certs for the test executor so that it can certify itself when connecting pcp-client.

[skip ci]

### DIFF
--- a/acceptance/setup/common/060_Setup_PCP_Client.rb
+++ b/acceptance/setup/common/060_Setup_PCP_Client.rb
@@ -1,0 +1,19 @@
+require 'pxp-agent/test_helper'
+require 'socket'
+test_name 'Set up SSL certs for pcp-client to use'
+
+step 'On master create certs for the host running pcp-client' do
+  hostname = Socket.gethostname
+  on master, puppet("cert generate #{hostname}")
+  on(master, puppet('config print ssldir')) do |result|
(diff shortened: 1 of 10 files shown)


richardc:
:+1:

acceptance/setup/common/060_Setup_PCP_Client.rb
```diff
@@ -0,0 +1,19 @@
+require 'pxp-agent/test_helper'
+require 'socket'
+test_name 'Set up SSL certs for pcp-client to use'
+
+step 'On master create certs for the host running pcp-client' do
+  hostname = Socket.gethostname
+  on master, puppet("cert generate #{hostname}")
```
richardc: As this is just a role identity, this could be something like `beaker-controller` and would remove the need to determine the hostname of the beaker host.

james-stocks:
Tested on staging Jenkins - URL for the results posted in the ticket

james-stocks:
Closing as this should be against stable, not master

PR #359, opened 2016-03-01
(PCP-296) Acceptance tests should use Puppet generated SSL certs

james-stocks:
Change the acceptance tests to use Puppet generated SSL certs instead of our test fixture certs.
For pcp-broker; the puppet certs are copied onto the standard test cert paths so that no change to pcp-broker code is needed.
Adds a pre-suite class to generate certs for the test executor so that it can certify itself when connecting pcp-client.

[skip ci]


james-stocks:
This was reviewed previously as https://github.com/puppetlabs/pxp-agent/pull/358 but that PR was mistakenly pointing at master

PR #363, opened 2016-03-07
(PCP-232) Acceptance - ensure pxp-agent is usable after restarting host

james-stocks:
[skip ci]

### DIFF
--- a/acceptance/tests/restart_host_run_puppet.rb
+++ b/acceptance/tests/restart_host_run_puppet.rb
@@ -1,0 +1,59 @@
+require 'pxp-agent/test_helper.rb'
+
+test_name 'C94777 - Ensure pxp-agent functions after agent host restart' do
+
+  step 'Ensure each agent host has pxp-agent service running and enabled' do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      cert_dir = configure_std_certs_on_host(agent)
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+      on agent, puppet('resource service pxp-agent ensure=running enable=true')
+      assert(is_associated?(master, "pcp://#{agent}/agent"),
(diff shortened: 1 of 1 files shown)


parisiale:
This looks good to me.

Since you have specific tests for pxp-module-puppet, would it make sense to use the "echo echo" action instead of "pxp-module-puppet run" to decouple testing pxp-agent from testing pxp-module-puppet?
Of course it depends on what you mean by usable.

In case you decide to use the echo module, the "params" entry should be in the format: `"params":{"argument":"maradona"}`.

james-stocks:
DO NOT MERGE at the moment - see PCP-232

james-stocks:
@parisiale I think it is valuable to ensure that a pxp-module-puppet run works following a host reboot. There are perhaps two distinct tests here: 1. pxp-agent service works following a restart; 2. pxp-module-puppet run works following a restart. In the trade-off between efficient use of CI resource (single test case) and distinct isolated test results (2 separate test cases); I think being efficient with CI resource is more valuable in this case

parisiale:
:+1:

johnduarte:
@james-stocks the test looks fine to me.

When merged on top of SHA 13846131ce0797e6cea98cb8e78e9d023f364442, the passes on Ubuntu 14.04, but fails on Windows2012r2 with the following.

```
 Begin tests/restart_host_run_puppet.rb
        tests/restart_host_run_puppet.rb failed in 113.45 seconds
                  Test Case tests/restart_host_run_puppet.rb reported: #<Minitest::Assertion: Agent ll4bvsyaqmau3x1.delivery.puppetlabs.net with PCP identity pcp://ll4bvsyaqmau3x1.delivery.puppetlabs.net/agent should be associated with pcp-broker>
```

Other tests in this suite fail in a similar manner so it is not specifically related to this test.

james-stocks:
@johnduarte  I believe the errors you saw when trialling these tests was PCP-355 and not related to the code in this PR

johnduarte:
@james-stocks you are correct, the previous failure was an intermittent unrelated failure. I have executed a successful run on Windows2012r2 with this test.

PR #364, opened 2016-03-08
(PCP-308) run modules in own contracts on Solaris

mruzicka:
Solaris keeps track of the state of the system services by employing so called contracts. A contract is effectively a group of processes. Each service has its own contract. When a service starts a new process it is by default a member of the service's contract. When a service is restarted all processes in its contract are killed (or at least the system waits until they exit).
This behavior isn't the desired one for ```pxp-agent``` as we want the modules it spawns to outlive it in case it is restarted.
The solution is to make sure the modules are executed in their own contracts so that they are not killed when the ```pxp-agent``` service is restarted. In this commit we achieve this be executing the modules via the ```ctrun``` command.

### DIFF
--- a/lib/src/external_module.cc
+++ b/lib/src/external_module.cc
@@ -342,8 +342,10 @@ ActionResponse ExternalModule::callNonBlockingAction(const ActionRequest& reques
     LOG_TRACE("Input for the %1%: %2%", request.prettyLabel(), input_txt);
 
     auto exec = lth_exec::execute(
-#ifdef _WIN32
+#if defined(_WIN32)
         "cmd.exe", { "/c", path_, action_name },
+#elif defined(__sun)
+        "/usr/bin/ctrun", { "-l", "child", path_, action_name },
 #else
         path_, { action_name },
 #endif


lib/src/external_module.cc
```diff
@@ -191,13 +191,15 @@ const PCPClient::Validator ExternalModule::metadata_validator_ {
...
 const lth_jc::JsonContainer ExternalModule::getModuleMetadata()
 {
-    auto exec =
-#ifdef _WIN32
-        lth_exec::execute("cmd.exe", { "/c", path_, "metadata" },
+    auto exec = lth_exec::execute(
+#if defined(_WIN32)
+        "cmd.exe", { "/c", path_, "metadata" },
+#elif defined(__sun)
+        "/usr/bin/ctrun", { "-l", "child", path_, "metadata" },
```
parisiale: You don't need ctrun for the "metadata" call.

lib/src/external_module.cc
```diff
@@ -313,8 +315,10 @@ ActionResponse ExternalModule::callBlockingAction(const ActionRequest& request)
     LOG_TRACE("Input for the %1%: %2%", request.prettyLabel(), action_args);
 
     auto exec = lth_exec::execute(
-#ifdef _WIN32
+#if defined(_WIN32)
         "cmd.exe", { "/c", path_, action_name },
+#elif defined(__sun)
+        "/usr/bin/ctrun", { "-l", "child", path_, action_name },
```
parisiale: I don't think we need ctrun for this blocking procedure.

lib/src/external_module.cc
```diff
@@ -342,8 +346,10 @@ ActionResponse ExternalModule::callNonBlockingAction(const ActionRequest& reques
     LOG_TRACE("Input for the %1%: %2%", request.prettyLabel(), input_txt);
 
     auto exec = lth_exec::execute(
-#ifdef _WIN32
+#if defined(_WIN32)
         "cmd.exe", { "/c", path_, action_name },
+#elif defined(__sun)
+        "/usr/bin/ctrun", { "-l", "child", path_, action_name },
```
parisiale: I'm not sure whether passing the absolute path of ctrun is the right choice.
facter does that for Solaris' zonename: https://github.com/puppetlabs/facter/blob/master/lib/src/facts/solaris/zone_resolver.cc#L17.
Should we rely on leatherman's which instead and only pass the executable name?
@MikaelSmith , @branan  thoughts?

lib/src/external_module.cc
mruzicka: I was considering not using ```ctrun``` here (and above), but in the end I decided to use it here to keep the behavior consistent with other OSes.

MikaelSmith: I'm not sure we've been consistent about that. Also pinging @peterhuene.

I think we've tended to use absolute paths for things that are intrinsic to the OS, and relative paths for packages that might be installed as dependencies (and we've been moving to allowing an absolute path to be specified at compile-time for those as we move them into the puppet-agent package).

branan: My opinion is that for stuff where we know it'll always be in the same place we should just use absolute paths. I use `which` when either the command might not exist, or if it might be in different places depending on the particular OS flavor.

For ctrun in solaris, I think we can just use an absolute path

parisiale:
Tracking the Travis failure in PCP-340, which is unrelated to this change.

PR #365, opened 2016-03-11
(maint) Consider output processing delay in component test

parisiale:
ExternalModule waits 100 ms before processing the output of a
non-blocking test; here, we're adding an extra pause to the component
tests in order to avoid transient failures when testing the
non-blocking response.

Note that the component tests are asynchronous; they are executed only
when the TEST_VIRTUAL flag is enabled.


mruzicka:
LGTM

PR #366, opened 2016-03-11
DONT MERGE (maint) Wait for the module completion in component tests

parisiale:
With this commit we add an extra pause to the component test that
verifies that the provisional response is sent. Also ensuring that the
final non-blocking response is not sent in case notify_outcome is
false.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 /build/
 /dev-resources/
 /lib/tests/resources/test_spool/
+/lib/tests/resources/tmp/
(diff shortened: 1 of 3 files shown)


mruzicka:
The changes look good to me but the travis check failed :-(

parisiale:
I'll deal with this within PCP-340.
I don't know what's going on with Travis, as I can't reproduce any of the failures on OS X.

mruzicka:
OK, merging then.

PR #367, opened 2016-03-14
(PCP-340) Wait for the module completion in component tests

parisiale:
With this commit we add a function to check whether the external module
run finished by looking at the exit code file.

Removing the unnecessary pause from the bloking tests.


mruzicka:
LGTM

PR #368, opened 2016-03-14
(PCP-306) run report in tempfile

mruzicka:
Before this commit we let puppet runs executed from px-module-puppet write their run reports to the default location. Since the file is overwritten on every puppet run, we could never be sure when reading the file whether it was produced by the puppet run triggered by pxp-module-puppet or by a concurrent puppet run.
To avoid this, we update the pxp-module-puppet in this commit such that it ensures (by passing the ```--lastrunreport``` option) that the puppet runs it triggers write their run reports to uniquely named temporary files rather than to the default location.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -37,7 +37,7 @@ def is_win?
   return !!File::ALT_SEPARATOR
 end
 
-def last_run_result(exitcode)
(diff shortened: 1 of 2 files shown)


mruzicka:
closing this, as we decided we don't want to break the last run report file contract

PR #371, opened 2016-03-21
(PCP-338) Handle Associate Session errors

parisiale:
Retrying to connect in case of errors during Session Association; that
is done after a random pause in the  [5 s, 10 s] interval.

Handling Session Association failures.

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -66,9 +66,12 @@ int startAgent(std::vector<std::string> arguments) {
     try {
         Agent agent { Configuration::Instance().getAgentConfiguration() };
         agent.start();
+    } catch (const Agent::PCPConfigurationError& e) {
+        exit_code = PXP_AGENT_CONFIGURATION_FAILURE;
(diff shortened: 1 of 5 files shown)


parisiale:
Don't merge before https://github.com/puppetlabs/cpp-pcp-client/pull/144/files.

exe/main.cc
```diff
@@ -66,9 +66,12 @@ int startAgent(std::vector<std::string> arguments) {
     try {
         Agent agent { Configuration::Instance().getAgentConfiguration() };
         agent.start();
+    } catch (const Agent::PCPConfigurationError& e) {
```
james-stocks: This new Error struct is used in the event that association fails - should it therefore be named PCPAssociationError or something similar? To a user or a support engineer; "PCP configuration" might suggest something in a config file. I think in this context "configuration" instead means the combination of PCP broker/client components?

parisiale: No.

mruzicka:
LGTM, although I agree with @james-stocks, the exception naming could use some improvement.

PR #373, opened 2016-03-29
(maint) Remove move() when returning unique_ptr

parisiale:
Removing the std::move() call when returning the PIDFile unique_ptr from
the daemonize() function, to avoid clang error during compilation
(clang-703.0.29 seems to be more restrictive). Note that unique_ptr only
supports move assignments, so the std::move() call was redundant.


mruzicka:
LGTM

PR #374, opened 2016-03-29
(maint) Add name prefix for ThreadContainer unit tests

parisiale:
Adding a prefix for the names of threads created for ThreadContainer, to
avoid errors due to duplicate names.


mruzicka:
makes sense

PR #375, opened 2016-03-30
(PCP-284) In modules interface, specify "metadata" arg

parisiale:
For external modules, require that the metadata action is executed when
the module is invoked with "metadata" as its only argument.


mruzicka:
LGTM

PR #379, opened 2016-04-03
PCP-362 use PCP::Client#close rather than EM weirdness

richardc:
ruby-pcp-client 0.3.0 adds a long-missing PCP::Client#close method
to shut down a client after we're done with it.

Here we switch to using this close method with a long-running
eventmachine reactor thread, avoiding the painful races caused by
trying to manage client lifecycle by restarting the supervising
eventmachine.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -84,10 +84,8 @@ end
 
 # Start an eventmachine reactor in a thread
 # @return Thread object where the EM reactor should be running
-def start_eventmachine_thread
+def assert_eventmachine_thread_running
   if EventMachine.reactor_running?
-    puts "EventMachine already running!"
-    puts caller
     return EventMachine.reactor_thread
   end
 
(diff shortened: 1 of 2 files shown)


richardc:
Can be reviewed now, but will need 0.3.0 of ruby-pcp-client shipped with at least this PR merged:  https://github.com/puppetlabs/ruby-pcp-client/pull/11

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -84,10 +84,8 @@ def get_pcp_broker_status(host)
 
 # Start an eventmachine reactor in a thread
```
james-stocks: This comment might need updated along with the method name being updated

richardc: How so?  It's still starting an eventmachine reactor in a thread.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -84,10 +84,8 @@ def get_pcp_broker_status(host)
...
 # @return Thread object where the EM reactor should be running
-def start_eventmachine_thread
+def assert_eventmachine_thread_running
```
james-stocks: The method name may be confusing. But so could the old name.

PCP-298 is a ticket to refactor these helper libs - I'll add a comment specifying that this EventMachine stuff should be contained within a pcp-client helper lib

acceptance/lib/pxp-agent/test_helper.rb
richardc: Could add ", if needed" to be really clear.

acceptance/lib/pxp-agent/test_helper.rb
richardc: OK, how about `ensure_eventmachine_running`?

james-stocks:
Code changes LGTM but I could also try running the tests off your branch before we merge

PR #386, opened 2016-04-08
PCP-355 show logs on failures

richardc:
Here we add a helper to show logs on assertion failures, and wrap the `is_not_associated?` `is_associated?` `assert` calls in it, to add context to the periodic failures we are tracking.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -43,6 +43,29 @@ def expect_file_on_host_to_contain(host, file, expected, seconds=30)
   end
 end
 
+# Show the logs of the broker and agents - useful to see why a test failed
+def show_pcp_logs
+  puts "---- Broker log -----"
+  on(master, "cat /var/log/pcp-broker.log") do |result|
+    puts result.stdout
+  end
+
+  agents.each do |agent|
+    puts "----- agent #{agent} log -----"
+    on(agent, "cat #{logfile(agent)}") do |result|
+      puts result.stdout
+    end
+  end
+end
+
+# Evaluate the block, show logs if test assertions were triggered
+def show_pcp_logs_on_failure(&block)
+  yield
+rescue MiniTest::Assertion => exception
+  show_pcp_logs
+  raise exception
+end
+
 # Some helpers for working with a pcp-broker 'lein tk' instance
 def run_pcp_broker(host)
   on(host, "cd #{GIT_CLONE_FOLDER}/pcp-broker; export LEIN_ROOT=ok; lein tk </dev/null >/var/log/pcp-broker.log 2>&1 &")
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -62,6 +62,8 @@ end
 def show_pcp_logs_on_failure(&block)
   yield
 rescue MiniTest::Assertion => exception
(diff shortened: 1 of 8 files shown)


acceptance/tests/invalid_ssl_config.rb
```diff
@@ -113,9 +115,11 @@
 
     step 'Start pxp-agent and assert that it does not connect to broker'
     on agent, puppet('resource service pxp-agent ensure=running')
-    assert(is_not_associated?(master, "pcp://#{agent}/agent"),
-           "Agent identity pcp://#{agent}/agent for agent host #{agent} should not appear in pcp-broker's inventory " \
-           "when pxp-agent attempts to connect by broker IP instead of broker certified hostname")
+    show_pcp_logs_on_failure do
```
james-stocks: The new logging helper is being added in multiple places in the tests. Short term this seems correct; to show the state of the SUT at each failure. Longer term once we fix the issue; I think the helper should either only be used once (in the association test case) or if we want general logging in future, it should be inside the is_associated?/is_not_associated? helpers.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -43,6 +43,29 @@ def expect_file_on_host_to_contain(host, file, expected, seconds=30)
   end
 end
 
+# Show the logs of the broker and agents - useful to see why a test failed
+def show_pcp_logs
+  puts "---- Broker log -----"
```
james-stocks: Ideally, the beaker logger instance should always be used instead of puts. I think in practice, in our current CI, it makes no difference. PCP-365 mentions doing a sweep through the acceptance to remove all puts and print statements (i.e. I'm not sure there's value in changing this immediately; we can do the more general tidy-up later)

james-stocks:
Added two comments; aside from that looks good, but I'd like to test out the code myself first (including testing deliberate fails to ensure we get the PCP logs AND the expected beaker test case result)

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -43,6 +43,29 @@ def expect_file_on_host_to_contain(host, file, expected, seconds=30)
...
+      puts result.stdout
+    end
+  end
+end
+
+# Evaluate the block, show logs if test assertions were triggered
+def show_pcp_logs_on_failure(&block)
+  yield
+rescue MiniTest::Assertion => exception
+  show_pcp_logs
```
james-stocks: I tested out a deliberate failure and got output like this:

```
C93807 - Associate pxp-agent with a PCP broker
localhost $ scp /tmp/beaker20160413-25826-emv3w8 l7vlvi75pwdnrz2.delivery.puppetlabs.net:/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/etc/pxp-agent.conf {:ignore => }

* Stop pxp-agent if it is currently running

* Assert that the agent is not listed in pcp-broker inventory

* Start pxp-agent service

* Assert that agent is listed in pcp-broker inventory
---- Broker log -----
2016-04-13 04:20:59,371 INFO  [async-dispatch-2] [p.t.s.w.jetty9-service] Initializing web server(s).
2016-04-13 04:20:59,409 INFO  [async-dispatch-2] [p.p.b.service] Initializing broker service
```

If anyone outside the current core team was to encounter this in Jenkins output; it would be helpful if it was prefaced with a log/puts message like "Assertion exception was caught - now outputting pcp-broker and pxp-agent logs" so that the appearance of a pcp-broker log in the output is explained

PR #387, opened 2016-04-11
(PCP-345) Create spool dir if needed

parisiale:
Creating the spool directory at start if it doesn't exist.

This is done in order to facilitate running pxp-agent without super user
privileges.

Also, new unit tests are added.


mruzicka:
LGTM

PR #388, opened 2016-04-11
(maint) Refactor Configuration

parisiale:
- README change: required leatherman version
 - Style changes
 - Remove HW alias from Configuration header
 - Use boost::format for error messages, to make things clearer
 - Remove duplicated code from Configuration functions

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -34,6 +34,8 @@
 
 namespace PXPAgent {
 
+namespace HW = HorseWhisperer;
+
 namespace fs = boost::filesystem;
 namespace lth_file = leatherman::file_util;
 namespace lth_jc = leatherman::json_container;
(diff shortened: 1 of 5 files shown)


lib/src/configuration.cc
```diff
@@ -499,135 +512,107 @@ void Configuration::setDefaultValues() {
...
-            std::string err { "field '" + key + "' is not a valid "
-                              "configuration variable" };
-            throw Configuration::Error { err };
+        if (opt_idx->ptr->configured)
+            continue;
+
+        switch (opt_idx->ptr->type) {
+            case Integer:
+                check_key_type(key, "Integer", lth_jc::DataType::Int);
+                HW::SetFlag<int>(key, config_json.get<int>(key));
```
mruzicka: should this use ```HorseWhisperer``` rather than ```HW```

parisiale: no; the "HW" alias is defined above

mruzicka:
LGTM

PR #390, opened 2016-04-15
(PCP-375) Remove usage of test fixture SSL files from acceptance tests

james-stocks:
PCP-370 will remove the test fixture SSL certs in the test-resources folder. This PR removes all usage of those certs from the pxp-agent acceptance tests


parisiale:
Awesome :+1:

PR #391, opened 2016-04-18
WIP (maint) Update ruby-pcp-client to 0.4.0

richardc:
0.4.0 adds ssl verification, so we need to pass the ssl_ca_cert when we
construct the PCP::Client


richardc:
Wrong base.

PR #393, opened 2016-04-18
(PCP-376) Acceptance - do not attempt timesync on OSX

james-stocks:
QENG-3786 is a bug with beaker's timesync method - it will fail on OSX due to ntpd service already running.
Until QENG-3786 is fixed, this commit skips attempting timesync on OSX hosts.
[skip ci]

### DIFF
--- a/acceptance/setup/common/005_SyncTime.rb
+++ b/acceptance/setup/common/005_SyncTime.rb
@@ -4,4 +4,6 @@ extend Beaker::HostPrebuiltSteps
 # Note: QENG-3641 would allow beaker to take care of this for us
 test_name 'Ensure hosts have synchronized clocks'
 
-timesync(hosts, {:logger => logger})
+# QENG-3786 - OSX is already running ntpd and manually running ntpdate will return 1
+applicable_hosts = hosts.select{|host| host['platform'] !~ /osx/}
+timesync(applicable_hosts, {:logger => logger})


acceptance/setup/common/005_SyncTime.rb
```diff
@@ -4,4 +4,8 @@
 # Note: QENG-3641 would allow beaker to take care of this for us
 test_name 'Ensure hosts have synchronized clocks'
 
-timesync(hosts, {:logger => logger})
+# QENG-3786 - OSX is already running ntpd and manually running ntpdate will return 1
+applicable_hosts = hosts.select{|host| host['platform'] !~ /osx/}
+if applicable_hosts then
```
richardc: What's this if statement trying to guard against?  Array#select will aways return an Array, and even the empty array is true.

james-stocks: The intention is to avoid passing beaker an empty array; as it generally doesn't validate its arguments before attempting to use them.
You are correct that the if statement is useless as it always resolves as true - I'll update to have it test on Array.empty? instead

richardc: In this case, it'll just log that the array was empty:  https://github.com/puppetlabs/beaker/blob/master/lib/beaker/shared/host_manager.rb#L107-L113

So yes, if you want to guard against that, check for !Array#empty? but it doesn't seem that serious to just have the log message occur.

james-stocks:
@richardc Updated to simply allow an empty array to be passed to beaker

PR #394, opened 2016-04-18
(PCP-370) Remove sets of SSL certificates

parisiale:
Removing two sets of SSL certificates from the test-resources directory,
as they were not used for testing; users should rely on their own
certificates.


mruzicka:
LGTM

PR #396, opened 2016-04-19
(maint) Make tests compatible with leatherman 0.4 and 0.5

parisiale:
Removing the Catch's applyCommandLine() and passing CLI arguments
directly to the Session::run() function; note that setting ConfigData
showDurations to "Always" works in the same way as it's not changed by
our CI pipeline via CLI.

### DIFF
--- a/lib/tests/main.cc
+++ b/lib/tests/main.cc
@@ -16,7 +16,7 @@
 #include <leatherman/logging/logging.hpp>
 #endif
 
-int main(int argc, const char** argv) {
+int main(int argc, char** argv) {
 #ifdef ENABLE_LOGGING
     leatherman::logging::setup_logging(boost::nowide::cout);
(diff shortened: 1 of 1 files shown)


lib/tests/main.cc
```diff
@@ -16,28 +16,28 @@
 #include <leatherman/logging/logging.hpp>
 #endif
 
-int main(int argc, const char** argv) {
+int main(int argc, char** argv) {
```
mruzicka: Why this?

mruzicka:
LGTM

lib/tests/main.cc
parisiale: Different signature in Catch 1.4.0.

mruzicka: ok

PR #399, opened 2016-04-21
(PCP-230) Acceptance - restart pxp-agent during Puppet run

james-stocks:
This adds a new test case where we kick off a non-blocking Puppet agent run; and then terminate pxp-agent in the middle of the agent run; then assert that when the run completes and pxp-agent is once again running; we can report on the status of the run.

This required adding a new test helper to perform a non-blocking PXP request

[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -246,6 +246,30 @@ end
 def rpc_blocking_request(broker, targets,
                          pxp_module = 'pxp-module-puppet', action = 'run',
                          params)
+  rpc_request(broker, targets, pxp_module, action, params, true)
+end
+
+# Make an rpc_non_blocking_request to pxp-agent
+# Reference: https://github.com/puppetlabs/pcp-specifications/blob/master/pxp/request_response.md
+# @param broker hostname or beaker host object of the machine running PCP broker
+# @param targets array of PCP identities to send request to
+#                e.g. ["pcp://client01.example.com/agent","pcp://client02.example.com/agent"]
+# @param pxp_module which PXP module to call, default pxp-module-puppet
+# @param action which action in the PXP module to call, default run
+# @param params params to send to the module. e.g for pxp-module-puppet:
+#               {:env => [], :flags => ['--noop', '--onetime', '--no-daemonize']}
+# @return hash of responses, with target identities as keys, and value being the response message as a hash
+#         Responses should be of message type http://puppetlabs.com/rpc_provisional_response
+# @raise String indicating something went wrong, test case can use to fail
+def rpc_non_blocking_request(broker, targets,
+                         pxp_module = 'pxp-module-puppet', action = 'run',
+                         params)
+  rpc_request(broker, targets, pxp_module, action, params, false)
+end
+
+def rpc_request(broker, targets,
+                pxp_module = 'pxp-module-puppet', action = 'run',
+                params, blocking)
   mutex = Mutex.new
   have_response = ConditionVariable.new
   responses = Hash.new
@@ -265,16 +289,20 @@ def rpc_blocking_request(broker, targets,
   end
 
   message = PCP::Message.new({
-    :message_type => 'http://puppetlabs.com/rpc_blocking_request',
+    :message_type => blocking ? 'http://puppetlabs.com/rpc_blocking_request' : 'http://puppetlabs.com/rpc_non_blocking_request',
     :targets => targets
   })
 
-  message.data = {
+  message_data = {
     :transaction_id => SecureRandom.uuid,
     :module         => pxp_module,
     :action         => action,
     :params         => params
-  }.to_json
+  }
+  if !blocking then
+    message_data[:notify_outcome] = false
+  end
+  message.data = message_data.to_json
 
   message_expiry = 10 # Seconds for the PCP message to be considered failed
   rpc_action_expiry = 60 # Seconds for the entire RPC action to be considered failed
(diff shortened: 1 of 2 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -265,16 +289,20 @@ def rpc_blocking_request(broker, targets,
   end
 
   message = PCP::Message.new({
-    :message_type => 'http://puppetlabs.com/rpc_blocking_request',
+    :message_type => blocking ? 'http://puppetlabs.com/rpc_blocking_request' : 'http://puppetlabs.com/rpc_non_blocking_request',
     :targets => targets
   })
 
-  message.data = {
+  message_data = {
```
parisiale: For the sake of dealing with restarts, would it be useful to print the transaction_id? I think you should in case you can receive responses related to previous runs.

You're printing all responses (from the on_message handler), but you don't actually enforce that they're all related to the current transaction.

parisiale:
@er0ck , could you please have a look at the new test?
I had only a comment about keeping track of transactions in the helper wrapper for connecting / sending / receiving.

acceptance/lib/pxp-agent/test_helper.rb
james-stocks: Yes these helper methods assume that the messages hitting the handler are the response for the expected transaction. This is reasonably safe as the client identity only exists for the duration of this method; so it shouldn't receive messages for anything other than this RPC action.

I will raise a separate ticket for the fact that we do not have full traceability of the transaction_id; it's code that existed before this ticket and shouldn't be addressed in this PR.

parisiale: 👍

acceptance/tests/pxp-module-puppet/restart_pxp_agent_during_non_blocking_run.rb
```diff
@@ -0,0 +1,133 @@
...
+MODULE_NAME = 'sleep'
+SECONDS_TO_SLEEP = 500 # The test will use SIGALARM to end this as soon as required
+STATUS_QUERY_MAX_RETRIES = 60
+STATUS_QUERY_INTERVAL_SECONDS = 5
+
+test_name 'C94705 - Run Puppet (non-blocking request) and restart pxp-agent service during run' do
+
+  applicable_agents = agents.select { |agent| agent['platform'] !~ /win/}
+  if applicable_agents.empty? then
+    skip_test('Skipping - All agent hosts are Windows and are not applicable to this test (see PCP-276)')
```
er0ck: PCP-276/LTH-50 has resolved.  can we remove the windows skip?

acceptance/tests/pxp-module-puppet/restart_pxp_agent_during_non_blocking_run.rb
```diff
@@ -0,0 +1,133 @@
...
+          on(agent 'kill -9 `cat /opt/puppetlabs/puppet/cache/state/agent_catalog_run.lock`')
+        end
+      end
+    end
+  end
+
+  step 'On master, create a new environment with a module that will take some time' do
+    environmentpath = master.puppet['environmentpath']
+    site_manifest = "#{environmentpath}/#{ENVIRONMENT_NAME}/manifests/site.pp"
+    on(master, "cp -r #{environmentpath}/production #{environmentpath}/#{ENVIRONMENT_NAME}")
```
er0ck: i've found it helpful to have these environments created in host.tmpdir and then linked into `environmentpath`.
this way they can be removed in the teardown, not pollute subsequent test setup, and still be used for debugging tests.
host.tmpdir also uniquely names the dir for idempotency.

acceptance/tests/pxp-module-puppet/restart_pxp_agent_during_non_blocking_run.rb
```diff
@@ -0,0 +1,133 @@
...
+MODULEPP
+    on(master, "chmod 644 #{module_manifest}")
+  end
+
+  step 'Ensure each agent host has pxp-agent running and associated' do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+      on agent, puppet('resource service pxp-agent ensure=running enable=true')
+      show_pcp_logs_on_failure do
```
er0ck: nice

acceptance/tests/pxp-module-puppet/restart_pxp_agent_during_non_blocking_run.rb
james-stocks: PCP-276 was resolved as "Won't Do"; there continues to be a limitation here

acceptance/tests/pxp-module-puppet/restart_pxp_agent_during_non_blocking_run.rb
james-stocks: Richard's work. Nice indeed.

acceptance/tests/pxp-module-puppet/restart_pxp_agent_during_non_blocking_run.rb
james-stocks: That does sound neater; particularly if we want to move towards some way of getting all the puppet-agent tests onto a single set of hosts. Will give this a go now

er0ck: note, i have one solution for this in PR-405.
https://github.com/puppetlabs/pxp-agent/pull/405

james-stocks:
@er0ck - the test has been updated to use a temp folder for the environment; symlinked to the proper environment path.
Eventually it should use theEnvironmentUtils module you are submitting with PCP-228. If there's no other feedback on this PR; I'd propose merging this PR as-is then creating a new ticket to update this test and the other existing test(s) to use EnvironmentUtils

er0ck:
sounds good to me.  maybe a different ticket for tech-debt, as needed.

PR #401, opened 2016-04-25
(PCP-379) Store expanded --spool-dir in HW singleton

parisiale:
After changes for PCP-345, Configuration did not update the spool-dir
entry with the expanded path in the HorseWhisperer singleton; this
was causing external modules runs to fail to write in the results
directory.


mruzicka:
LGTM

PR #404, opened 2016-04-27
(maint) Look for .dylib in Findcpp-pcp-client.cmake

parisiale:
This is needed on OS X after PCP-372.


MikaelSmith:
`master` works without modification for me.
```
- Found cpp-pcp-client: /Users/michaelsmith/puppetlabs/deps/lib/libcpp-pcp-client.dylib
```

The explicit version is only necessary when we're not using the system default, so we could potentially actually get rid of specifying `libcpp-pcp-client.so` eventually.

PR #405, opened 2016-04-27
(pcp-228) pxp agent acceptance:  run puppet and expect failure

er0ck:


er0ck:
@james-stocks FYI, herein is my version of the tmp_environment stuff i use in puppet and orch tests.  in this case i also add a teardown registration to remove the environment link.

er0ck:
looks like we could abstract the stuff here where we ensure a running pxp-agent and associated agents.
it's used verbatim in many tests.
probably could do something similar in how we send (non)blocking requests to the broker.

i could refactor this and other tests if we think that'd be valuable in this time frame.

acceptance/lib/puppet/acceptance/environment_utils.rb
```diff
@@ -0,0 +1,44 @@
...
+        while @@tmp_environment_set.include?(tmp_environment = environment.downcase + '_' + random_string) do
+          break if (loop_num = loop_num + 1) > deadman
+        end
+        @@tmp_environment_set << tmp_environment
+        tmpdir = File.join('','tmp',tmp_environment)
+
+        on master, "mkdir -p #{tmpdir}/manifests #{tmpdir}/modules"
+
+        # register teardown to remove the link below
+        teardown do
```
james-stocks: Does this teardown block apply to any beaker test instance that includes this module? I didn't know we could do that, cool :)
It does mean that the module is presuming what the test coder will want. Since all tests are supposed to be isolated; the assumption here to remove the symlink is reasonable

acceptance/lib/puppet/acceptance/environment_utils.rb
```diff
@@ -0,0 +1,44 @@
...
+    module EnvironmentUtils
+
+     # generate a random string of 6 letters and numbers.  NOT secure
+      def random_string
+        [*('a'..'z'),*('0'..'9')].shuffle[0,8].join
+      end
+      private :random_string
+
+      # if the first test to call this has changed the environmentpath, this will cause trouble
+      #   maybe not the best idea to memoize this?
```
james-stocks: It does seem potentially troublesome to memorise the environmentpath. As a test developer; I would (like to) assume that helper lib methods are stateless.

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,53 @@
+require 'pxp-agent/test_helper.rb'
+require 'puppet/acceptance/environment_utils'
+
+test_name 'C93063 - pxp-module-puppet run expect puppet failure' do
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  app_type = File.basename(__FILE__, '.*')
```
james-stocks: Does this get the filename, so we can name the temp environment after the test name? It might be useful to add a code comment here; or rename app_type to something like "test_file_name"

I think (at least irb allows this) you can also simply refer to $0 (arguable if that is more or less readable though)

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,53 @@
...
+  tmp_environment   = mk_tmp_environment(app_type)
+
+  step 'On master, create a new environment that will result in Failed' do
+    site_manifest = File.join(environmentpath, tmp_environment, 'manifests', 'site.pp')
+    create_remote_file(master, site_manifest, <<-SITEPP)
```
james-stocks: In my experience, create_remote_file does not work for Puppet environment files unless followed by a chmod command - does this work OK?

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,53 @@
...
+                                      })
+    rescue => exception
+      fail("Exception occurred when trying to run Puppet on all agents: #{exception.message}")
+    end
+    agents.each do |agent|
+      step "Check Run Puppet response for #{agent}" do
+        identity = "pcp://#{agent}/agent"
+        action_result = responses[identity][:data]["results"]
+        assert(action_result.has_key?('status'), "Results for pxp-module-puppet run on #{agent} should contain a 'status' field")
+        assert_equal('failed', action_result['status'], "Result of pxp-module-puppet run on #{agent} should be 'failed'. was: '#{action_result["status"]}'")
```
james-stocks: Possibly an assertion on the error message "intended failure" could be done here; to ensure that we got our planned failure instead of some other failure

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,53 @@
...
+                                                              '--onetime']
+                                      })
+    rescue => exception
+      fail("Exception occurred when trying to run Puppet on all agents: #{exception.message}")
+    end
+    agents.each do |agent|
+      step "Check Run Puppet response for #{agent}" do
+        identity = "pcp://#{agent}/agent"
+        action_result = responses[identity][:data]["results"]
+        assert(action_result.has_key?('status'), "Results for pxp-module-puppet run on #{agent} should contain a 'status' field")
```
james-stocks: The existing tests may be missing this; but it would be valuable to have an assertion (before asserting the result) that the environment in the results is indeed our temp environment. There are some cases where if Puppet is not happy, it will switch from your designated environment back to Production (then give result Unchanged instead of Failure); and a clear failure on the environment not being used will save having to troubleshoot why the result was wrong.

.gitignore
```diff
@@ -4,3 +4,5 @@
 /lib/tests/resources/test_spool/
 .idea/
 .DS_Store
+acceptance/empty.rb
```
james-stocks: Is this a file that might commonly appear in the directory?

acceptance/Rakefile
```diff
@@ -18,16 +18,11 @@ module HarnessOptions
     :type => 'aio',
     :helper => ['lib/helper.rb'],
     :tests  => ['tests'],
-    :log_level => 'info',
```
james-stocks: Thanks for cleaning this up; I didn't know enough about beaker to be sure which options were useless

acceptance/Rakefile
```diff
@@ -18,16 +18,11 @@ module HarnessOptions
...
-    :color => false,
-    :root_keys => true,
     :ssh => {
       :keys => ["~/.ssh/id_rsa-acceptance"],
     },
-    :xml => true,
     :timesync => false,
```
james-stocks: Would enabling this make beaker call timesync on all hosts? If so (and *only* if QENG-3786 is fixed), we should set this to true then delete setup/common/005_SyncTime.rb

acceptance/lib/puppet/acceptance/environment_utils.rb
```diff
@@ -0,0 +1,44 @@
...
+        @@memoized_environmentpath ||= master.puppet['environmentpath']
+      end
+
+      # create a tmpdir to hold a temporary environment bound by puppet environment naming restrictions
+      # symbolically link environment into environmentpath
+      def mk_tmp_environment(environment)
+        # add the tmp_environment to a set to ensure no collisions
+        @@tmp_environment_set ||= Set.new
+        deadman = 100; loop_num = 0
+        while @@tmp_environment_set.include?(tmp_environment = environment.downcase + '_' + random_string) do
```
james-stocks: beaker.tmpdir could/should be used here. It's not perfect because its permissions need adjusted after creating the dir; and I'm not sure it gives any guarantee against random string collisions; but it is a beaker method that is having its functionality repeated here.

acceptance/lib/puppet/acceptance/environment_utils.rb
er0ck: yep

acceptance/lib/puppet/acceptance/environment_utils.rb
er0ck: i think there is a tradeoff here between speed/efficiency and state.  every call to host.puppet['blah'] runs puppet on the host with --configprint.  this is slow.

i think this is a good option, when the vast majority just wants the default environmentpath

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
er0ck: yeah, works fine. i've never had to chmod the files/dirs.

.gitignore
er0ck: i create empty.rb a lot so i can run the pre-suites from the Rake tasks without running any tests.  
`TESTS=empty.rb`

acceptance/Rakefile
er0ck: good call.  i'll add a comment there as to why it's false.  

I'd really prefer the rakefile to be as dumb as possible, so people can easily run from either beaker or the Rakefile.  we've had this issue on other projects.
the rakefile can provide sugar to help with some tasks like host generator, and abstracting the CI job config back into this repo.

acceptance/lib/puppet/acceptance/environment_utils.rb
er0ck: this is a great idea.  but unfortunately tmpdir creates file paths with characters that are not allowed in environment names.

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
er0ck: good idea, but the fail message doesn't come back in the responses:
```
[1] pry(#<Beaker::TestCase>)> responses
=> {"pcp://a2uylov6sb4vc6m.delivery.puppetlabs.net/agent"=>
  {:envelope=>
    {:id=>"e236526e-04bb-4c03-b06f-f11bce4b8a88",
     :targets=>["pcp://shell/ruby-pcp-client-6207"],
     :message_type=>"http://puppetlabs.com/rpc_blocking_response",
     :sender=>"pcp://a2uylov6sb4vc6m.delivery.puppetlabs.net/agent",
     :expires=>"2016-04-28T21:31:34.192626Z"},
   :data=>
    {"transaction_id"=>"73b45fc8-d76c-4021-8feb-05a08d0eb62b",
     "results"=>
      {"kind"=>"apply",
       "time"=>"2016-04-28 14:31:30 -0700",
       "transaction_uuid"=>"8e7d50ef-4ab1-41e5-878b-97a4b8c81dbc",
       "environment"=>"run_puppet_result_failed_bt3j2knm",
       "status"=>"failed",
       "exitcode"=>1,
       "version"=>1,
       "error_type"=>"agent_exit_non_zero",
       "error"=>"Puppet agent exited with a non 0 exitcode"}}}}
```

we could inspect the agent logs, but this is actually very difficult to do correctly, particularly on windows agents.

er0ck:
@james-stocks 
updated based on feedback.

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
james-stocks: Using the fail() function is a little different to what this test case intended - it causes a parsing error creating the catalog, and the agent errors out because it did not get a catalog to run with....

`[root@wlu11s2tuopmhnu ~]# puppet agent -t --environment failure
Info: Using configured environment 'failure'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Function Call, Intended failure at /etc/puppetlabs/code/environments/failure/manifests/site.pp:2:3 on node wlu11s2tuopmhnu.delivery.puppetlabs.net
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
[root@wlu11s2tuopmhnu ~]# echo $?
1`

This is different to a failure occurring during the catalog run e.g. if I try to ensure a file inside a directory that doesn't exist:

`[root@wlu11s2tuopmhnu ~]# puppet agent -t --environment failure
Info: Using configured environment 'failure'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for wlu11s2tuopmhnu.delivery.puppetlabs.net
Info: Applying configuration version '1461933225'
Error: Could not set 'file' on ensure: No such file or directory @ rb_sysopen - /tmp/badfolder/jimmy at 2:/etc/puppetlabs/code/environments/failure/manifests/site.pp
Error: Could not set 'file' on ensure: No such file or directory @ rb_sysopen - /tmp/badfolder/jimmy at 2:/etc/puppetlabs/code/environments/failure/manifests/site.pp
Wrapped exception:
No such file or directory @ rb_sysopen - /tmp/badfolder/jimmy
Error: /Stage[main]/Main/Node[default]/File[/tmp/badfolder/jimmy]/ensure: change from absent to file failed: Could not set 'file' on ensure: No such file or directory @ rb_sysopen - /tmp/badfolder/jimmy at 2:/etc/puppetlabs/code/environments/failure/manifests/site.pp
Notice: Applied catalog in 0.01 seconds
[root@wlu11s2tuopmhnu ~]# echo $?
4`

I think there are two potential test cases here; using fail() shows what PXP does if the catalog compilation fails; the latter case shows what PXP does if a failure occurs during catalog execution


btw: +1 on not inspecting logs unless absolutely necessary

james-stocks:
@er0ck I added a comment reply against an outdated diff up there ^^^

I did not specify in Test Rail what "failure" means; sorry. In my head, I was thinking of a catalog run that contained failed resources when the agent applied it. This PR tests a failure in catalog compilation, the agent never gets to run

er0ck:
@james-stocks no worries.  maybe this test could/should do both?

i added another request with an exec resource calling `/bin/false`, which of course fails on windows, but fails in the same way O:)

er0ck:
i believe this is ready to go, once CI goes green

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,94 @@
...
+    agents.each do |agent|
+      step "Check Run Puppet response for #{agent}" do
+        identity = "pcp://#{agent}/agent"
+        action_result = responses[identity][:data]["results"]
+        assert(action_result.has_key?('status'),
+               "Results for pxp-module-puppet run on #{agent} should contain a 'status' field")
+        assert_equal(tmp_environment2, action_result['environment'],
+                     "Result of pxp-module-puppet run on #{agent} should run with the "\
+                     "#{tmp_environment2} environment")
+        assert_equal('failed', action_result['status'],
```
james-stocks: It may be valuable to assert the exit code in this case; but when I had a quick look last week, I noticed it was possibly coming out wrong - I'll raise a separate ticket to look into that instead of dragging out this ticket longer

acceptance/Rakefile
```diff
@@ -96,7 +91,8 @@ EOS
 
   overriding_options = ENV['OPTIONS']
 
-  args = ["--options-file", options_file, config_opt, tests_opt, overriding_options].compact
+  args = ["--options-file", options_file, config_opt, tests_opt,
```
james-stocks: This causes an error when I try to run

    Jamess-MacBook-Pro:acceptance jstocks$ rake ci:test:aio
    Adding Redhat 7 master to ENV['TEST_TARGET']. Value is now 'redhat7-64m-redhat7-64a'
    ./config/aio/options.rb
    ./local_options.rb
    rake aborted!
    NoMethodError: undefined method `split' for nil:NilClass
    /Users/jstocks/GitHub/cthun-agent/acceptance/Rakefile:95:in `beaker_test'
    /Users/jstocks/GitHub/cthun-agent/acceptance/Rakefile:276:in `block (3 levels) in <top (required)>'
    /Users/jstocks/.rvm/gems/ruby-2.1.1/bin/ruby_executable_hooks:15:in `eval'
    /Users/jstocks/.rvm/gems/ruby-2.1.1/bin/ruby_executable_hooks:15:in `<main>'
    Tasks: TOP => ci:test:aio
    (See full trace by running task with --trace)

Is this because the new code does not handle ENV['OPTIONS'] being unset?

acceptance/Rakefile
```diff
@@ -18,16 +18,11 @@ module HarnessOptions
     :type => 'aio',
     :helper => ['lib/helper.rb'],
     :tests  => ['tests'],
-    :log_level => 'info',
-    :color => false,
```
james-stocks: Yay, colors!

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+    end
+  end
+
+  step "Send an rpc_blocking_request to all agents for #{tmp_environment2}" do
+    responses = nil # Declare here so not local to begin/rescue below
+    begin
+      responses = rpc_blocking_request(master, target_identities,
+                                      'pxp-module-puppet', 'run',
+                                      {:env => [], :flags => ['--environment', tmp_environment2,
```
james-stocks: This doesn't work for me:
    [root@rvmgw0nbcg7tghl ~]# puppet agent -t --environment environment2
    Info: Using configured environment 'environment2'
    Info: Retrieving pluginfacts
    Info: Retrieving plugin
    Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not parse for environment environment2: Permission denied - /etc/puppetlabs/code/environments/environment2/manifests/site.pp on node rvmgw0nbcg7tghl.delivery.puppetlabs.net
    Warning: Not using cache on failed catalog
    Error: Could not retrieve catalog; skipping run

I think permissions need modified for site.pp after using create_remote_file to create it
chmod 644 site.pp results in:

    [root@rvmgw0nbcg7tghl ~]# puppet agent -t --environment environment2
    Info: Using configured environment 'environment2'
    Info: Retrieving pluginfacts
    Info: Retrieving plugin
    Info: Caching catalog for rvmgw0nbcg7tghl.delivery.puppetlabs.net
    Info: Applying configuration version '1462359992'
    Error: /bin/false returned 1 instead of one of [0]
    Error: /Stage[main]/Main/Node[default]/Exec[/bin/false]/returns: change from notrun to 0 failed: /bin/false returned 1 instead of one of [0]
    Notice: Applied catalog in 0.02 seconds

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,94 @@
...
+      step "Check Run Puppet response for #{agent}" do
+        identity = "pcp://#{agent}/agent"
+        action_result = responses[identity][:data]["results"]
+        assert(action_result.has_key?('status'),
+               "Results for pxp-module-puppet run on #{agent} should contain a 'status' field")
+        assert_equal(tmp_environment2, action_result['environment'],
+                     "Result of pxp-module-puppet run on #{agent} should run with the "\
+                     "#{tmp_environment2} environment")
+        assert_equal('failed', action_result['status'],
+                     "Result of pxp-module-puppet run on #{agent} should be 'failed'. was: "\
```
james-stocks: I  think if the comment above is addressed; the result JSON will contain both a PXP action result and Puppet agent results, with the action result being "success" and the Puppet agent result being "failed" - this test should probably assert both

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,94 @@
...
+    site_manifest = File.join(environmentpath, tmp_environment, 'manifests', 'site.pp')
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  fail('intended failure')
+}
+SITEPP
+    site_manifest2 = File.join(environmentpath, tmp_environment2, 'manifests', 'site.pp')
+    create_remote_file(master, site_manifest2, <<-SITEPP)
+node default {
+  exec{'/bin/false':}
```
james-stocks: After we resolve the comment below about running the environment; it would be worth double checking that this still results in an equivalent failure on Windows (and I think some posix platforms will be /usr/bin/false) - I *think* it will; as Puppet will get to the point of catalog execution and consider this resource a failed resource

acceptance/Rakefile
er0ck: oh.  looks like i need to rebase

acceptance/Rakefile
er0ck: i don't believe so.  we have color turned on in other jenkins jobs.  we turn color output off in some clients under tests (eg: orch-client) so we can easily parse the output

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
er0ck: i've seen this fail as well, but only on some platforms, but only on some tests.  so strange.  
what layout/master were you using?

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
er0ck: i could also just change it to 'nonesuchbinary'

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
james-stocks: redhat7-64m and redhat7-64a

I didn't complain about beaker's methods creating files with restricted permissions; as I assumed it was done deliberately as security best practice. Maybe it's actually a beaker bug. I suppose ideally beaker should always create files with the same permissions you'd get if you manually ssh'd in and created the file

er0ck: strange.  works for me here.  i'm running against stable

er0ck:
updated based on feedback. @james-stocks 
i did not change anything to fix your environment permissions failures.  
i'm hoping it's a beaker issue in an old beaker on your system (see comments above)

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
james-stocks: I have 2.40.0

From looking at beaker's code; I'm suspicious that when I try this code snippet:

    2.1.1 :001 > require 'tempfile'
     => true
    2.1.1 :002 > file = Tempfile.new('foo')
     => #<Tempfile:/var/folders/86/dcqmwc991h50ksqbq_l7z5400000gn/T/foo20160505-87693-mgtbka>
    2.1.1 :003 > file.path
     => "/var/folders/86/dcqmwc991h50ksqbq_l7z5400000gn/T/foo20160505-87693-mgtbka"

I get these permissions:

    Jamess-MacBook-Pro:~ jstocks$ ls -ltra /var/folders/86/dcqmwc991h50ksqbq_l7z5400000gn/T/foo20160505-87693-mgtbka
    -rw-------  1 jstocks  staff  0  5 May 10:51 /var/folders/86/dcqmwc991h50ksqbq_l7z5400000gn/T/foo20160505-87693-mgtbka

In case it was just my Macbook; I also tried on the shell box:

    james.stocks@shell:~$ ls -ltr /tmp/foo20160505-4223-a5q33g
    -rw------- 1 james.stocks allstaff 0 May  5 03:07 /tmp/foo20160505-4223-a5q33g

Would you mind trying the same snippet and seeing if you get the same or more permissive permissions? I think beyond this; beaker just uses net-scp with no options (so I don't see an opportunity to get inconsistent results).

If this continues to be a problem (for me); could we include a chmod command so it always works; then have a ticket for me to investigate; fix and tidy up?

james-stocks:
I sent you an email @er0ck so we can discuss offline from this PR; I maybe see what's happening here (maybe)

acceptance/tests/pxp-module-puppet/run_puppet_result_failed.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def assert_failed_agent_run(environment, action_result)
+    assert(action_result.has_key?('status'),
+           "Results for pxp-module-puppet run on #{agent} should contain a 'status' field")
+    assert_equal(environment, action_result['environment'],
+                 "Result of pxp-module-puppet run on #{agent} should run with the "\
+                 "#{environment} environment")
+    assert_equal('unchanged', action_result['status'],
```
er0ck: very strange.  so this definitely does not exit 0 when puppet is run.  but pxp reports unchanged. 
similar with the `error_type` i removed below.

```
[root@fn9m821ucm8kkdi ~]# puppet apply -e "exec{'nonesuchbinary':}"
Notice: Compiled catalog for fn9m821ucm8kkdi.delivery.puppetlabs.net in environment production in 0.07 seconds
Error: Validation of Exec[nonesuchbinary] failed: 'nonesuchbinary' is not qualified and no path was specified. Please qualify the command or specify a path. at line 1
[root@fn9m821ucm8kkdi ~]# echo $?
1
```

er0ck:
sheesh.  so, it would appear that (default?) manifests need to have execute permissions to be found/used.  

@james-stocks this should be ready to go

PR #406, opened 2016-04-28
(maint) Enable builds with static cpp-pcp-client

MikaelSmith:
On Windows, Leatherman's logging/locale libraries need to be contained in
the same DLL. That can be done by either making them shared libraries, or
by creating a monolithic pxp-agent binary.

Currently our packaging makes it messy to use shared Leatherman DLLs, and
they don't bring us any advantages. Make it simple to use cpp-pcp-client
as a static library instead so we can produce a smaller package.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -54,6 +54,12 @@ set (LIBS
     ${LEATHERMAN_LIBRARIES}
 )
 
+if (WIN32)
+    # Necessary when statically linking cpp-pcp-client on Windows.
+    # Shouldn't hurt when cpp-pcp-client is a DLL.
+    list(APPEND LIBS Ws2_32)
+endif()
+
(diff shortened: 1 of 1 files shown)


mruzicka:
LGTM

lib/CMakeLists.txt
```diff
@@ -54,6 +54,12 @@ set (LIBS
     ${LEATHERMAN_LIBRARIES}
 )
 
+if (WIN32)
+    # Necessary when statically linking cpp-pcp-client on Windows.
+    # Shouldn't hurt when cpp-pcp-client is a DLL.
```
mruzicka: Can't we just test what kind of cpp-pcp-client library we are linking against? Whether static or shared?

MikaelSmith: I'm not sure there's a way to tell from here. /cc @branan 

I could add a CMake config file to the cpp-pcp-client project that gets installed, and would replace the Findcpp-pcp-client script in this project. That would include link dependencies and whether it's static or dynamic. However there's pretty much no downside with linking an extra system library.

PR #411, opened 2016-05-02
(PCP-371) Spike i18n support

parisiale:
Enable the creation of the .pot file that includes the strings that
we want to translate. Don't enable localization now; we'll do that on
PCP-257.

Use leatherman::locale::format() and translate() for the strings that we
want to translate.

Use "{}" instead of "%%" for string interpolation.

Don't use leatherman::util::plural(); we'll deal with plurals once we
enable localization on PCP-257.

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -6,7 +6,7 @@
 #include <cpp-pcp-client/util/thread.hpp>
 #include <cpp-pcp-client/util/chrono.hpp>
 
-#include <leatherman/file_util/file.hpp>
+#include <leatherman/locale/locale.hpp>
 
 #define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.main"
 #include <leatherman/logging/logging.hpp>
@@ -21,7 +21,7 @@
 namespace PXPAgent {
 
 namespace HW = HorseWhisperer;
-namespace lth_file = leatherman::file_util;
+namespace lth_loc = leatherman::locale;
 
 // Exit code returned after a successful execution
 static int PXP_AGENT_SUCCESS = 0;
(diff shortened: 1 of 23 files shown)


exe/main.cc
```diff
@@ -124,9 +124,10 @@ int main(int argc, char *argv[]) {
     }
     if (!err_msg.empty()) {
         tryLogError(err_msg);
-        boost::nowide::cout << err_msg
-                            << "\nCannot start pxp-agent"
-                            << std::endl;
+        boost::nowide::cout
+            << err_msg << "\n"
```
MikaelSmith: If logging configuration fails, there's a decent chance translation will also fail. I'm not sure it will do so gracefully.

lib/src/module.cc
```diff
@@ -83,20 +87,16 @@ ActionResponse Module::executeAction(const ActionRequest& request)
...
-        err_msg += "Error: ";
-        err_msg += e.what();
+        err_msg += lth_loc::format("Error: {1}", e.what());
     } catch (std::exception& e) {
-        err_msg += "Unexpected error: ";
-        err_msg += e.what();
+        err_msg += lth_loc::format("Unexpected error: {1}", e.what());
     } catch (...) {
-        err_msg = "Unexpected excption.";
+        err_msg = lth_loc::translate("Unexpected excption.");
```
MikaelSmith: Typo in `excption`.

lib/src/util/posix/pid_file.cc
```diff
@@ -210,11 +210,9 @@ bool PIDFile::canLockFile(int fd, int lock_type) {
     fl.l_start = 0;
     fl.l_len = 0;
 
-    if (fcntl(fd, F_GETLK, &fl) == -1) {
+    if (fcntl(fd, F_GETLK, &fl) == -1)
         // Failure
-        std::string msg { "errno=" + std::to_string(errno) };
-        throw PIDFile::Error { msg };
-    }
+        throw PIDFile::Error { lth_loc::format("errno={1}", errno) };
```
MikaelSmith: :( I'm surprised our linter lets this pass. Oh well, maybe it catches if you have two indented lines.

parisiale:
Fixed typo in module.cc.

MikaelSmith:
👍 I'll merge after CI passes.

MikaelSmith:
AppVeyor started failing independently, and I'm pretty sure the last push to this PR didn't cause the failure. Merging this, and I'll review the AppVeyor failures separately.

PR #415, opened 2016-05-04
(PCP-381) Fix usage of env values in acceptance Rakefile

james-stocks:
ci:test:aio was failing in the puppet-agent pipelines when a static hosts file (e.g. aix-71-power.yaml) was needed.
This commit updates Rakefile to work with ENV values passed from CI in the same way as all other puppet-agent components.

[skip ci]


james-stocks:
I'm not 100% happy with the code being added here; but the objective is simply to get this repo consistent with other puppet-agent components. For any feedback on code added in this commit; a PA ticket should be raised to apply the improvement across all component repos

PR #416, opened 2016-05-04
(PCP-306) Use last_run_report mtime to check for updates

MikaelSmith:
Previously last_run_report would be assumed to use a newer time stamp if
the triggered Puppet run successfully updated the report. NTP updates
during the Puppet run could potentially violate that assumption. Switch
to checking mtime on the report, and assuming if mtime has changed that
a new report was written.


MikaelSmith:
/cc @nicklewis

The only case where I see this getting us into trouble is if the environment undergoes some sort of rollback or someone manually copies an old report into place after a run. Both of those seem unlikely.

MikaelSmith:
AppVeyor won't pass until https://github.com/puppetlabs/pxp-agent/pull/414 is merged.

parisiale:
AppVeyor is happy after #414 fix.
👍 

Pinging @nicklewis.

parisiale:
Also, pinging @mruzicka.

nicklewis:
This approach makes sense to me.

parisiale:
Will merge as soon as puppet-agent CI is back to green.

PR #420, opened 2016-05-06
(PCP-383) Add broker failover

MikaelSmith:
Adds the failover-ws-uri option. When specified, will use that as a fallback when the primary broker is unavailable. To avoid split-brain installations, only allow one broker to be active at a time.


MikaelSmith:
Requires https://github.com/puppetlabs/cpp-pcp-client/pull/161.

PR #422, opened 2016-05-10
(maint) Restore plural messages

MikaelSmith:
Until Leatherman adds plural i18n helpers, we don't want a release to
backtrack on pluralizing message strings. Restore basic pluralization in
a way similar to what the i18n helpers will do.

### DIFF
--- a/lib/src/results_storage.cc
+++ b/lib/src/results_storage.cc
@@ -262,8 +262,13 @@ unsigned int ResultsStorage::purge(
         });
 
     // TODO(ale): deal with locale & plural (PCP-257)
-    LOG_INFO("Removed {1} directories from '{2}'",
-             num_purged_dirs, spool_dir_path_.string());
+    if (num_purged_dirs == 1) {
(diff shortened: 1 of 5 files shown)


MikaelSmith:
@kylog

lib/src/results_storage.cc
```diff
@@ -262,8 +262,13 @@ unsigned int ResultsStorage::purge(
         });
 
     // TODO(ale): deal with locale & plural (PCP-257)
-    LOG_INFO("Removed {1} directories from '{2}'",
-             num_purged_dirs, spool_dir_path_.string());
+    if (num_purged_dirs == 1) {
+        LOG_INFO("Removed {1} directorie from '{2}'",
```
kylog: directory

kylog:
LGTM, other than the one typo'd singular.

lib/src/results_storage.cc
MikaelSmith: Fixed.

parisiale:
👍

PR #424, opened 2016-05-10
(maint) Pin pcp-broker version to 0.6.1

MikaelSmith:
Known issues in 0.7.0 seem to be causing testing to be unstable. Pin
pxp-agent#stable testing to version 0.6.1 until resolved.

/cc @james-stocks @er0ck 

It'd be nice to have a better way to pin stable builds. Also, why are we using git rather than ezbake helpers?


MikaelSmith:
Not sure this is ok yet. Got a setup failure trying to run acceptance.

PR #427, opened 2016-05-16
(PCP-349) support running puppet as non-root

mruzicka:
This commit makes it possible to run puppet under a non root user.
Previously this was not possible because puppet failed to expand the user home directory. This was due to the environment cleaning performed by the `Puppet::Util::Execution.execute` function used for the puppet run execution which included unsetting of the `HOME` environment variable.
This patch solves that by fixing the environment (re-adding the necessary environment variables to it) after the cleaning. The values of the re-added variables are obtained by reading the pwentry corresponding to the current effective user id.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -28,7 +28,9 @@ DEFAULT_EXITCODE = -1
 
 env_fix_up = nil
 
-def get_env_fix_up
+# use this way of defining the `get_env_fix_up` method so that it can "see"
+# the `env_fix_up` variable
+self.class.send(:define_method, :get_env_fix_up) do
   # Prepare an environment fix-up to make up for its cleansing performed
   # by the Puppet::Util::Execution.execute function.
   # Without this fix-up it's impossible to run puppet under a non-root
@@ -45,8 +47,11 @@ def get_env_fix_up
       {"USER"    => pwentry.name,
        "LOGNAME" => pwentry.name,
        "HOME"    => pwentry.dir}
-    rescue
+    rescue => e
       # oh well ..., let's give it a try without the environment fix-up
+      myname = File.basename($0)
+      $stderr.puts "#{myname}: Could not fix environment for effective UID #{Process.euid}: #{e.message}"
+      $stderr.puts "#{myname}: Expect puppet run problems"
       {}
     end
   end
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -26,14 +26,30 @@ end
 
 DEFAULT_EXITCODE = -1
 
-# This method exists to facilitate testing
-def last_child_exit_status
-  return $?
-end
+env_fix_up = nil
 
-# This method exists to facilitate testing
-def is_win?
-  return !!File::ALT_SEPARATOR
+def get_env_fix_up
+  # Prepare an environment fix-up to make up for its cleansing performed
+  # by the Puppet::Util::Execution.execute function.
+  # Without this fix-up it's impossible to run puppet under a non-root
+  # user because it cannot find the user's HOME directory.
+  env_fix_up ||= if Puppet.features.microsoft_windows? || Process.euid == 0
+    # no environment fix-up is needed on windows or for root
+    {}
+  else
+    begin
+      require 'etc'
+
+      pwentry = Etc.getpwuid(Process.euid)
+
+      {"USER"    => pwentry.name,
+       "LOGNAME" => pwentry.name,
+       "HOME"    => pwentry.dir}
+    rescue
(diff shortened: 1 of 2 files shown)


modules/pxp-module-puppet
```diff
@@ -26,14 +26,30 @@ end
...
 
-# This method exists to facilitate testing
-def last_child_exit_status
-  return $?
-end
+env_fix_up = nil
+
+def get_env_fix_up
+  # Prepare an environment fix-up to make up for its cleansing performed
+  # by both pxp-agent and the Puppet::Util::Execution.execute function.
```
parisiale: Where does pxp-agent modify the environment?

mruzicka: Hmm, perhaps when it's daemonizing?
When I was logging the environment the ```pxp-module-puppet``` gets, I think it already didn't contain the ```USER``` environment variable.

mruzicka: Hmm no, I've tried again and it's there. So probably I just overlooked it. I'll update the comments.

parisiale: If you want to do so with POSIX's daemonize(), you need some extra code. We don't do that; forked processes inherit the parent's env.
Besides, if I log {{LOG_INFO("daemonization complete; HOME = {1}", std::getenv("HOME"));}} just after daemonize(), I get this when I run pxp-agent as `alessandro`:
`
2016-05-16 14:16:10.534428 INFO  puppetlabs.pxp_agent.main:57 - daemonization complete; HOME = /Users/alessandro
`

Also, pxp-agent should not be changing the environment when it calls leatherman's execute(): https://github.com/puppetlabs/pxp-agent/blob/master/lib/src/external_module.cc#L360

We should look into that if it's really the case.

modules/pxp-module-puppet
```diff
@@ -26,14 +26,30 @@ end
...
+      pwentry = Etc.getpwuid(Process.euid)
 
-# This method exists to facilitate testing
-def is_win?
-  return !!File::ALT_SEPARATOR
+      {"USER"    => pwentry.name,
+       "LOGNAME" => pwentry.name,
+       "HOME"    => pwentry.dir}
+    rescue
+      # oh well ..., let's give it a try without the environment fix-up
```
parisiale: Can we write something on stderr?

modules/pxp-module-puppet
```diff
@@ -48,7 +64,8 @@ end
 
 def check_config_print(cli_arg, config)
   command_array = [config["puppet_bin"], "agent", "--configprint", cli_arg]
-  process_output = Puppet::Util::Execution.execute(command_array)
+  process_output = Puppet::Util::Execution.execute(command_array,
+                                                   {:custom_environment => get_env_fix_up()})
```
parisiale: Is the new env dictionary gonna be merged with the process' actual env?

parisiale:
@MikaelSmith please have a look

modules/pxp-module-puppet
parisiale: Just for clarity, the two above messages went out of sync... :)

modules/pxp-module-puppet
mruzicka: yes:
https://github.com/puppetlabs/puppet/blob/4.4.2/lib/puppet/util/execution.rb#L297
and:
https://github.com/puppetlabs/puppet/blob/4.4.2/lib/puppet/util.rb#L102-L135

modules/pxp-module-puppet
mruzicka: k

modules/pxp-module-puppet
parisiale: 👍

mruzicka:
@MikaelSmith please don't merge this before I squash it

parisiale:
Travis is not happy with the `modules` test:
```
/home/travis/build/puppetlabs/pxp-agent/modules/pxp-module-puppet:33:in `<top (required)>': undefined method `define_method' for main:Object (NoMethodError)
```

mruzicka:
Fixed for older ruby versions.

parisiale:
Different error now:
```
     NoMethodError:
       undefined method `get_env_fix_up' for #<RSpec::ExampleGroups::PxpModulePuppet::StartRun:0x00000002aae360>
     # ./pxp-module-puppet:92:in `make_environment_hash'
     # ./pxp-module-puppet:149:in `start_run'
     # ./spec/unit/modules/puppet_spec.rb:239:in `block (3 levels) in <top (required)>'
```

mruzicka:
So hopefully now.

MikaelSmith:
That looks reasonable. 👍

PR #430, opened 2016-05-24
(PCP-428) restart_host_run_puppet.rb transient fails on ubuntu1404

er0ck:
This change tries to make the restart_host test more robust by using two
sleeps.  One to ensure the agent is actually rebooted (see BKR-812) and
one to retry up to 10 times until pxp-agent is back up.

[skip ci]

### DIFF
--- a/acceptance/tests/restart_host_run_puppet.rb
+++ b/acceptance/tests/restart_host_run_puppet.rb
@@ -22,13 +22,26 @@ test_name 'C94777 - Ensure pxp-agent functions after agent host restart' do
   step "restart each agent" do
     applicable_agents.each do |agent|
       agent.reboot
-      assert(agent.up?, "Agent #{agent} should be up after reboot")
-      on(agent, puppet('resource service pxp-agent ')) do |result|
-        assert_match(/ensure => .running.,/, result.stdout,
-                     "pxp-agent service should be running after reboot")
+      # BKR-812
+      begin
+        Timeout.timeout(10) do
+          until agent.up?
+            sleep 1
+          end
+        end
+      rescue Timeout::Error => e
+        raise "Agent did not come back up within #{timeout} seconds."
       end
-      assert(is_associated?(master, "pcp://#{agent}/agent"),
-                            "Agent #{agent} should be associated with pcp-broker following host reboot")
+
+      step "wait until pxp-agent is back up and associated on #{agent}" do
+        opts = {
+          :max_retries => 10,
+          :retry_interval => 1,
+        }
+        retry_on(agent, 'puppet resource service pxp-agent | grep running', opts)
+        assert(is_associated?(master, "pcp://#{agent}/agent"),
+               "Agent #{agent} should be associated with pcp-broker following host reboot")
+      end
     end
   end
 


acceptance/tests/restart_host_run_puppet.rb
```diff
@@ -22,8 +22,14 @@
...
       agent.reboot
+      # BKR-812
+      sleep 1 if !agent.up?
       assert(agent.up?, "Agent #{agent} should be up after reboot")
-      on(agent, puppet('resource service pxp-agent ')) do |result|
+      opts = {
+        :max_retries => 10,
+        :retry_interval => 1,
+      }
+      retry_on(agent, puppet('resource service pxp-agent '), opts) do |result|
```
james-stocks: This retry will not help for the case where pxp-agent service reports as being "ensure=>stopped" after startup (PCP-393). In that case, this retry_on will succeed on its first attempt, with the result being "ensure=>stopped".

acceptance/tests/restart_host_run_puppet.rb
```diff
@@ -22,8 +22,14 @@
   step "restart each agent" do
     applicable_agents.each do |agent|
       agent.reboot
+      # BKR-812
+      sleep 1 if !agent.up?
```
james-stocks: Why not sleep for longer, or test agent.up? multiple times? I'm not sure it's possible to test that 1 second sleep will make a difference for the Ubuntu 14.04 problem; and there's a risk that at some point any other host might need longer

acceptance/tests/restart_host_run_puppet.rb
er0ck: ah.  you are right.  i suspect we are getting lucky with just a single retry required, in both these changes.

acceptance/tests/restart_host_run_puppet.rb
er0ck: it worked locally.  test repeatedly fails without it and repeatedly passes with it.  but i can do some fall-back type stuff.  we just need to get CI green so we can ship PA 1.5.1

acceptance/tests/restart_host_run_puppet.rb
er0ck: interestingly this test fails locally with the current stable PA build (12841f398f1ea187fccda2a0ed99cd505df7de2a) without this retry_on, and works with it.  i'm not quite sure why.

er0ck:
@james-stocks updated

acceptance/tests/restart_host_run_puppet.rb
james-stocks: Not sure... if you fixed Ubuntu 14.04 so we no longer fail 'host.up?' then maybe we moved to the point where the host wasn't quite ready to complete the 'on' command yet?

acceptance/tests/restart_host_run_puppet.rb
```diff
@@ -22,13 +22,26 @@
...
-                     "pxp-agent service should be running after reboot")
+      # BKR-812
+      begin
+        Timeout.timeout(10) do
+          until agent.up?
+            sleep 1
+          end
+        end
+      rescue Timeout::Error => e
+        raise "Agent did not come back up within #{timeout} seconds."
```
james-stocks: I won't hold up merging for this, but is 'timeout' defined?

PR #432, opened 2016-05-27
(PUP-428) fix restart_host test on windows agents

er0ck:
The previous commit works around some flaky host and pxp-agent restart
issues by using retry_on(), but did not take into account puppet command
pathing issues in windows.  this commit fixes this.

### DIFF
--- a/acceptance/tests/restart_host_run_puppet.rb
+++ b/acceptance/tests/restart_host_run_puppet.rb
@@ -23,7 +23,7 @@ test_name 'C94777 - Ensure pxp-agent functions after agent host restart' do
     applicable_agents.each do |agent|
       agent.reboot
       # BKR-812
-      timeout = 10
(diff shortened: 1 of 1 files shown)


james-stocks:
Can the commit be re-written - I think the ticket reference should be to PCP instead of PUP (or just (maint))

acceptance/tests/restart_host_run_puppet.rb
```diff
@@ -39,7 +39,7 @@
           :max_retries => 10,
           :retry_interval => 1,
         }
-        retry_on(agent, 'puppet resource service pxp-agent | grep running', opts)
+        retry_on(agent, puppet('resource service pxp-agent | grep running'), opts)
```
james-stocks: +1 on this, have tested it myself locally

PR #435, opened 2016-06-02
(PCP-423) Add `allowed-keepalive-timeouts` option

MikaelSmith:
Adds the `allowed-keepalive-timeouts` option, with a default of 2. The
option is used to decide how many consecutive keepalive responses can be
lost before closing the connection and trying to reconnect. With the
default of 2, the connection will be closed after 3 lost keepalive
responses; the agent will then attempt to reconnect to the broker and/or
failover broker.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -284,7 +284,8 @@ const Configuration::Agent& Configuration::getAgentConfiguration() const
         AGENT_CLIENT_TYPE,
         HW::GetFlag<int>("connection-timeout") * 1000,
         static_cast<uint32_t >(HW::GetFlag<int>("association-timeout")),
-        static_cast<uint32_t >(HW::GetFlag<int>("pcp-message-timeout")) };
+        static_cast<uint32_t >(HW::GetFlag<int>("pcp-message-timeout")),
+        static_cast<uint32_t >(HW::GetFlag<int>("allowed-keepalive-timeouts")) };
     return agent_configuration_;
 }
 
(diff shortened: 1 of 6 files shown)


MikaelSmith:
Depends on https://github.com/puppetlabs/cpp-pcp-client/pull/178.

lib/src/configuration.cc
```diff
@@ -358,6 +359,17 @@ void Configuration::defineDefaultValues()
                     Types::Int,
                     5) } });
 
+    defaults_.insert(
+        Option { "allowed-keepalive-timeouts",
+                 Base_ptr { new Entry<int>(
+                    "allowed-keepalive-timeouts",
+                    "",
+                    lth_loc::translate("Number of consecutive keepalive response "
```
MikaelSmith: Not sure about making this public, as using it also requires knowing that keepalives happen every 15 seconds with a 5 second timeout. Tempted to make it hidden as a default, because hopefully people won't need to tweak it.

@kylog any further thoughts?

lib/src/configuration.cc
```diff
@@ -358,6 +359,17 @@ void Configuration::defineDefaultValues()
...
+                                       "timeouts to accept before assuming the "
```
kylog: Drop "assuming the" here? Or I'm not quite parsing the message.

MikaelSmith: Oops, I changed my mind mid-way through that sentence. It's all going away for the `<hidden>` tag though.

lib/src/configuration.cc
```diff
@@ -358,25 +359,35 @@ void Configuration::defineDefaultValues()
                     Types::Int,
                     5) } });
 
+    // Hidden option: number of ping/pong timeouts to allow before disconnecting, default: 2
+    defaults_.insert(
+        Option { "allowed-keepalive-timeouts",
```
MikaelSmith: This doesn't match how cpp-pcp-client thinks of the option, but I think it's easier to describe succinctly.

parisiale: I agree.

parisiale:
👍

lib/src/configuration.cc
MikaelSmith: Thought of another label: `keepalive-failure-threshold`.

PR #436, opened 2016-06-02
(PCP-453) Only kill pxp-agent if it's running

sarameisburger:
If logrotate ran and pxp-agent wasn't running it would fail to kill it. Now it
will only try to kill it if it's running.


sarameisburger:
@MikaelSmith can you take a look at this?

PR #437, opened 2016-06-03
(PCP-454) Block execution thread once monitoring starts

parisiale:
The new PCPClient::Connector that starts the monitoring thread does not
block anymore, hence we not enter in an idle loop once we started the
connection monitoring thread.


parisiale:
This needs https://github.com/puppetlabs/cpp-pcp-client/pull/179

lib/src/agent.cc
```diff
@@ -63,13 +63,17 @@ void Agent::start() {
...
         // Note that, in case the underlying connection drops, the
         // connector will keep trying to re-establish it indefinitely
         // (the max_connect_attempts is 0 by default).
         LOG_DEBUG("PCP connection established; about to start monitoring it");
-        connector_ptr_->monitorConnection();
+        connector_ptr_->startMonitoring();
+        pcp_util::chrono::seconds pause { 1 };
+
+        while (true)
+            pcp_util::this_thread::sleep_for(pause);
```
MikaelSmith: I'm not sure why this pause doesn't delay shutdown when I do `Ctrl-C`. But it seems to work fine.

MikaelSmith: Oh, because we pretty much hard stop on SIGINT. It looks like only a connection error might cause it to stop.

parisiale: As a side note, the agent was born like that: https://github.com/puppetlabs/pxp-agent/blob/34ea1d01ec449bbeb524f7d4a8e46b21c22fde55/src/agent.cpp#L166-L168

PR #438, opened 2016-06-07
(PCP-452) Add option to configure Association timeout

parisiale:
The `association-timeout` option is now used to set the timeout for the
overall Association process (default to 15 s).

The old `association-timeout` (ttl for the Association request) can now
be specified by the `association-request-ttl` option (default to 10 s).

Renaming the `pcp-message-timeout` option to 'pcp-messag-ttl', for
consistency with the other ttl option.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -284,7 +284,8 @@ const Configuration::Agent& Configuration::getAgentConfiguration() const
         AGENT_CLIENT_TYPE,
         HW::GetFlag<int>("connection-timeout") * 1000,
         static_cast<uint32_t >(HW::GetFlag<int>("association-timeout")),
-        static_cast<uint32_t >(HW::GetFlag<int>("pcp-message-timeout")),
+        static_cast<uint32_t >(HW::GetFlag<int>("association-request-ttl")),
+        static_cast<uint32_t >(HW::GetFlag<int>("pcp-message-ttl")),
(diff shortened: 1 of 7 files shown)


parisiale:
Requires https://github.com/puppetlabs/cpp-pcp-client/pull/180.

lib/src/configuration.cc
```diff
@@ -818,9 +830,13 @@ void Configuration::validateAndNormalizeOtherSettings()
         throw Configuration::Error {
             lth_loc::translate("association-timeout must be positive") };
 
-    if (HW::GetFlag<int>("pcp-message-timeout") < 0)
+    if (HW::GetFlag<int>("association-request-ttl") < 0)
+        throw Configuration::Error {
+                lth_loc::translate("association-request-ttl must be positive") };
```
mruzicka: Is zero legal? The test does allow it.

lib/src/configuration.cc
```diff
@@ -818,9 +830,13 @@ void Configuration::validateAndNormalizeOtherSettings()
...
+
+    if (HW::GetFlag<int>("pcp-message-ttl") < 0)
         throw Configuration::Error {
-            lth_loc::translate("pcp-message-timeout must be positive") };
+            lth_loc::translate("association-request-ttl must be positive") };
```
mruzicka: Same as above.

parisiale: If you configure 0, you should know what you're doing.

mruzicka:
LGTM

PR #439, opened 2016-06-10
(PCP-463) Restore blocking monitor action

MikaelSmith:
Revert to using `monitorConnection`. It ensures exceptions thrown during
monitoring are propagated to the caller.


parisiale:
Requires https://github.com/puppetlabs/cpp-pcp-client/pull/182.

PR #440, opened 2016-06-16
(maint) Add libpxp-agent dependency on horsewhisperer

MikaelSmith:
This avoids compilation failure during parallel builds (as in `make -j`)
when horsewhisperer is unpacked after starting to compile source for
libpxp-agent.


parisiale:
👍

PR #442, opened 2016-06-28
(PCP-405) Make broker-ws-uris a config file only option

MikaelSmith:
Changes broker-ws-uris to only be settable via a config file. Also
clarifies how it and broker-ws-uri interract, and adds documentation in
the README.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -109,6 +109,7 @@ void Configuration::initialize(
     HW::SetAppName("pxp-agent");
     HW::SetHelpBanner(lth_loc::translate("Usage: pxp-agent [options]"));
     HW::SetVersion(std::string { PXP_AGENT_VERSION } + "\n");
+    broker_ws_uris_.clear();
     valid_ = false;
 
     // Initialize boost filesystem's locale to a UTF-8 default.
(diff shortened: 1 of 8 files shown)


MikaelSmith:
The conclusion on how to handle multiple values on the command-line was either go comma-separated, or avoid it altogether. I went with avoid it.

An alternative would be to use comma-separated, and not introduce a new setting. Rather, `broker-ws-uri` could take either a string, or an array of strings. I'm not sure that's a great idea, as I think JSON schema would not support specifying multiple possible types for a property. However, the implementation would be quite a bit simpler, and shouldn't be ambiguous in our usage.

P.S. In general Websocket URIs could include commas in a query string, but I don't foresee us having a query string as part of broker URIs.

lib/src/configuration.cc
```diff
@@ -618,13 +606,22 @@ void Configuration::parseConfigFile()
...
     for (const auto& key : config_json.keys()) {
         const auto& opt_idx = defaults_.get<Option::ByName>().find(key);
 
-        if (opt_idx == defaults_.get<Option::ByName>().end())
-            throw Configuration::Error {
-                lth_loc::format("field '{1}' is not a valid configuration variable",
-                                key) };
+        if (opt_idx == defaults_.get<Option::ByName>().end()) {
+            if (key == "broker-ws-uris") {
+                check_key_type(key, "Array", lth_jc::DataType::Array);
```
parisiale: Perhaps we should add a comment about `broker-ws-uris` being a configfile-only option.

parisiale:
👍 
Just a comment request, as Configuration is pretty complex.

MikaelSmith:
Added the comment.

PR #444, opened 2016-06-28
(PCP-431) pxp agent acceptance should support a failover pcp broker

er0ck:
also: 
(maint) acceptance: move helper config option out of Rakefile
* make default log_level verbose for beaker

### DIFF
--- a/acceptance/lib/pxp-agent/config_helper.rb
+++ b/acceptance/lib/pxp-agent/config_helper.rb
@@ -2,9 +2,10 @@ require 'json'
 # This file contains common strings and functions for working with pxp-agent's config file
 
 PXP_CONFIG_DIR_CYGPATH = '/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/etc/'
-PXP_CONFIG_DIR_POSIX = '/etc/puppetlabs/pxp-agent/'
+PXP_CONFIG_DIR_POSIX   = '/etc/puppetlabs/pxp-agent/'
 
-PCP_BROKER_PORT = 8142
+PCP_BROKER_PORTS       = [8142, 8143]
+PCP_BROKER_REPL_PORTS  = [7888, 7889]
 
 def windows?(host)
   host.platform.upcase.start_with?('WINDOWS')
@@ -66,7 +67,7 @@ end
 # @param broker_host the host name or beaker host object for the pcp-broker host
 # @return the broker-ws-uri config string
 def broker_ws_uri(broker_host)
-  "wss://#{broker_host}:#{PCP_BROKER_PORT}/pcp/"
+  "wss://#{broker_host}:#{PCP_BROKER_PORTS[broker_host[:pcp_broker_instance]]}/pcp/"
 end
 
 # @param host the beaker host (to determine the correct path for the OS)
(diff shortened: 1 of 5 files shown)


acceptance/setup/common/050_Setup_Broker.rb
```diff
@@ -29,20 +52,18 @@
...
-    on master, "\\cp #{puppet_ssldir}/ca/ca_crt.pem #{broker_ssldir}/ca/ca_crt.pem"
-    on master, "\\cp #{puppet_ssldir}/ca/ca_crl.pem #{broker_ssldir}/ca/ca_crl.pem"
+  for i in 0..NUM_BROKERS-1
+    on master, "cd #{GIT_CLONE_FOLDER}/pcp-broker#{i}; export LEIN_ROOT=ok; lein deps"
   end
 end
 
-step "Run pcp-broker in trapperkeeper in background and wait for it to report status 'running'" do
-  run_pcp_broker(master)
+#TODO:   see if we can use instance 1 in the tests
```
MikaelSmith: When do you plan to revisit this TODO?

er0ck:
oops.  i did that.  artifact removal, commencing...

acceptance/setup/common/050_Setup_Broker.rb
```diff
@@ -2,19 +2,42 @@
...
+
+step 'Clone pcp-broker to broker_hosts' do
   clone_git_repo_on(master, GIT_CLONE_FOLDER,
     extract_repo_info_from(build_git_url('pcp-broker', nil, nil, 'https')))
+
+  step 'Replace PCP test certs with the Puppet certs of this SUT' do
+    on(master, puppet('config print ssldir')) do |result|
+      puppet_ssldir = result.stdout.chomp
+      broker_ssldir = "#{GIT_CLONE_FOLDER}/pcp-broker/test-resources/ssl"
+      on master, "\\cp #{puppet_ssldir}/certs/#{master}.pem #{broker_ssldir}/certs/broker.example.com.pem"
```
MikaelSmith: I don't understand the leading backslashes, but they were there before so 🐑

er0ck: yeah, i left them in to be safe.  they are meant to remove any alias that exists on some machines.  beaker should probably do this for us...

acceptance/lib/pxp-agent/config_helper.rb
```diff
@@ -76,3 +77,22 @@ def pxp_agent_config_file(host)
...
+def configure_pcp_broker(host,instance)
+  set_broker_port(host,instance)
+  set_broker_nrepl_port(host,instance)
+end
+
+def set_broker_port(host,instance)
+  broker_config_path = "#{GIT_CLONE_FOLDER}/pcp-broker#{instance}/test-resources/conf.d/webserver.conf"
+  port_config_name  = 'ssl-port'
+  broker_port = PCP_BROKER_PORTS[instance]
+  on(host, "sed --in-place'' --regexp-extended 's/#{port_config_name}:\\s?[0-9]+/#{port_config_name}: #{broker_port}/' #{broker_config_path}")
```
MikaelSmith: This could be ruby-hocon, but seems safe enough.

er0ck: i agree.  i think it's a balance between adding more dependencies and the relative ease/safety of hand-parsing this.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -69,10 +69,11 @@ def show_pcp_logs_on_failure(&block)
 end
 
 # Some helpers for working with a pcp-broker 'lein tk' instance
-def run_pcp_broker(host)
-  on(host, "cd #{GIT_CLONE_FOLDER}/pcp-broker; export LEIN_ROOT=ok; lein tk </dev/null >/var/log/pcp-broker.log 2>&1 &")
-  assert(port_open_within?(host, PCP_BROKER_PORT, 60),
-         "pcp-broker port #{PCP_BROKER_PORT.to_s} not open within 1 minutes of starting the broker")
+def run_pcp_broker(host, instance=0)
+  host[:pcp_broker_instance] = instance
```
MikaelSmith: If we're adding a property to the `host`, why not add the port itself? You could use that as the unique identifier for machines as well.

The one tricky part about that is probably using run_pcp_broker within tests.

er0ck: right.  my thoughts were that the 'user' (test author) should never need to know what the ports are, just which instance they want to use.

PR #445, opened 2016-06-29
(PCP-472) Ignore pxp-module-puppet `env` input arg

parisiale:
Ignoring the `env` array to avoid letting the user to set custom
environment variables as we currently don't have a use case for that.

The `env` entry of the JSON metadata's input object remains defined
for backwards compatibility, but it's not a required entry anymore.
Thus, pxp-module-puppet interface remains the same; in the future we may
want to completely remove the `env` entry, but that would require
bumping the interface version.

Updating unit tests.

Updating docs.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -33,8 +33,8 @@ env_fix_up = nil
 self.class.send(:define_method, :get_env_fix_up) do
   # Prepare an environment fix-up to make up for its cleansing performed
   # by the Puppet::Util::Execution.execute function.
-  # Without this fix-up it's impossible to run puppet under a non-root
-  # user because it cannot find the user's HOME directory.
+  # This fix-up is meant for running puppet under a non-root user;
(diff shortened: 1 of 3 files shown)


modules/pxp-module-puppet
```diff
@@ -33,8 +33,8 @@ env_fix_up = nil
 self.class.send(:define_method, :get_env_fix_up) do
   # Prepare an environment fix-up to make up for its cleansing performed
   # by the Puppet::Util::Execution.execute function.
-  # Without this fix-up it's impossible to run puppet under a non-root
-  # user because it cannot find the user's HOME directory.
+  # This fix-up is meant for running puppet under a non-root user;
+  # pxp-module-puppet cannot find the user's HOME directory otherwise.
```
mruzicka: it's thge puppet agent who crashes if it can't find the user's HOME directory, not the pxp-module-puppet

mruzicka:
except for the minor comment this LGTM

parisiale:
updated

PR #448, opened 2016-07-05
(PCP-479) Update pxp-agent to install batch wrapper

McdonaldSeanp:
The pxp-module-puppet batch wrapper needs to be installed on windows platforms
to correctly load the module.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,13 @@ install(FILES modules/pxp-module-puppet
         DESTINATION ${MODULES_INSTALL_PATH}
         PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
 
+if (WIN32)
+# Add the module install batch file on windows platforms
+install(FILES modules/pxp-module-puppet.bat
+        DESTINATION ${MODULES_INSTALL_PATH}
+        PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+endif()
+
 # Add the test suite
 enable_testing()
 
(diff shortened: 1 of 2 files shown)


modules/pxp-module-puppet.bat
```diff
@@ -3,4 +3,4 @@ SETLOCAL
 
 call "%~dp0..\..\bin\environment.bat" %0 %*
 
-ruby -S -- "%~dp0%SCRIPT_NAME%" %*
+ruby -S -- "%~dp0\pxp-module-puppet" %*
```
MikaelSmith: Why did this need to change?

CMakeLists.txt
```diff
@@ -98,6 +98,13 @@ install(FILES modules/pxp-module-puppet
         DESTINATION ${MODULES_INSTALL_PATH}
         PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
 
+if (WIN32)
```
MikaelSmith: 👍

MikaelSmith:
Can you make this a PCP ticket? We'll need to track releasing it for 1.6.0.

modules/pxp-module-puppet.bat
McdonaldSeanp: For some reason SCRIPT_NAME wouldn't resolve to anything. For the other batch files we are installing for puppet-agent we have also just started hard coding the name. Is there a reason to keep SCRIPT_NAME and not change it?

MikaelSmith: Not that I know of; @joshcooper probably wrote them and may have more context.

McdonaldSeanp:
https://tickets.puppetlabs.com/browse/PCP-479

modules/pxp-module-puppet.bat
McdonaldSeanp: Great.

PR #450, opened 2016-07-11
(PCP-430) Automate pxp-agent broker failover acceptance test

er0ck:
do not merge prior to PR-444

* configure agents for multiple broker uris
* ensure association with primary broker
* stop first broker, start replica
* ensure association with replica

[skip ci]

### DIFF
--- a/acceptance/tests/failover/intentional_shutdown_failover.rb
+++ b/acceptance/tests/failover/intentional_shutdown_failover.rb
@@ -1,0 +1,50 @@
+require 'pxp-agent/config_helper.rb'
+require 'pxp-agent/test_helper.rb'
+
+test_name 'C97934 - agent should use next broker if primary is intentionally shutdown'
+
+teardown do
+  broker_instance = 0 # 0 indexed
+  kill_pcp_broker(master)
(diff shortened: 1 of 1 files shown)


acceptance/tests/failover/intentional_shutdown_failover.rb
```diff
@@ -0,0 +1,50 @@
...
+
+test_name 'C97934 - agent should use next broker if primary is intentionally shutdown'
+
+teardown do
+  broker_instance = 0 # 0 indexed
+  kill_pcp_broker(master)
+  run_pcp_broker(master, broker_instance)
+end
+
+def pxp_config_hash_using_puppet_certs_multiple_uris(broker, agent)
```
MikaelSmith: This seems like it might be needed in test_helper eventually.

MikaelSmith:
Seems legit 👍

acceptance/tests/failover/intentional_shutdown_failover.rb
er0ck: agreed, that's why i abstracted it... somewhat.  but yagni rules apply, so i left it here

PR #451, opened 2016-07-13
(PCP-431) fix previous acceptance commit

er0ck:
add workaround for tests which need a working broker URI but may not
have started it yet.

[skip ci]


er0ck:
i don't understand why this has conflicts.  git thinks it's up to date on my end, looks like it is here too.

PR #454, opened 2016-07-15
(PCP-457) Acceptance: Automate pxp-agent broker timeout failover test

er0ck:
[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -178,7 +178,7 @@ def pcp_broker_inventory(broker, query)
       end
     end
   rescue Timeout::Error
-    raise "Didn't receive a response for PCP inventory request"
+    raise "Didn't receive a response for PCP inventory request in #{inventory_expiry} seconds"
   ensure
     client.close
   end # wait for message
@@ -200,7 +200,8 @@ end
 # @param retries the number of times to retry checking the inventory. Default 60 to allow for slow test VMs. Set to 0 to only check once.
 # @return if the identity is in the broker's inventory within the allowed number of retries
 #         false if the identity remains absent from the broker's inventory after the allowed number of retries
-def is_associated?(broker, identity, retries = 60)
+PCP_INVENTORY_RETRIES = 60
+def is_associated?(broker, identity, retries = PCP_INVENTORY_RETRIES)
   if retries == 0
     return pcp_broker_inventory(broker,identity).include?(identity)
   end
@@ -220,7 +221,7 @@ end
 # @param retries the number of times to retry checking the inventory. Default 60 to allow for slow test VMs. Set to 0 to only check once.
 # @return true if the identity is absent from the broker's inventory within the allowed number of retries
 #         false if the identity persists in the broker's inventory after the allowed number of retries
-def is_not_associated?(broker, identity, retries = 60)
+def is_not_associated?(broker, identity, retries = PCP_INVENTORY_RETRIES)
   if retries == 0
     return !pcp_broker_inventory(broker,identity).include?(identity)
   end
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -95,6 +95,15 @@ def kill_all_pcp_brokers(host)
   end
 end
 
+def block_pcp_broker(host, instance)
+  on(host, "iptables -A INPUT -p tcp --destination-port #{PCP_BROKER_PORTS[instance].to_s} -j DROP; iptables -S")
+end
+
+def unblock_pcp_broker(host, instance)
+  host[:pcp_broker_instance] = instance
+  on(host, "iptables -D INPUT -p tcp --destination-port #{PCP_BROKER_PORTS[instance].to_s} -j DROP; iptables -S")
+end
+
 def get_pcp_broker_status(host)
   uri = URI.parse("https://#{host}:#{PCP_BROKER_PORTS[host[:pcp_broker_instance]]}/status/v1/services/broker-service")
   begin
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -86,10 +86,12 @@ def run_pcp_broker(host, instance=0)
   assert_equal("running", broker_state, "Shortly after startup, pcp-broker should report its state as being 'running'")
 end
 
-def kill_pcp_broker(host)
+def kill_all_pcp_brokers(host)
   on(host, "ps -C java -f | grep pcp-broker | sed 's/[^0-9]*//' | cut -d\\  -f1") do |result|
-    pid = result.stdout.chomp
-    on(host, "kill -9 #{pid}")
+    pids = result.stdout.chomp.split("\n")
+    pids.each do |pid|
+      on(host, "kill -9 #{pid}")
+    end
   end
 end
 
--- a/acceptance/tests/failover/intentional_shutdown_failover.rb
+++ b/acceptance/tests/failover/intentional_shutdown_failover.rb
@@ -1,50 +1,40 @@
 require 'pxp-agent/config_helper.rb'
 require 'pxp-agent/test_helper.rb'
 
-test_name 'C97934 - agent should use next broker if primary is intentionally shutdown'
+test_name 'C97934 - agent should use next broker if primary is intentionally shutdown' do
 
-teardown do
-  broker_instance = 0 # 0 indexed
-  kill_pcp_broker(master)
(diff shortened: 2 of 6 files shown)


er0ck:
@MikaelSmith, others.  this works, but requires pxp-agent to be restarted after the primary broker's port is blocked.  
the timeout on association checks is set to 60 seconds, but i waited at least 3 minutes on one occasion with no change to association on the failover broker.  

how long should we have to wait, and/or should i have to restart/HUP the pxp-agent?

MikaelSmith:
The agent should failover in ~1 minute, so there's a chance we broke something.

MikaelSmith:
File name has a typo.

acceptance/tests/failover/intentional_shutdown_failover.rb
```diff
@@ -1,50 +1,41 @@
...
 
-def pxp_config_hash_using_puppet_certs_multiple_uris(broker, agent)
-  broker_uris = [broker_ws_uri(master), broker_ws_uri(master)]
-  broker_uris[1].sub!(PCP_BROKER_PORTS[0].to_s,PCP_BROKER_PORTS[1].to_s)
+  step 'Ensure each agent host has pxp-agent configured with multiple uris, running and associated' do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      num_brokers = 2
+      pxp_config = pxp_config_hash_using_puppet_certs(master, agent, num_brokers)
+      pxp_config['allowed-keepalive-timeouts'] = 0
```
MikaelSmith: `allowed-keepalive-timeouts` should have no effect on this test.

MikaelSmith:
👍

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -88,11 +88,23 @@ def run_pcp_broker(host, instance=0)
 
 def kill_pcp_broker(host)
   on(host, "ps -C java -f | grep pcp-broker | sed 's/[^0-9]*//' | cut -d\\  -f1") do |result|
-    pid = result.stdout.chomp
-    on(host, "kill -9 #{pid}")
+    pids = result.stdout.chomp.split("\n")
```
james-stocks: Minor comment: This method might now need renamed "kill_pcp_brokers" and/or have an optional argument to select which brokers to kill

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -88,11 +88,23 @@ def run_pcp_broker(host, instance=0)
...
+    pids.each do |pid|
+      on(host, "kill -9 #{pid}")
+    end
   end
 end
 
+def block_pcp_broker(host, instance)
+  host[:pcp_broker_instance] = instance+1
```
james-stocks: What's the point of line 99? If host[:pcp_broker_instance] stores the broker index that should be considered the current broker, is it correct for this method to change it? We might have HA tests where we want to block a broker that is neither the current nor the next broker we want to use.

And is the +1 because we're incrementing, or because we're moving from 0-index to 1-index? If it's the former, then does this method work when we're on the last broker and need to cycle back round? If it's the latter, then is line 104 correct?

acceptance/tests/failover/timeout_failover.rb
```diff
@@ -0,0 +1,44 @@
...
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      num_brokers = 2
+      pxp_config = pxp_config_hash_using_puppet_certs(master, agent, num_brokers)
+      pxp_config['allowed-keepalive-timeouts'] = 0
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config.to_json.to_s)
+      on(agent, "rm -rf #{logfile(agent)}")
+      on agent, puppet('resource service pxp-agent ensure=running')
+      show_pcp_logs_on_failure do
+        assert(is_associated?(master, "pcp://#{agent}/agent"),
```
james-stocks: There are two brokers in play now; what exactly is being asserted here? From reading the recent changes to the helper code; I can see that this call to is_associated? will go and use broker 0; but I think the test could be more readable and state that it is asserting association on the primary broker.

It might also be diligent to assert that the agent is not associated on the replica broker. I don't currently know any possibility how it might report as associated on both; but if we ever got into that weird state, it would be nicer to fail the test now rather than get a false result

acceptance/tests/failover/timeout_failover.rb
```diff
@@ -0,0 +1,44 @@
...
+               "Agent identity pcp://#{agent}/agent for agent host #{agent} does not appear in pcp-broker's client inventory")
```
james-stocks: Related to the comment above, the assertion string should specify which broker was being checked, now that 2 brokers are in play

acceptance/tests/failover/timeout_failover.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    end
+  end
+
+  step 'Block primary broker, start replica' do
+    block_pcp_broker(master,blocked_broker_instance)
```
james-stocks: "blocked_broker_instance" may be a confusing variable name, as at the start of the test, nothing is blocked. After a broker is blocked, the name makes sense but then it's not used again.
Perhaps the test could declare PRIMARY_BROKER=0; FAILOVER_BROKER=1 and then remove the "blocked_broker_instance" and "broker_instance" variables completely? Just call block_pcp_broker(master, PRIMARY_BROKER; run_pcp_broker(master, FAILOVER_BROKER)

acceptance/tests/failover/timeout_failover.rb
```diff
@@ -0,0 +1,44 @@
...
+    broker_instance = 1 # 0 indexed
+    run_pcp_broker(master, broker_instance)
+  end
+
+  step 'On each agent, test that a new association has occurred' do
```
james-stocks: Similar to earlier comment; the step string, assertion statement, and assertion string should all specify that we are asserting association on the failover broker; even though I can see that the helper methods in this case are going to use the failover broker.

At this point it would be valuable to assert that we are not associated with the primary broker any longer. Failover is poor if we did connect to the replica but didn't cleanly disconnect from the primary. That might need to be a separate step, as it would need to be unblocked so it can respond to inventory_requests (or, unless the iptables rules are made more specific so that pxp-agent is cut off from the broker, but ruby-pcp-client can still connect and query the inventory while pxp-agent is cut off)

james-stocks:
There are helper updates, updates to existing tests, and a brand new test case all in a single commit here. Minor comment, as I doubt we'll ever need to pick amongst these changes, but ideally there should be at least a couple of commits.

acceptance/tests/failover/timeout_failover.rb
er0ck: is_associated ends up doing an inventory request, which finds a broker uri, which can only send the request to the correct broker.  so this is sure to show that we are talking to the primary/correct broker.  same as in the reconnect test.  

we can't assert anything on the replica broker, because it's not running at this point.  i suppose it could be.

er0ck:
> there should be at least a couple of commits.

👍

acceptance/tests/failover/timeout_failover.rb
james-stocks: I agree 100% that your code works, my query is just on test readability; understanding this assertion's value requires knowledge of how all the helper code works. 

It could be made obvious by adding a preceding assertion on the value of agent[:pcp_broker_instance] ? It's a functionally pointless assertion, but would make the test readable imo

er0ck: good call.  i like it!

acceptance/tests/failover/timeout_failover.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  step 'Block primary broker, start replica' do
+    block_pcp_broker(master,PRIMARY_BROKER_INSTANCE)
+    run_pcp_broker(master,  REPLICA_BROKER_INSTANCE)
+  end
+
+  step 'On each agent, test that a new association has occurred' do
+    assert_equal(master[:pcp_broker_instance], REPLICA_BROKER_INSTANCE, "broker instance is not set correctly: #{master[:pcp_broker_instance]}")
+    agents.each do |agent|
+      assert(is_associated?(master, "pcp://#{agent}/agent"),
```
james-stocks: In terms of how much time the test allows for failover to occur; I think this line imposes around 60 seconds (after the broker appears to be up) for the agent to failover or else we fail - is that long enough? e.g. if it takes 90 seconds or 120 seconds, do we want this test to show as failed?

er0ck:
ok.  upped the failover timeout to 120 seconds.  ready to go!

PR #461, opened 2016-07-26
(PCP-457) Acceptance: remove primary dis-association check

er0ck:
Previously, this test checked that the primary no longer has our agents
associated.  But this can fail, in a flaky fashion, on windows, when the
broker may or may not receive socket close due to different TCP
defaults.

[skip ci]


MikaelSmith:
@james-stocks going to merge to unblock, but pinging you if you think we still need to test this somehow

PR #462, opened 2016-07-27
(PCP-515) Acceptance - update ticket reference for skipped test on Arista

james-stocks:
We currently skip the host restart test on Arista VMs due to their reboot behaviour.
Fixing this will require some RE work - this commit simply updates the ticket reference in the test from PCP-515 to RE-7673

[skip ci]


parisiale:
Message content change.
👍

PR #464, opened 2016-08-03
(maint) Fix path for pxp-agent using new Windows MSI layout

MikaelSmith:
The Windows MSI layout changed to sharing binaries for C++ applications.
That means pxp-agent.exe now relies on files in puppet/bin. Update the
test to include that in PATH.

For now hard-code in the test. It's not trivial to amend the
privatebindir in config/aio/options.rb, since the setting is
platform-dependent. This might eventually live in Beaker, but since it's
only needed for this test that may not be worth it.


MikaelSmith:
@shrug this is required for the new Windows MSI layout.

@james-stocks can you take a look and merge this asap?

shrug:
👍

PR #465, opened 2016-08-03
(PCP-494) Acceptance test of including non-ASCII character in Puppet manifest

james-stocks:
[skip ci]

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_nonASCII.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_nonASCII.rb
@@ -1,13 +1,14 @@
 require 'pxp-agent/test_helper.rb'
-
-ENVIRONMENT_NAME = 'non_ASCII'
+require 'puppet/acceptance/environment_utils'
 
 test_name 'C98107 - Run puppet with non-ASCII characters in Puppet code' do
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  env_name = test_file_name = File.basename(__FILE__, '.*')
+  environment_name = mk_tmp_environment(env_name)
 
   step 'On master, create a new environment that includes a non-ASCII character and will result in Changed' do
-    environmentpath = master.puppet['environmentpath']
-    site_manifest = "#{environmentpath}/#{ENVIRONMENT_NAME}/manifests/site.pp"
-    on(master, "cp -r #{environmentpath}/production #{environmentpath}/#{ENVIRONMENT_NAME}")
+    site_manifest = "#{environmentpath}/#{environment_name}/manifests/site.pp"
     create_remote_file(master, site_manifest, <<-SITEPP)
 node default {
   notify {'☃':}
@@ -37,7 +38,7 @@ SITEPP
     begin
       responses = rpc_blocking_request(master, target_identities,
                                       'pxp-module-puppet', 'run',
-                                      {:env => [], :flags => ['--environment', ENVIRONMENT_NAME]})
+                                      {:env => [], :flags => ['--environment', environment_name]})
     rescue => exception
       fail("Exception occurred when trying to run Puppet on all agents: #{exception.message}")
     end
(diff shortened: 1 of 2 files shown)


acceptance/tests/pxp-module-puppet/run_puppet_nonASCII.rb
```diff
@@ -0,0 +1,58 @@
+require 'pxp-agent/test_helper.rb'
+
+ENVIRONMENT_NAME = 'non_ASCII'
+
+test_name 'C98107 - Run puppet with non-ASCII characters in Puppet code' do
+
+  step 'On master, create a new environment that includes a non-ASCII character and will result in Changed' do
+    environmentpath = master.puppet['environmentpath']
```
er0ck: we have a helper for this, and this environment needs a teardown.  
actually we should probably just use mk_tmp_environment_on() or whatever it's called in our libs

acceptance/tests/pxp-module-puppet/run_puppet_nonASCII.rb
```diff
@@ -0,0 +1,61 @@
+require 'pxp-agent/test_helper.rb'
+require 'puppet/acceptance/environment_utils'
+
+test_name 'C98107 - Run puppet with non-ASCII characters in Puppet code' do
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  env_name = test_file_name = File.basename(__FILE__, '.*')
+  ENVIRONMENT_NAME = mk_tmp_environment(env_name)
```
er0ck: this should be a variable.

acceptance/tests/pxp-module-puppet/run_puppet_nonASCII.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  step 'On master, create a new environment that includes a non-ASCII character and will result in Changed' do
+    environmentpath = master.puppet['environmentpath']
```
er0ck: no longer needed.  `#environmentpath` is a helper in EnvironmentUtils

acceptance/tests/pxp-module-puppet/run_puppet_nonASCII.rb
```diff
@@ -0,0 +1,61 @@
...
+    site_manifest = "#{environmentpath}/#{ENVIRONMENT_NAME}/manifests/site.pp"
+    on(master, "cp -r #{environmentpath}/production #{environmentpath}/#{ENVIRONMENT_NAME}")
```
er0ck: no longer needed.  the `#mk_tmp_environment_on` links the tmp environment into the environmentpath (but not into production)

er0ck:
let's hold on this until PA ships

acceptance/tests/pxp-module-puppet/run_puppet_result_changed.rb
```diff
@@ -1,13 +1,14 @@
...
+  environment_name = mk_tmp_environment(env_name)
 
   step 'On master, create a new environment that will result in Changed' do
-    environmentpath = master.puppet['environmentpath']
-    site_manifest = "#{environmentpath}/#{ENVIRONMENT_NAME}/manifests/site.pp"
-    on(master, "cp -r #{environmentpath}/production #{environmentpath}/#{ENVIRONMENT_NAME}")
+    site_manifest = "#{environmentpath}/#{environment_name}/manifests/site.pp"
     create_remote_file(master, site_manifest, <<-SITEPP)
 node default {
   notify {'Notify resources cause a Puppet run to have a \\'changed\\' outcome ☃':}
```
MikaelSmith: This has a snow-man in it, so these are both testing non-ASCII. That's what I meant when I said there was an acceptance test (I added the snowman in my PR fixing the issue).

james-stocks: I didn't notice that, sorry.
This case might warrant having its own test case; so in the event of a test run failing we have 1 test result for a Changed puppet run and a separate test result for a Puppet run with non-ASCII characters.
On the other hand; a single test is more efficient in CI.

We can reject this PR; but we should at least update the test case to advertise that it is testing non-ASCII

MikaelSmith: I'm fine with a single test; we're going to run the test a lot and probably break it pretty rarely. Updating it to advertise the non-ASCII portion makes sense.

er0ck:
This duplicates some testing from changes in the upstream PR.
Subsequent changes will update that test to make it more clear that it's testing unicode puppet output and fixup the test to use the env helper.

PR #466, opened 2016-08-05
(PCP-542) Allow parsing last_run_report with Puppet objects

MikaelSmith:
Support parsing a last_run_report.yaml file containing Puppet objects -
sometimes introduced by resource types that don't serialize properly -
by dropping them from the report. We only look at a few known fields, so
it's safe to ignore the rest of the report.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -102,6 +102,20 @@ def make_error_result(exitcode, error_type, error_message)
   return result
 end
 
+def parse_report(filename)
+  # Read the report and drop Ruby objects first. We can't parse the Ruby objects
+  # (because we don't have Puppet loaded) and don't need that data.
+  # Psych::Nodes::Node#each iterates over each node in the parsed document tree.
+  # YAML.parse_file returns a Psych::Nodes::Document, and #root returns the
+  # root-level Node.
+  data = YAML.parse_file(filename)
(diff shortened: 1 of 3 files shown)


MikaelSmith:
This may need to be retargeted depending on when we merge master->stable and 1.2.0 goes out.

parisiale:
Makes sense to me.
@joshcooper or @nicklewis, I could use another pair of eyes here.

modules/pxp-module-puppet
```diff
@@ -102,6 +102,18 @@ def make_error_result(exitcode, error_type, error_message)
   return result
 end
 
+def parse_report(filename)
+  # Read the report and drop Ruby objects first.
+  # We can't parse the Ruby objects (because we don't have Puppet loaded)
+  # and don't need that data.
+  data = YAML.parse_file(filename)
+  data.root.each do |o|
```
nicklewis: It might be helpful to note here that this iterates the entire tree and not just the immediate children, as I assumed on first glance.

nicklewis:
:+1: Looks good to me

modules/pxp-module-puppet
MikaelSmith: Updated the comment; Ruby's YAML defers to a Psych implementation and took a little digging to find the right classes/methods.

PR #467, opened 2016-08-05
(PCP-354) Acceptance - pcp-broker should use internal mirror for deps

james-stocks:
The acceptance test suite relies on being able to download all the dependencies for pcp-broker from clojars.org.
This commit changes the 'lein deps' command to use a new profile that will also add our internal clojar repo URL.

[skip ci]


parisiale:
👍

PR #468, opened 2016-08-05
(PCP-545) Force interpreting Puppet's output as UTF-8 in C/POSIX locale

MikaelSmith:
In a C/POSIX locale, Puppet will still output unicode as UTF-8. Ensure
we parse as UTF-8 despite the locale. Without this change regex
comparison throws an exception when attempting to compare against
ASCII_8BIT output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: cpp
 compiler:
   - gcc
 sudo: false
+ruby: '2.1.9'
 addons:
   apt:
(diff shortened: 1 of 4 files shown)


MikaelSmith:
@james-stocks @parisiale this should fix the remaining acceptance failures, and we should be clear to merge it (won't affect upcoming release).

parisiale:
👍

PR #471, opened 2016-08-12
(PCP-545) Always force Puppet output to UTF-8

MikaelSmith:
Puppet output is always UTF-8 encoded. Even if the locale implies
otherwise, interpret output as UTF-8.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -32,13 +32,19 @@ env_fix_up = nil
 # use this way of defining the `get_env_fix_up` method so that it can "see"
 # the `env_fix_up` variable
 self.class.send(:define_method, :get_env_fix_up) do
+  # If running in a C or POSIX locale, ask Puppet to use UTF-8
+  base_env = {}
+  if Encoding.default_external == Encoding::US_ASCII
+    base_env = {"RUBYOPT" => "#{ENV['RUBYOPT']} -EUTF-8"}
+  end
+
   # Prepare an environment fix-up to make up for its cleansing performed
   # by the Puppet::Util::Execution.execute function.
   # This fix-up is meant for running puppet under a non-root user;
   # puppet cannot find the user's HOME directory otherwise.
   env_fix_up ||= if Puppet.features.microsoft_windows? || Process.euid == 0
     # no environment fix-up is needed on windows or for root
-    {}
+    base_env
   else
     begin
       require 'etc'
@@ -47,13 +53,13 @@ self.class.send(:define_method, :get_env_fix_up) do
 
       {"USER"    => pwentry.name,
        "LOGNAME" => pwentry.name,
-       "HOME"    => pwentry.dir}
+       "HOME"    => pwentry.dir}.merge base_env
     rescue => e
       # oh well ..., let's give it a try without the environment fix-up
       myname = File.basename($0)
       $stderr.puts "#{myname}: Could not fix environment for effective UID #{Process.euid}: #{e.message}"
       $stderr.puts "#{myname}: Expect puppet run problems"
-      {}
+      base_env
     end
   end
 end
@@ -69,11 +75,13 @@ def last_run_result(exitcode)
 end
 
 def force_unicode(s)
-  # Puppet still returns UTF-8 when locale is C or POSIX, so force encoding to that.
-  if s.encoding == Encoding::ASCII_8BIT
+  begin
+    # Later comparisons assume UTF-8. Convert to that encoding now.
+    s.encode(Encoding::UTF_8)
+  rescue Encoding::InvalidByteSequenceError
+    # Found non-native characters, hope it's a UTF-8 string. Since this is Puppet, and
+    # incorrect characters probably means we're in a C or POSIX locale, this is usually safe.
     s.force_encoding(Encoding::UTF_8)
-  else
-    s
   end
 end
 
--- a/modules/spec/unit/modules/puppet_spec.rb
+++ b/modules/spec/unit/modules/puppet_spec.rb
@@ -42,7 +42,7 @@ describe "pxp-module-puppet" do
       expect(Puppet::Util::Execution).to receive(:execute).with(cli_vec,
                                                                 {:custom_environment => {"FIXVAR" => "fixvalue"},
                                                                  :override_locale => false}).
-                                                                 and_return("value☃".force_encoding(Encoding::ASCII_8BIT))
+                                                                 and_return("value☃".force_encoding(Encoding::US_ASCII))
       expect(check_config_print("value", default_configuration)).to be == "value☃"
     end
   end
@@ -292,8 +292,8 @@ describe "pxp-module-puppet" do
       start_run(default_configuration, default_input)
     end
 
-    it "processes output when puppet's output is ASCII_8BIT (POSIX or C locale) and contains non-ASCII" do
-      output = "everything normal, we're all fine here ☃".force_encoding(Encoding::ASCII_8BIT)
+    it "processes output when puppet's output is US_ASCII (POSIX or C locale) and contains non-ASCII" do
+      output = "everything normal, we're all fine here ☃".force_encoding(Encoding::US_ASCII)
       allow(Puppet::Util::Execution).to receive(:execute).and_return(output)
       allow(output).to receive(:exitstatus).and_return(0)
       allow(output).to receive(:to_s).and_return(output)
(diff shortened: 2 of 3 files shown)


MikaelSmith:
@Iristyle @joshcooper aware of any issues with assuming Puppet output is UTF-8?

modules/pxp-module-puppet
```diff
@@ -70,7 +70,7 @@ end
 
 def force_unicode(s)
   # Puppet still returns UTF-8 when locale is C or POSIX, so force encoding to that.
-  if s.encoding == Encoding::ASCII_8BIT
+  if s.encoding != Encoding::UTF_8
```
Iristyle: So this is a change from checking for binary to UTF-8... seems peculiar to me.

MikaelSmith: The issue it's trying to fix is that SLES 10 and 11 (and a few other platforms) run services (i.e. pxp-agent) with a POSIX locale, which results in encoding coming back as `US_ASCII` (I mistakenly read it as `ASCII_8BIT` before).

However, Puppet's output can still include non-ASCII characters, resulting in an error thrown later when we try to use a regex.

Iristyle:
@MikaelSmith you can't assume it's UTF-8 - on Windows it will be whatever the codepage dictates  (though there may be some low-level ways around that that when using `Puppet::Util::Execution.execute` that I'm still researching - the problem with using that helper is that it writes stdin / stdout / etc to a file but doesn't specify encoding - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/execution.rb#L187-L189). 

I think the only way to guarantee Puppet emits UTF-8 over stdout, in a cross-platform way.. is to set `RUBYOPT` like `RUBYOPT=-EUTF-8` before invoking the Ruby process.  

This is why I was wondering how this initial implementation worked in the first place TBH.

We can chat more about this later today if you'd like - I think it's pretty tricky to get right.

Iristyle:
Some other thoughts I've had about invoking Puppet that could help here.. is if we had a helper to write to log instead of stdout.  Then we could always write the file as UTF-8.

MikaelSmith:
The right fix is to run Puppet in-process and read the report, so I might do that. That's trickier to do right.

I don't think the previous fixes had any effect on Windows.

Iristyle:
Running Puppet in-proc is also a good thing to do.  That's the approach that I used at https://github.com/puppetlabs/puppetlabs-puppet_enterprise/pull/756/files for capturing fact values as well.

MikaelSmith:
Another short-term option might be to just enforce on non-Windows.

MikaelSmith:
Pushed up a fix that excludes Windows. This would unblock CI; pretty soon I'd like to rewrite pxp-module-puppet as started in https://github.com/puppetlabs/pxp-agent/pull/472.

MikaelSmith:
This is a bit of a hack, but I can't think of any other scenarios where it'd cause issues.

MikaelSmith:
@Iristyle any concerns with the current change? I'd like to mostly get CI passing in a targeted way, and make this better overall with #472.

modules/pxp-module-puppet
```diff
@@ -70,7 +70,8 @@ end
 
 def force_unicode(s)
   # Puppet still returns UTF-8 when locale is C or POSIX, so force encoding to that.
-  if s.encoding == Encoding::ASCII_8BIT
+  if !Puppet.features.microsoft_windows? &&
+    (s.encoding == Encoding::US_ASCII || s.encoding == Encoding::ASCII_8BIT)
```
kylog: It looks like this will change the behavior if `s.encoding == Encoding::ASCII_8BIT` on Windows. From the comment thread, I had the impression the intention was not to change Windows but tbh I'm out of my depth wrt Encoding on Windows. Is this change on Windows what we want here?

MikaelSmith: The check for ASCII_8BIT was never released, it was part of the first commit for PCP-545.

Iristyle: `ASCII_8BIT` is the same thing as `BINARY` - you get the actual bytes, with no encoding at all.

MikaelSmith: Trying to force C locale on Windows never prompted this issue, so I prefer to target the fix at non-Windows. Targeting it at US-ASCII and ASCII-8BIT also seems safe.

MikaelSmith: pxp-agent (and thus Puppet) is operating in a POSIX locale as the default on some systems (SLES 10/11, arista, cumulus vx), with default_external encoding as US-ASCII. It prints UTF-8 because that's present in the manifest. Either we consider this a non-issue and change the testing, or we have to force_encoding to UTF-8. I think the only question here is how narrowly we target the change.

`if (s.encoding == Encoding::US_ASCII)` is sufficient to address all the problem platforms we test, and should be perfectly safe (as UTF-8 is a superset of US-ASCII).

kylog:
LGTM. @Iristyle?

modules/spec/unit/modules/puppet_spec.rb
```diff
@@ -292,8 +292,8 @@
       start_run(default_configuration, default_input)
     end
 
-    it "processes output when puppet's output is ASCII_8BIT (POSIX or C locale) and contains non-ASCII" do
-      output = "everything normal, we're all fine here ☃".force_encoding(Encoding::ASCII_8BIT)
+    it "processes output when puppet's output is US_ASCII (POSIX or C locale) and contains non-ASCII" do
+      output = "everything normal, we're all fine here ☃".force_encoding(Encoding::US_ASCII)
```
MikaelSmith: These tests may not be doing anything :/

Iristyle: I was wondering if that was the case

Iristyle:
Just for ref ... re: stdout / stdin ... and how they are captured in Puppet with `Puppet::Util::Execution.execute`:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/execution.rb#L187-L189

modules/pxp-module-puppet
```diff
@@ -32,13 +32,19 @@ env_fix_up = nil
 # use this way of defining the `get_env_fix_up` method so that it can "see"
 # the `env_fix_up` variable
 self.class.send(:define_method, :get_env_fix_up) do
+  # If running in a C or POSIX locale, ask Puppet to use UTF-8
+  base_env = {}
+  if Encoding.default_external == Encoding::US_ASCII
```
MikaelSmith: @Iristyle updated. The theory with this version is that: when Ruby's default encoding is US-ASCII, Puppet may still use other character encodings from the manifest. Request that characters are printed as UTF-8 (to hopefully handle any encodings Puppet may deal with). Later, try to change all execution output to UTF-8 (because that's what we want to compare against with regexes anyway), and if that fails assume we have UTF-8 characters but used a non-UTF-8 locale.

There may be some un-handled corner-cases, but it shouldn't make them worse. If you try to run with a German codepage on Windows, but a manifest has UTF-8 characters that can't be represented in it, you already have a problem.

modules/pxp-module-puppet
```diff
@@ -69,11 +75,13 @@ def last_run_result(exitcode)
 end
 
 def force_unicode(s)
-  # Puppet still returns UTF-8 when locale is C or POSIX, so force encoding to that.
-  if s.encoding == Encoding::ASCII_8BIT
+  begin
+    # Later comparisons assume UTF-8. Convert to that encoding now.
+    s.encoding(Encoding::UTF_8)
```
MikaelSmith: This is probably actually wrong.

MikaelSmith:
This passes on Windows 2012r2, Redhat 7, and all problem platforms (SLES 10/11, arista, cumulus vx).

modules/pxp-module-puppet
Iristyle: I think isolating this to environments where Ruby has set the external encoding to `Encoding::US_ASCII` seems reasonable...

MikaelSmith: I made this the commit message.

Iristyle:
As discussed, we think this improves the situation, doesn't make it worse, and we realize it isn't yet the complete solution - so some tradeoffs have been made for the sake of the timeline.  In the spirit of incremental improvements we're going to get this merged in now and address the other larger looming issues later.

Iristyle:
Looks like we're seeing failures on some platforms (Arista, Cumulus, SLES):

```
stderr: \n/opt/puppetlabs/pxp-agent/modules/pxp-module-puppet:93:in `disabled?': invalid byte sequence in US-ASCII (ArgumentError)\n\tfrom /opt/puppetlabs/pxp-agent/modules/pxp-module-puppet:188:in `start_run'\n\tfrom /opt/puppetlabs/pxp-agent/modules/pxp-module-puppet:271:in `run'\n\tfrom /opt/puppetlabs/pxp-agent/modules/pxp-module-puppet:407:in `action_run'\n\tfrom /opt/puppetlabs/pxp-agent/modules/pxp-module-puppet:416:in `<main>'\n"
```

MikaelSmith:
Hadn't been merged down to master yet. I've done so now.

PR #472, opened 2016-08-12
(PCP-587) pxp-module-puppet rewrite

MikaelSmith:
Rewrites pxp-module-puppet to invoke Puppet in-process. This has several benefits:
- simplified code
- avoids encoding issues from reading console output
- it potentially makes implementing [PCP-285](https://tickets.puppetlabs.com/browse/PCP-285) simpler
- providing more status updates during the Puppet run (via callbacks) is only feasible with Puppet in-process

This was a spike on the idea to see what it would take. Needs a ticket sometime.


MikaelSmith:
Specs will need a major update. I don't think I want to do that until I have a final plan for this.

modules/pxp-module-puppet
```diff
@@ -111,89 +51,49 @@ def make_error_result(exitcode, error_type, error_message)
...
+  start_time = nil
+  report = nil
+  begin
+    cmd = Puppet::Util::CommandLine.new('agent', action_input['flags'])
+    Puppet.initialize_settings(cmd.args)
+    agent = Puppet::Agent.new(Puppet::Configurer, false)
+    transaction_uuid = SecureRandom.uuid
+    report = Puppet::Transaction::Report.new("apply", nil, Puppet[:environment], transaction_uuid)
+    exitcode = agent.run(:report => report, :transaction_uuid => transaction_uuid)
+  rescue => e
```
er0ck: i know this was here before, but we really shouldn't be rescuing any exception.  this does all sorts of nasty things like block `Interrupt` and `SignalException` and can result in false passes as it can block beaker's exception handling.

MikaelSmith: I'll look into it more. There are some exceptions specifically exempted from the blanket `rescue` statement. We want to catch anything that might prevent us from forming a useful error response.

modules/pxp-module-puppet
```diff
@@ -120,59 +51,26 @@ def make_error_result(exitcode, error_type, error_message)
...
-  data = YAML.parse_file(filename)
-  data.root.each do |o|
-    o.tag = nil if o.respond_to?(:tag=)
-  end
-
-  data.to_ruby
-end
-
 def nest_metrics(metrics)
   metrics.fetch('resources', {}).fetch('values', {}).inject({}) do |result, (name,human_name,value)|
```
MikaelSmith: This doesn't work with the actual Metrics object.

modules/Gemfile
```diff
@@ -1,6 +1,16 @@
 source 'https://rubygems.org'
 
-gem 'puppet', '~> 4.2'
+def vanagon_location_for(place)
```
nicklewis: Why `vanagon`? Seems like this was copy-pasted.

modules/pxp-module-puppet
```diff
@@ -187,53 +87,37 @@ def wait_for_lockfile(lockfile = '', check_interval = 0.1, give_up_after = 30)
...
-  cmd = make_command_array(config, action_input)
-  env = make_environment_hash()
-
-  last_run_report = check_config_print("lastrunreport", config)
-  if last_run_report.empty?
-    return make_error_result(exitcode, Errors::NoLastRunReport,
-                             "could not find the location of the last run report")
-  end
+  cmd = Puppet::Util::CommandLine.new('agent', action_input['flags'])
+  Puppet.initialize_settings(cmd.args)
```
nicklewis: `--debug` and `--verbose` won't work with this approach, because they're set by the agent application, not by the settings system.

nicklewis: This code also doesn't handle run mode, which means any settings set in puppet.conf under the `[agent]` section wouldn't be honored.

nicklewis: This also doesn't configure terminuses, which means that no report will be submitted to the master, and that the cached catalog won't be updated.

nicklewis: Actually it also seems that no catalog will be retrieved, because the default catalog terminus is "compiler".

PR #473, opened 2016-08-18
(PCP-558) Give UTF8 test case its own acceptance test

james-stocks:
PCP-494 added a UTF8 character into an existing pxp-module-puppet test case to ensure we support them (commit 85e36fb).
This is currently failing on some platforms in CI (see PCP-558).
This PR splits the test with a UTF8 character into its own test, so it gets its own result in our result matrices. It also adds a specific assertion so that if PCP-494 does regress, we get an explicit assertion failure for it.

[skip ci]


Iristyle:
Could you mention the commit that introduced the original test (the one from PUP-494)

james-stocks:
@Iristyle added the original commit ref 85e36fb to both commit message and the PR summary

PR #474, opened 2016-08-18
(PCP-549) Wait for other Puppet runs to finish

MikaelSmith:
When `pxp-module-puppet run` is triggered and Puppet was already
running, the module would previously identify the failure and return an
error message saying Puppet could not be run because another run was in
progress. This leads to a potentially slow feedback cycle of retrying
Puppet until it can run, when all the module user wants is to ensure a
Puppet run happens with the arguments they specified.

Change pxp-module-puppet to wait until the prior Puppet run completes
and attempt to run again until it succeeds or encounters a different
error.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -394,3 +394,112 @@ def have_all_rpc_responses?(targets, responses)
   end
   return true
 end
+
+# A suite of functions for creating a manifest that sleeps for a long period
+# and interacting with Puppet runs of that manifest.
+def create_sleep_manifest(master, manifest, seconds_to_sleep)
+    create_remote_file(master, manifest, <<-MODULEPP)
+node default {
+  case $::osfamily {
+    'windows': { exec { 'sleep': command => 'C:/Windows/system32/ping.exe 127.0.0.1 -n #{seconds_to_sleep}', returns => 1 } }
+    'Darwin':  { exec { 'sleep': command => '/bin/sleep #{seconds_to_sleep} || /usr/bin/true', } }
+    default:   { exec { 'sleep': command => '/bin/sleep #{seconds_to_sleep} || /bin/true', } }
+  }
+}
+MODULEPP
+    on(master, "chmod 644 #{manifest}")
+end
+
+def wait_for_sleep_process(target)
+  begin
+    ps_cmd = target['platform'] =~ /win/ ? 'ps -efW' : 'ps -ef'
+    sleep_process = target['platform'] =~ /win/ ? 'PING' : '/bin/sleep'
+    retry_on(target, "#{ps_cmd} | grep '#{sleep_process}' | grep -v 'grep'", {:max_retries => 15, :retry_interval => 2})
+  rescue
+    fail("After triggering a puppet run on #{target} the expected sleep process did not appear in process list")
+  end
+end
+
+def stop_sleep_process(targets)
+  targets = [targets].flatten
+
+  targets.each do |target|
+    case target['platform']
+    when /osx/
+      command = "ps -e -o pid,comm | grep sleep | sed 's/^[^0-9]*//g' | cut -d\\  -f1"
+    when /win/
+      command = "ps -efW | grep PING | sed 's/^[^0-9]*[0-9]*[^0-9]*//g' | cut -d ' ' -f1"
+    else
+      command = "ps -ef | grep 'bin/sleep ' | grep -v 'grep' | grep -v 'true' | sed 's/^[^0-9]*//g' | cut -d\\  -f1"
+    end
+
+    # A failed test may leave an orphaned sleep process, handle multiple matches.
+    pids = nil
+    on(target, command) do |output|
+      pids = output.stdout.chomp.split
+      if pids.empty?
+        fail('Did not find a pid for the sleep process holding up Puppet - cannot test PXP response if Puppet isn\'t sleeping')
+      end
+    end
+
+    pids.each do |pid|
+      target['platform'] =~ /win/ ?
+        on(target, "taskkill /F /pid #{pid}") :
+        on(target, "kill -s ALRM #{pid}")
+    end
+  end
+end
+
+# Initiates an rpc_non_blocking_request, verifies that it receives a provisional response for each target,
+# and returns the transaction ids.
+def start_puppet_non_blocking_request(broker, targets, environment = 'production')
+  transaction_ids = []
+
+  provisional_responses = nil
+  begin
+    provisional_responses = rpc_non_blocking_request(broker, targets,
+                                                     'pxp-module-puppet', 'run',
+                                                     {:flags => ['--onetime',
+                                                                 '--no-daemonize',
+                                                                 '--environment', environment]})
+  rescue => exception
+    fail("Exception occured when trying to send rpc_non_blocking_request to all agents: #{exception}")
+  end
+
+  targets.each do |identity|
+    assert_equal("http://puppetlabs.com/rpc_provisional_response",
+                 provisional_responses[identity][:envelope][:message_type],
+                 "Did not receive expected rpc_provisional_response in reply to non-blocking request")
+    transaction_ids << provisional_responses[identity][:data]["transaction_id"]
+  end
+  transaction_ids
+end
+
+def check_puppet_non_blocking_response(identity, transaction_id, max_retries, query_interval,
+                                       expected_result, expected_environment = 'production')
+  puppet_run_result = nil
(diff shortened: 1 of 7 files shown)


MikaelSmith:
This should have an acceptance test.

This could get stuck in a couple ways:
- a Puppet run could crash or be killed and leave the lockfile behind. This suggests we should possibly have a longer retry period where it tries a Puppet run anyway (which has logic for cleaning up orphaned lockfiles).
- Puppet hangs, in which case this request would also hang. Should it ever stop? The initial response was no; if it's hanging manual intervention is required anyway.

MikaelSmith:
Updated to handle orphaned lockfiles by retrying Puppet every 30 seconds.

MikaelSmith:
@nicklewis @adreyer any more thoughts on this behavior?

nicklewis:
Is there an advantage to checking the lock file before trying to run for the first time? My thinking is that if the lock file is orphaned, we'll wait 30 seconds unnecessarily before even attempting to run (which would succeed).

MikaelSmith:
The first check for the lockfile is a no-op, because it checks for an empty string (always returns false). So it'll check `File.exist?('')` (which returns false), run Puppet, detect that Puppet is already running, then look for the lockfile and wait for it to go away (or 30 seconds to pass, and retry Puppet).

MikaelSmith:
The acceptance tests still need work on Windows, pulling them out for now.

MikaelSmith:
Ok, they work now. Acceptance tests probably need more work to make them reliable.

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -0,0 +1,95 @@
...
+  environment_name = mk_tmp_environment(env_name)
+
+  step 'On master, create a new environment that will result in a slow run' do
+    site_manifest = "#{environmentpath}/#{environment_name}/manifests/site.pp"
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  case $::osfamily {
+    'windows': { exec { 'sleep':
+                        command => 'true',
+                        unless  => 'sleep 10', #PUP-5806
```
james-stocks: PUP-5806 is now closed, it advises that the Windows command should be:
    command => 'cmd.exe /c sleep 5 || true'
(That might have the undesirable behaviour of not having the sleep process appear in cygwin's ps though)

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -0,0 +1,95 @@
...
+                        path    => 'C:\\cygwin64\\bin',} }
```
james-stocks: This looks like it will fail on 32bit Windows (and so will fail in the puppet-agent pipelines).
Note that the existing acceptance test that has this exec resource does also have this problem already, but that test is currently skipped on all Windows variants.

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+  end
+
+  step 'Send two requests to all agents, the first with a long-running task' do
+    target_identities = []
+    agents.each do |agent|
+      target_identities << "pcp://#{agent}/agent"
+    end
+
+    provisional_responses = rpc_non_blocking_request(master, target_identities,
```
james-stocks: This should maybe have a begin/rescue block so that any error here results in a clean beaker failure instead of an unhandled error.
Or, if a begin/rescue for every rpc call in every acceptance test seems excessive, consider removing the one at lines 51-58 - someone reading this code might wonder why the blocking request gets a rescue but the non-blocking doesn't

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -0,0 +1,95 @@
...
+        # The first runs should be done, because the 2nd run shouldn't have even started until
+        # the first finished.
+        assert_equal("http://puppetlabs.com/rpc_provisional_response",
+                     provisional_responses[identity][:envelope][:message_type],
+                     "Did not receive expected rpc_provisional_response in reply to non-blocking request")
+        transaction_id = provisional_responses[identity][:data]["transaction_id"]
+
+        query_responses = rpc_blocking_request(master, [identity],
+                                               'status', 'query', {:transaction_id => transaction_id})
+        action_result = query_responses[identity][:data]["results"]
```
james-stocks: PCP-494 showed that in the event of a problem with the PXP run, there may be no "results" and it may therefore be valuable to have one more assertion on .has_key('results') that includes the contents of [identity][:data] in the assertion message - that gives visibility of what the response  contents actually was instead of just erroring out on a cryptic NilClass exception

acceptance/tests/pxp-module-puppet/run_puppet_killed_puppet.rb
```diff
@@ -0,0 +1,132 @@
...
+      on agent, puppet('agent', '--test', '--environment', environment_name, '--server', "#{master}",
+                       '>/dev/null & echo $!')
+    end
+  end
+
+  # TODO: switch to checking all agents after each sleep
+  agents.zip(lockfiles).each do |agent, lockfile|
+    step "Check for lockfile on #{agent}" do
+      lockfile_exists = false
+      for i in 1..50
```
james-stocks: You could maybe use [retry_on](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker%2FDSL%2FHelpers%2FHostHelpers%3Aretry_on) here to save writing a loop.
If using retry_on , it needs to a begin/rescue to translate anything it throws into a meaningful assertion failure.

MikaelSmith: I'm not sure how to do that with the action: `agent.file_exist?`.

MikaelSmith:
I've pulled out acceptance tests to simplify this review. I can add them back if it hasn't been merged before I finish cleaning them up, but this can be merged as-is.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -394,3 +394,112 @@ def have_all_rpc_responses?(targets, responses)
...
+node default {
+  case $::osfamily {
+    'windows': { exec { 'sleep': command => 'C:/Windows/system32/ping.exe 127.0.0.1 -n #{seconds_to_sleep}', returns => 1 } }
+    'Darwin':  { exec { 'sleep': command => '/bin/sleep #{seconds_to_sleep} || /usr/bin/true', } }
+    default:   { exec { 'sleep': command => '/bin/sleep #{seconds_to_sleep} || /bin/true', } }
+  }
+}
+MODULEPP
+    on(master, "chmod 644 #{manifest}")
+end
```
er0ck: in the puppet repo we have a `#create_sitepp`  we might want to use that here later, but it shouldn't hold up the merge.

acceptance/tests/pxp-module-puppet/run_puppet_during_puppet.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  target_identities = []
+  agents.each do |agent|
+    target_identities << "pcp://#{agent}/agent"
+  end
+
+  transaction_ids = []
+  step 'Run Puppet on agents' do
```
er0ck: might want to document this is running puppet via PCP, but i'm okay as is.

er0ck:
acceptance tests look good to me. nice work!
👍

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -394,3 +394,112 @@ def have_all_rpc_responses?(targets, responses)
...
+  targets = [targets].flatten
+
+  targets.each do |target|
+    case target['platform']
+    when /osx/
+      command = "ps -e -o pid,comm | grep sleep | sed 's/^[^0-9]*//g' | cut -d\\  -f1"
+    when /win/
+      command = "ps -efW | grep PING | sed 's/^[^0-9]*[0-9]*[^0-9]*//g' | cut -d ' ' -f1"
+    else
+      command = "ps -ef | grep 'bin/sleep ' | grep -v 'grep' | grep -v 'true' | sed 's/^[^0-9]*//g' | cut -d\\  -f1"
```
mruzicka: What is the ```grep -v 'true'``` excluding?

MikaelSmith: I'm not really sure, I carried this over from what `restart_pxp_agent_during_non_blocking_run` was doing before.

mruzicka:
LGTM

PR #476, opened 2016-08-23
(maint) Fix Travis script to ensure pot file is regenerated

MikaelSmith:
The previous change referenced the wrong file. Fix it so it deletes the
correct gettext pot file, ensuring CI will test regenerating it.


parisiale:
👍

PR #479, opened 2016-08-25
(PCP-562) Refactor dependency on cpp-pcp-client

MikaelSmith:


MikaelSmith:
Depends on https://github.com/puppetlabs/cpp-pcp-client/pull/201. Also passes an acceptance run on EL 7.

PR #482, opened 2016-09-01
(PCP-572) Update how acceptance terminates sleep processes

james-stocks:
The method of ending a running sleep process using SIGALRM does not work on Solaris 10.
We don't need SIGALRM specifically, we just need 'sleep' to terminate; so this commit sends SIGTERM instead of SIGALRM.

Because the Solaris failure caused knock-on failures in other tests; this commit also adds a teardown to the tests that make use of sleep/kill.

[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -421,7 +421,7 @@ def wait_for_sleep_process(target)
   end
 end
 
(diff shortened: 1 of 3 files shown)


james-stocks:
Tested the acceptance/tests/pxp-module-puppet/ tests with Solaris 10 64bit, SLES 11, Windows 2012R2 64bit, Cisco NX (Cisco NX had one unrelated failure)

PR #483, opened 2016-09-02
(PCP-577) Don't run tests that restart broker

MikaelSmith:
Don't restart the broker when running tests to restart pxp-agent. This
prevents restarting the agent host when that host also includes the
master role, so that we only restart agents that don't include the
broker.


mruzicka:
LGTM

PR #484, opened 2016-09-02
(PCP-576) Ensure cisconx tests can read lock file

MikaelSmith:
CiscoNX tests don't run as root. Ensure we read the lockfile using
Beaker methods to pickup any helpers (like running sudo).

[skip ci]


mruzicka:
LGTM

PR #487, opened 2016-09-14
(PCP-286) Ensure service is running in service test teardown

james-stocks:
PCP-286 had occasional failures in acceptance CI on Windows.
They were caused by tests/service_stop_start.rb leaving the service in the Paused state.
This commit updates the teardown block of tests/service_stop_start.rb to make sure the pxp-agent is in a working running state.
PUP-6702 means that the teardown needs to directly use the 'net' command on Windows rather than puppet managing the service state.

[skip-ci]

### DIFF
--- a/acceptance/tests/service_stop_start.rb
+++ b/acceptance/tests/service_stop_start.rb
@@ -7,8 +7,22 @@ test_name 'Service Start stop/start, with configuration)'
 # On teardown, restore configuration file on each agent
 teardown do
   agents.each do |agent|
+    # Create a valid config file
+    create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+
+    # Ensure pxp-agent service is running
+    # PUP-6702 - on Windows, the service may be in the Paused state
+    # 'puppet resource service pxp-agent' will not work as expected
+    # Workaround: stop/start it using net
+    if(windows?(agent)) then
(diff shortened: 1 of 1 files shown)


acceptance/tests/service_stop_start.rb
```diff
@@ -7,8 +7,21 @@
...
   agents.each do |agent|
+    # Create a valid config file
+    create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+
+    # Ensure pxp-agent service is running
+    # PUP-6702 - on Windows, the service may be in the Paused state
+    # 'puppet resource service pxp-agent' will not work as expected
+    # Workaround: stop/start it using net
+    if(windows?(agent)) then
+      on(agent, 'net stop pxp-agent && net start pxp-agent')
```
glennsarti: Note that if the service is already in a stopped state , net stop will throw an error and will _not_ run the net start command.  It may make more sense to run `net stop` as one command and ignore errors and then run `net start` e.g.

```
on(agent, 'net stop ...', :fail_on_fail => false)
on(agent, 'net start ...')
```

MosesMendoza: 👍  to that comment

james-stocks:
Thanks @glennsarti - that might have been a real problem if the test had failed and entered teardown midway. I've squashed that change into the original commit

PR #491, opened 2016-10-04
(PCP-605) Disable restart_host_run_puppet on Fedora

MosesMendoza:
Due to a bug (tracked by QENG-4371) the fedora hosts in our VM Pooler take
anywhere from 5 to 20 minutes to reboot, causing this test to time out and fail
consistently in Puppet Agent testing. Until this is resolved, the test running
on Fedora is causing consistent triage effort in Agent with little value. This
commit follows the existing pattern for excluding agents to exclude the Fedora
platform from testing. The long reboot times have been seen across fedora
arches and versions so this generically excludes the platform rather than a
specific version/arch target.

Signed-off-by: Moses Mendoza <moses@puppet.com>

PR #492, opened 2016-10-04
(PCP-605) (PCP-585) Disable restart_host_run_puppet on Fedora

MosesMendoza:
Due to a bug (tracked by QENG-4371) the fedora hosts in our VM Pooler take
anywhere from 5 to 20 minutes to reboot, causing this test to time out and fail
consistently in Puppet Agent testing. Until this is resolved, the test running
on Fedora is causing consistent triage effort in Agent with little value. This
commit follows the existing pattern for excluding agents to exclude the Fedora
platform from testing. The long reboot times have been seen across fedora
arches and versions so this generically excludes the platform rather than a
specific version/arch target. Secondarily, we are also seeing sporadic issues
on occassion reconnecting to the broker after reboot (PCP-585), which is also
impetus for this change.

Signed-off-by: Moses Mendoza <moses@puppet.com>


MosesMendoza:
This PR supersedes https://github.com/puppetlabs/pxp-agent/pull/491

PR #495, opened 2016-10-12
(BKR-960) Skip timesync for ciscoxr-64a

james-stocks:
BKR-960 is a beaker bug where the timesync helper method doesn't support cisco xr ios.
As a workaround, this commit just skips timesync for that platform.
This leaves the tests at risk of failing due to clock differences, but allows the tests to complete otherwise.

[skip ci]


MikaelSmith:
Is that fixable in BKR?

Magisus:
👍

james-stocks:
@MikaelSmith yes, the ticket will remain open to fix it in beaker. This commit is to allow us to get further with testing on Cisco XR. I'll comment on BKR-960 that part of the acceptance criteria must be to revert this workaround

PR #496, opened 2016-10-13
(maint) Use AppVeyor OpenSSL

MikaelSmith:
Stop fighting against AppVeyor including OpenSSL and use the
pre-installed version. Also restrict PATH to only use known paths. This
fixes builds that accidentally loaded a different OpenSSL from
C:\Windows\system32.


MikaelSmith:
ping @mruzicka @parisiale

PR #498, opened 2016-10-18
(PCP-611) Ensure run_puppet_twice.rb fails when appropriate

james-stocks:
run_puppet_twice.rb continues to pass even when the pxp-module-puppet script
does not actually handle re-running Puppet (i.e. if a stable packages is accidentally used).
This is due to a race where the test case end the 1st puppet run
before the 2nd puppet run has had time to launch puppet and detect that puppet is
already running.

This commit ensures that the 2nd puppet run has exited before continuing and
ending the 1st puppet run.

[skip ci]

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
@@ -53,6 +53,62 @@ test_name 'Run Puppet while a Puppet Agent run is in-progress, wait for completi
     transaction_ids_2 = start_puppet_non_blocking_request(master, target_identities)
   end
 
+  # Wait for the 2nd puppet agent process to exit
+  #
+  # Passing case: 2nd puppet process starts, exits on puppet_already_running error, pxp-module-puppet waits to retry.
+  # Failing case: 2nd puppet process starts, exits on puppet_already_running error, pxp-module-puppet gives up.
+  #
+  # If the test continued without waiting for the 2nd instance to exit, the 1st agent instance might exit before the
+  # 2nd agent instance was fully started; causing a false positive test result.
+  # 
+  # Because beaker's sampling of the puppet agent PIDs might begin before the 2nd puppet agent appears, OR
+  # it might begin after the 2nd puppet agent has already shut down; this step needs to use a heuristic approach.
+  # It succeeds if either:
+  #   a. It observes 2 PIDs on one sample, then only 1 PID on the subsequent sample
+  #   b. It observes 1 PID for at least 10 concurrent samples
+  #
+  step 'Wait for only one puppet agent PID to exist' do
+    agents.each do |agent|
+
+      satisfied = false
+      pid_counting_attempts = 0
+      MAX_PID_COUNTING_ATTEMPTS = 30
+      concurrent_samples_of_1_pid = 0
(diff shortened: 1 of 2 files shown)


james-stocks:
Tested both the passing case (with a master package) and failing case (with a stable package) on redhat7-64a, windows2008r2-64a, windows 10 and osx1011-64a

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -53,6 +53,58 @@
...
+    agents.each do |agent|
+
+      satisfied = false
+      pid_counting_attempts = 0
+      MAX_PID_COUNTING_ATTEMPTS = 30
+      concurrent_samples_of_1_pid = 0
+      REQUIRED_SAMPLES_OF_1_PID = 10
+      two_pids_observed = false
+
+      while pid_counting_attempts < MAX_PID_COUNTING_ATTEMPTS do
```
lucywyman: Should you add '&& satisfied == false' so that it doesn't keep polling once one of the conditions has been met?

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -53,6 +53,58 @@
...
+            satisfied = true
+          end
+          if !satisfied && concurrent_samples_of_1_pid >= REQUIRED_SAMPLES_OF_1_PID then
+            logger.debug "Assuming 2nd agent has stopped due to #{concurrent_samples_of_1_pid} concurrent "\
+                         "pid checks that only returned 1 running agent pid"
+            satisfied = true
+          end
+        else
+          concurrent_samples_of_1_pid = 0
+          two_pids_observed = true if puppet_agent_pids.length == 2
```
james-stocks: I was considering that as I wrote the code. If there are 0, or 3, or more than 3 processes, then something is not right. 
Thinking about it, the test should error if there is an unexpected number of pids. The whole point to this code is that we control what happens and avoid a false positive test result. If we lose control of pids then we lose control of the test - I'll add a check for 1/3/3+ pids and raise an exception if that ever happens

acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
```diff
@@ -53,6 +53,62 @@
...
+      pid_counting_attempts = 0
+      MAX_PID_COUNTING_ATTEMPTS = 30
+      concurrent_samples_of_1_pid = 0
+      REQUIRED_SAMPLES_OF_1_PID = 10
+      two_pids_observed = false
+
+      while pid_counting_attempts < MAX_PID_COUNTING_ATTEMPTS do
+        puppet_agent_pids = get_puppet_agent_pids(agent)
+        if puppet_agent_pids.length == 1 then
+          concurrent_samples_of_1_pid += 1
```
james-stocks: If there was ever 0 PIDs then the test would error out at line 98.

This counter should increment on 1 PID, reset to 0 on 2 PIDs, and the test will error on any other number of PIDs.

lucywyman: +1

lucywyman:
:+1:  from me!

PR #504, opened 2016-10-25
(PCP-625) Skip beaker time sync on Windows Sever 2016

Magisus:
When testing a 32-bit puppet-agent MSI on Windows Server 2016, forcing a
time sync through beaker was causing the host to temporarily have a time
far in the past. This was preventing retrieval of the signed SSL
certificate from the master. However, the host appears to sync correctly
on its own, so this commit causes beaker to skip forcing a sync.

### DIFF
--- a/acceptance/setup/common/005_SyncTime.rb
+++ b/acceptance/setup/common/005_SyncTime.rb
@@ -8,6 +8,10 @@ test_name 'Ensure hosts have synchronized clocks'
 # Affects OSX and Ubuntu on vmpooler
 applicable_hosts = hosts.select{|host| host['platform'] !~ /osx|ubuntu/}
 
+# PCP-625 the beaker timesync method causes strange behavior on Windows Server 2016,
+# temporarily setting the date far into the past, which breaks SSL certificate retrieval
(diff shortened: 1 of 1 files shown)


Magisus:
This has been manually validated with a local beaker run. The run proceeds past the cert retrieval steps, and all tests pass.

acceptance/setup/common/005_SyncTime.rb
```diff
@@ -8,6 +8,10 @@
 # Affects OSX and Ubuntu on vmpooler
 applicable_hosts = hosts.select{|host| host['platform'] !~ /osx|ubuntu/}
 
+# PCP-625 the beaker timesync method causes strange behavior on Windows Server 2016,
+# temporarily setting the date far into the past, which breaks SSL certificate retrieval
+applicable_hosts = hosts.select{|host| host['platform'] !~ /windows-2016/}
```
MikaelSmith: Oh, this breaks skipping on ubuntu and osx. You need to modify `applicable_hosts`.

Magisus: Updated. Nice catch!

PR #509, opened 2016-10-31
(PCP-631) Prevent hanging on Windows ssh after shutdown

MikaelSmith:
The beaker restart helper sends shutdown to Windows, but then hangs for
a long timeout as ssh doesn't close cleanly. This is filed as BKR-971.

As a work-around, add a 5s timeout after sending shutdown to allow it be
begin, then close the SSH connection and start attempting to reconnect.

### DIFF
--- a/acceptance/tests/restart_host_run_puppet.rb
+++ b/acceptance/tests/restart_host_run_puppet.rb
@@ -48,7 +48,23 @@ test_name 'C94777 - Ensure pxp-agent functions after agent host restart' do
 
   step "restart each agent" do
     applicable_agents.each do |agent|
-      agent.reboot
+      # Work-around reboot hanging on reply (BKR-971)
+      rebooter = Thread.start { agent.reboot }
+      sleep 5
+
+      # For the truly paranoid, we handle exceptions for closing the SSH connection and rejoining thread separately.
+      begin
+        agent.close
+      rescue Exception => e
+        logger.debug "Exception closing agent connection after reboot: #{e}"
+      end
+
+      begin
+        rebooter.join
+      rescue Exception => e
+        logger.debug "Exception waiting for ssh close after reboot: #{e}"
+      end
+
       # BKR-812
       timeout = 30
       begin


MikaelSmith:
@joshcooper I think this could be a work-around for long Windows runs.

acceptance/tests/restart_host_run_puppet.rb
```diff
@@ -48,7 +48,14 @@
 
   step "restart each agent" do
     applicable_agents.each do |agent|
-      agent.reboot
+      # Work-around reboot hanging on reply (BKR-971)
+      begin
+        Timeout.timeout(5) do
+          agent.reboot
+        end
```
joshcooper: This may not work as expected... If agent.reboot calls into a native syscall, then ruby won't be able to interrupt the thread. It might be better to just sleep for 5 seconds and then call close explicitly.

http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/
http://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/

MikaelSmith: Sleep doesn't work, because it's the `agent.reboot` itself that's hanging as it waits for a response to its shutdown command.

joshcooper: Ah, in that case, I'd probably execute the `reboot` call in a separate thread.

Also it looks like beaker tried to workaround the issue in https://github.com/puppetlabs/beaker/commit/f96a15aa3d0f701618bf1e63f78a40f2329c4298#diff-3556c18d916e298ae0708391158c86d0R6, but I don't think adding the `sleep(30)` is going to help anything, since beaker is blocked waiting for the response to the command.

MikaelSmith: Something like
```
      rebooter = Thread.start { agent.reboot }
      sleep 5
      agent.close
      rebooter.join
```

MikaelSmith: Might need a rescue in there somewhere, just ran across an exception from closing the connection during the reboot call.

MikaelSmith:
Updated.

acceptance/tests/restart_host_run_puppet.rb
```diff
@@ -48,7 +48,16 @@
 
   step "restart each agent" do
     applicable_agents.each do |agent|
-      agent.reboot
+      # Work-around reboot hanging on reply (BKR-971)
+      rebooter = Thread.start { agent.reboot }
+      sleep 5
+      begin
+        agent.close
+        rebooter.join
```
joshcooper: If `close` raises, we don't `join`. Should we move `join` after the begin/rescue?

MikaelSmith: I wasn't sure which was raising. I could rescue each separately; I suspect it's the join that would propagate an exception from the other thread.

MikaelSmith:
Updated.

parisiale:
👍 
Pinging @joshcooper

joshcooper:
Looks good, only takes 79 seconds.

```
  * restart each agent

    tlsw7lxrvhlcs1a.delivery.puppetlabs.net (windows2012r2-64-1) 11:16:50$ shutdown /f /r /t 0 /d p:4:1 /c "Beaker::Host reboot command issued"

    tlsw7lxrvhlcs1a.delivery.puppetlabs.net (windows2012r2-64-1) executed in 0.11 seconds
    Waiting for connection failure on windows2012r2-64-1 (attempt 1, try again in 3 second(s))

    windows2012r2-64-1 11:16:50$ echo echo
    sleep 3 second(s):     .    .    .
    Waiting for connection failure on windows2012r2-64-1 (attempt 2, try again in 5 second(s))

    windows2012r2-64-1 11:16:53$ echo echo
    sleep 5 second(s):     .    .    .    .    .
    Waiting for connection failure on windows2012r2-64-1 (attempt 3, try again in 8 second(s))

    windows2012r2-64-1 11:17:03$ echo echo
    sleep 8 second(s):     .    .    .    .    .    Warning: Attemped ssh.close, (caught Errno::ETIMEDOUT - Operation timed out - recvfrom(2)).
    ssh connection to windows2012r2-64-1 has been terminated
    .    .    .
    Waiting for connection failure on windows2012r2-64-1 (attempt 4, try again in 13 second(s))

    windows2012r2-64-1 11:17:32$ echo echo
    Exception waiting for ssh close after reboot: undefined method `open_channel' for nil:NilClass

    * wait until pxp-agent is back up and associated on tlsw7lxrvhlcs1a.delivery.puppetlabs.net

      tlsw7lxrvhlcs1a.delivery.puppetlabs.net (windows2012r2-64-1) 11:17:32$ #<Beaker::Command:0x007fc3f9d60488>
        Trying command 30 times.
      .        Attempting ssh connection to tlsw7lxrvhlcs1a.delivery.puppetlabs.net, user: Administrator, opts: {:config=>false, :paranoid=>false, :auth_methods=>["publickey"], :port=>22, :forward_agent=>true, :keys=>["~/.ssh/id_rsa-acceptance"], :user_known_hosts_file=>"/Users/josh/.ssh/known_hosts", :keepalive=>true, :user=>"Administrator"}

      tlsw7lxrvhlcs1a.delivery.puppetlabs.net (windows2012r2-64-1) 11:17:34$ #<Beaker::Command:0x007fc3f9d60488> ostensibly successful.

  * Send an rpc_blocking_request to all agents after reboot
{:envelope=>{:id=>"422430ca-eb88-44c2-8b65-e4b920eca90b", :targets=>["pcp://chara.corp.puppetlabs.net/ruby-pcp-client-77992"], :message_type=>"http://puppetlabs.com/rpc_blocking_response", :sender=>"pcp://tlsw7lxrvhlcs1a.delivery.puppetlabs.net/agent", :expires=>"2016-11-04T18:17:51.718334Z"}, :data=>{"transaction_id"=>"bcd7e903-df4f-4699-bd9e-677a26423810", "results"=>{"kind"=>"apply", "time"=>"2016-11-04 18:17:44 +0000", "transaction_uuid"=>"7bb7510e-72ac-4d66-a3a0-39607b7ee9df", "environment"=>"production", "status"=>"unchanged", "exitcode"=>0, "version"=>1}}}
    * Check Run Puppet response for tlsw7lxrvhlcs1a.delivery.puppetlabs.net
Begin teardown
End teardown
tests/restart_host_run_puppet.rb passed in 79.14 seconds
```

On one hand, the problem is only triggered if the SUT gets a new IP after reboot, so this may not prove anything. But on the other hand, we explicitly close the connection, resulting in an exception in `recvfrom` and terminating the ssh connection fairly early on. Without this fix it was taking beaker 25+ mins to detect the ssh connection was bad and to terminate the connection itself:

```
09:55:46     q1f5xpbbrulnivq.delivery.puppetlabs.net (windows2012r2-64-1) 16:55:46$ shutdown /f /r /t 0 /d p:4:1 /c "Beaker::Host reboot command issued"
09:55:46     
...
10:10:37     windows2012r2-64-1 17:10:37$ echo echo
10:10:37     sleep 55 second(s):     .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .    .
10:35:22     Waiting for connection failure on windows2012r2-64-1 (attempt 8, try again in 89 second(s))
10:35:22     
10:35:22     windows2012r2-64-1 17:35:22$ echo echo
10:35:22     Connection on windows2012r2-64-1 failed as expected (Errno::ETIMEDOUT - Connection timed out - recvfrom(2))
10:35:22     Warning: Attemped ssh.close, (caught Net::SSH::Disconnect - connection closed by remote host).
10:35:22     ssh connection to windows2012r2-64-1 has been terminated
```

So I'm 👍 , but can we target LTS-1.7 instead (since the acceptance test exists there) and merge up to stable and master?

MikaelSmith:
Retargeted.

joshcooper:
I merged LTS-1.7 up to stable and master

PR #510, opened 2016-11-01
(PCP-632) Fix run_puppet_twice.rb on OSX

james-stocks:
This test fails in CI when it requires either 1 or 2 puppet agent processes to be running but discovers 3.
1 instance is from puppet running as a service (despite the pre-suite stopping this).
Another instance is from puppet agent being used for another purpose (puppet agent --configprint).

This commit updates the test to not fail immediately on finding 3 pids and also adds a workaround in shutting down puppet at the start of the test.

[skip ci]

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_twice.rb
@@ -32,6 +32,10 @@ test_name 'Run Puppet while a Puppet Agent run is in-progress, wait for completi
     end
   end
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
This will probably work. pxp-module-puppet also invokes `puppet agent --configprint lastrunreport` before running `puppet agent --...`. We could try to filter out `--configprint` from the grep, but at worst it's an occasional false positive (the test could pass when it should fail if the code changed).

This test will need rework when we rewrite pxp-module-puppet to run Puppet directly anyway (PCP-587) - we'll want to look for invocations of that module - so I'm fine with it as-is.

PR #515, opened 2016-11-10
(PCP-625) Skip beaker time sync on Windows Sever 2016

Magisus:
This is a cherry-pick from master to the LTS branch.

When testing a 32-bit puppet-agent MSI on Windows Server 2016, forcing a
time sync through beaker was causing the host to temporarily have a time
far in the past. This was preventing retrieval of the signed SSL
certificate from the master. However, the host appears to sync correctly
on its own, so this commit causes beaker to skip forcing a sync.

[skip ci]


ScottGarman:
:+1:

PR #516, opened 2016-11-11
(PCP-657) Allow unknown status in acceptance tests

MikaelSmith:
Handles pxp-agent returning unknown status. This is currently considered
normal operation when an action has been queued but has not yet invoked
a process.


er0ck:
lgtm

PR #523, opened 2016-12-07
(PCP-592) Add acceptance test for syntax significant characters

james-stocks:
Add an additional acceptance test that ensures that syntax significant characters are properly handled by PXP

[skip ci]

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_syntax_chars.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_syntax_chars.rb
@@ -1,0 +1,61 @@
+require 'pxp-agent/test_helper.rb'
+require 'puppet/acceptance/environment_utils'
+
+test_name 'C98237 - Run puppet with syntax significant characters in Puppet code' do
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  env_name = test_file_name = File.basename(__FILE__, '.*')
+  environment_name = mk_tmp_environment(env_name)
+
+  step 'On master, create a new environment that includes syntax significant characters in resources and will result in Changed' do
+    site_manifest = "#{environmentpath}/#{environment_name}/manifests/site.pp"
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  notify {'\\'\\"&+/\\()?,.#!@$_-~':}
+}
+SITEPP
+    on(master, "chmod 644 #{site_manifest}")
+  end
+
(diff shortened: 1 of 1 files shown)


james-stocks:
@MikaelSmith @lucywyman @zreichert As well as reviewing the code added here, we should review if the test coverage being added is worth the extra CI expense

james-stocks:
Also note this largely copies an existing test; and refactoring the duplicate code should be done at some point

acceptance/tests/pxp-module-puppet/run_puppet_syntax_chars.rb
```diff
@@ -0,0 +1,61 @@
...
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  env_name = test_file_name = File.basename(__FILE__, '.*')
+  environment_name = mk_tmp_environment(env_name)
+
+  step 'On master, create a new environment that includes syntax significant characters in resources and will result in Changed' do
+    site_manifest = "#{environmentpath}/#{environment_name}/manifests/site.pp"
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  notify {'\\'\\"&+/\\()?,.#!@$_-~':}
```
mruzicka: Would it make sense to include some characters outside of ASCII? ☃? Or is this covered by some other test?

james-stocks: pxp-module-puppet/run_puppet_nonASCII.rb already does that.

We _could_ throw all the characters into a single test; but our general testing strategy is to get a detailed test result set, so we want separate test results for multibyte characters and for potential poorly-escaped syntax characters

MikaelSmith: This seems like it mainly ensures pxp-module-puppet handles output safely. Could this be accomplished with a spec test instead?

james-stocks: I think you're right @MikaelSmith - this shouldn't be an acceptance test unless we suspect different behaviours on different platforms. If we trust the Ruby execution of pxp-module-puppet to be consistent across all agent platforms; this can be a spec test.

james-stocks:
Closing this PR - the only thing being tested (that is the responsibility of this repo's code) is the parsing of Puppet stdout by pxp-module-puppet - that can be tested a lot quicker with a spec test

PR #527, opened 2016-12-13
(PCP-592) pxp-module-puppet spec test for syntax chars

james-stocks:
Ensure that pxp-module-puppet works if Puppet's output contains potentially syntax significant characters.


james-stocks:
@MikaelSmith @mruzicka @lucywyman @zreichert this PR replaces PR 523 (which was going to be a new acceptance test)

james-stocks:
Discussing with Michal; I think that this test is fairly pointless for the current version of pxp-module-puppet. It does give us some safeguard if pxp-module-puppet has more processing added to it in the future.

PR #529, opened 2016-12-20
(PCP-645) Append client_type to URI for pcp-broker 1.0 compatibility

MikaelSmith:
pcp-broker 1.0 will include breaking changes for PCP v1 clients, notably
that association requests are deprecated. They'll only be supported for
the agent type without modification; other types can be used if that
type is also appended to the connection URI (the type used for
connection URI and association request must match).


MikaelSmith:
This allows acceptance tests to work against the current pcp-broker#master. This can be tested via the PCP_BROKER_REF variable noted in https://github.com/puppetlabs/pxp-agent/blob/master/acceptance/README.md.

PR #535, opened 2017-01-12
(PCP-660) Add test case for connecting to a PCP v1 endpoint

MikaelSmith:
Adds a test case that verifies sending and receiving messages over a PCP
v1 connection.

[skip ci]

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_v1.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_v1.rb
@@ -1,0 +1,65 @@
+require 'pxp-agent/test_helper.rb'
+
+test_name 'pxp-module-puppet run with PCP v1' do
+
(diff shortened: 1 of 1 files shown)


nicklewis:
Should this test have a `teardown` step that reconnects the agent back to v2?

MikaelSmith:
That seems useful, yes. I'm pretty sure every test actually does its own configuration and restarts pxp-agent, but luckily that doesn't take too long. The benchmark for whether we need teardown would be invalid_ssl_config.rb, which does include it, so I'll add some.

MikaelSmith:
Updated.

MikaelSmith:
Tested with
```
be rake ci:test:aio SHA=e5645317f977e322492b526728084dca6ff9db27 TEST=tests TEST_TARGET=centos7-64a
```

nicklewis:
:+1:

PR #538, opened 2017-01-12
(PCP-685) Skip testing Cisco XR platform

LuvCurves:


james-stocks:
pxp_agent_version.rb , service_stop_start.rb and validate_file_paths.rb will pass for this platform but this PR is skipping them.
I think skipping them is OK because until PCP-612 is resolved, we don't get much value from those test results.

james-stocks:
There is repetitive code here that we might forget to add to new tests; and will take a little more work when we eventually want to change/remove it.
I don't think beaker supports any such thing, but it would be good if the repeated code was bolted onto all tests so it is included automatically.

We're also wasting time on running the pre-suite (with the exception of catching install errors, but I think you'd catch any of those in the other component pre-suites).

These tests could be completely excluded at the Jenkins job config level - adding an expression to the "Combination Filter" under the Configuration Matrix settings would tell Jenkins to skip this cell of the matrix.

MikaelSmith:
If we're not running any of the tests, it does seem better to exit out in the Rakefile or Jenkins config.

LuvCurves:
@MikaelSmith @james-stocks This was merged before I was really ready for it to be. I am against doing this "platform avoidance" in Jenkins, as if this is run outside of Jenkins we should have the same behavior.  I am fine with moving this into the rakefile.  I don't have lots of experience with that but I'll give it a go.

PR #540, opened 2017-01-18
(PCP-365) Acceptance should not use puts or prints

james-stocks:
Replace occurrences of puts or prints with logger entries.
Also improves the message output when get_pcp_broker_status fails to connect, to hint that it is expected if the broker is still starting up.

[skip ci]


james-stocks:
I want to re-target the PR to LTS
Also I want to adjust the logging level used. Will update the PR shortly

james-stocks:
Closing because I can't just change the base branch here in the PR, rebasing is needed

PR #541, opened 2017-01-19
(PCP-365) Acceptance should not use puts or prints

james-stocks:
Replace occurrences of puts or prints with logger entries.
Also improves the message output when get_pcp_broker_status fails to connect, to hint that it is expected if the broker is still starting up.

[skip ci]


james-stocks:
ping @kurtwall @MikaelSmith I moved this commit to LTS

PR #547, opened 2017-01-23
(maint) skipping Cisco XR moved to Rakefile

LuvCurves:


LuvCurves:
@MikaelSmith @james-stocks @kurtwall I took Mikael's suggestion and moved this into the Rakefile to avoid even installing on Cisco XR.

PR #548, opened 2017-01-24
(maint) Reset acceptance to use pcp-broker#master

MikaelSmith:
Updates acceptance tests to operate against pcp-broker#master. This
helps catch compatibility failures between client and master.


MikaelSmith:
@puppetlabs/pxp-agent-maintainers does this seem like a good idea still? I could also update it each time we release pcp-broker (like now, for 1.0.1).

james-stocks:
+1 on using master.

It does mean that this pipeline will catch breaking changes to pcp-broker, when it is only supposed to test pxp-agent; but 1. we will want to know about any such breaking changes and 2. we have a backlogged ticket to add pcp-broker acceptance that would guard against breaking changes in pcp-broker directly.

PR #551, opened 2017-01-27
(PCP-627) Send provisional response for conflicting transaction-id

MikaelSmith:
Changes the behavior of a duplicate nonblocking request to send a
provisional response instead of a pxp error.

Previously we checked to make sure the transaction ID was not
currently in a thread. This commit checks the on disk storage as well
in case the agent has restarted.

### DIFF
--- a/acceptance/tests/pxp-module-puppet/rerun_transaction.rb
+++ b/acceptance/tests/pxp-module-puppet/rerun_transaction.rb
@@ -1,0 +1,60 @@
+require 'pxp-agent/test_helper.rb'
+
+
+STATUS_QUERY_MAX_RETRIES = 60
+STATUS_QUERY_INTERVAL_SECONDS = 1
+
+test_name 'C99777 - two runs with same transaction_id' do
+
+  step 'Ensure each agent host has pxp-agent running and associated' do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+      on agent, puppet('resource service pxp-agent ensure=running')
+      show_pcp_logs_on_failure do
+        assert(is_associated?(master, "pcp://#{agent}/agent"),
+               "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
+      end
+    end
+  end
+
+  target_identities = agents.map {|agent| "pcp://#{agent}/agent"}
+  transaction_ids = []
+  run_results = []
+
+  step "run puppet to generate a transaction_id" do
+    transaction_ids = start_puppet_non_blocking_request(master, target_identities)
+    target_identities.zip(transaction_ids).each do |identity, transaction_id|
+      run_results << check_puppet_non_blocking_response(identity, transaction_id, STATUS_QUERY_MAX_RETRIES, STATUS_QUERY_INTERVAL_SECONDS, "unchanged")
+    end
+  end
+
+  step "restart pxp" do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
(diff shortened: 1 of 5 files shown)


MikaelSmith:
I think I managed to write pretty good unit tests for this.

MikaelSmith:
Tested with `rake ci:test:aio SHA=2becb830e06b9c68143d74efdb384eeb7aee0eec TEST=tests/pxp-module-puppet/rerun_transaction.rb TEST_TARGET=centos7-64a`.

adreyer:
It seems like it might be useful to differentiate which response was returned in response to the module actually starting. Does the pxp version contract allow us to add fields to the schema or will PE break?

Otherwise I think this makes sense and I'm +1 if we can't add more information to the response.

MikaelSmith:
@adreyer I'd probably opt for adding a new message type instead. Adding a field probably wouldn't break anything though.

In a client planning to support older pxp-agent versions that might trigger the old code that sends an error message, we'll still have to handle error message responses, but at least we know the text of that error will never change.

acceptance/tests/pxp-module-puppet/rerun_transaction.rb
```diff
@@ -0,0 +1,60 @@
...
+      # This should fail!
+      response = rpc_request(master, [identity],
+                             'pxp-module-puppet', 'run',
+                             {:flags => ['--onetime',
+                                         '--no-daemonize',
+                                         '--environment', 'production']},
+                                         false,
+                                         transaction_id)[identity]
+
+      assert(response[:envelope][:message_type] == 'http://puppetlabs.com/rpc_provisional_response', 'Did not receive rpc provisional response')
```
james-stocks: Can assert_equal be used here? That would have MiniTest tell us explicitly what the expected and actual message types were. Alternatively update the message to state the actual :message_type

acceptance/tests/pxp-module-puppet/rerun_transaction.rb
```diff
@@ -0,0 +1,60 @@
...
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+      on agent, puppet('resource service pxp-agent ensure=running')
+      show_pcp_logs_on_failure do
+        assert(is_associated?(master, "pcp://#{agent}/agent"),
+               "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
+      end
+    end
+  end
+
+  step "rerun with same transaction_id, expect error and status queryable" do
```
james-stocks: This says "expect error" but (unless I completely misunderstand) there isn't an assertion for getting an error here.

acceptance/tests/pxp-module-puppet/rerun_transaction.rb
```diff
@@ -0,0 +1,60 @@
+require 'pxp-agent/test_helper.rb'
+
+
+STATUS_QUERY_MAX_RETRIES = 60
+STATUS_QUERY_INTERVAL_SECONDS = 1
+
+test_name 'two runs with same transaction_id' do
```
james-stocks: Acceptance tests need to be tracked in Test Rail - I created test case C99777 for this new test. Can the test_name string be updated to 'C99777 - two runs with same transaction_id' please? 
If everything else is done with this PR; go ahead and merge and I'll update this later

MikaelSmith:
I'll fix those things up. There's still possibly a change for @adreyer's suggestion.

MikaelSmith:
@james-stocks updated.

adreyer:
That would mean adding version specific logic to the request processor right?
I don't know if it's worth doing that for a new message type we may never need but I leave that decision up to you

MikaelSmith:
Adding a new flag or a new message would mess up old clients that received that message instead if they use strict schema checking (which ours do). They'd reject the message, and expect that the response timed out. So if we use a new message, it'd have to rely on that logic never being triggered by an old client.

MikaelSmith:
Given that, I think @adreyer and I agree on using this implementation (that should be compatible with old clients). If we need to identify a duplicate transaction, the pxp-agent log will still mention it.

MikaelSmith:
@puppetlabs/pxp-agent-maintainers anymore review on this? Can I get it merged?

PR #557, opened 2017-02-07
(PCP-529) Archive logs from acceptance runs

MikaelSmith:
@james-stocks what would you suggest for addressing the places we delete log files? It'd be useful to keep most of the logs, so maybe we should use different locations for the tests that want to use a fresh log file.

### DIFF
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -17,4 +17,7 @@
     'setup/common/050_Setup_Broker.rb',
     'setup/common/060_Setup_PCP_Client.rb',
   ],
+  :post_suite => [
+    'teardown/common/099_Archive_Logs.rb',
+  ],
 }
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -55,31 +55,6 @@ def expect_file_on_host_to_contain(host, file, expected, seconds=30)
   end
 end
 
-# Show the logs of the broker and agents - useful to see why a test failed
-def show_pcp_logs
-  logger.notify "---- Broker log -----"
-  on(master, "cat /var/log/puppetlabs/pcp-broker.log") do |result|
-    logger.notify result.stdout
-  end
-
-  agents.each do |agent|
-    logger.notify "----- agent #{agent} log -----"
-    on(agent, "cat #{logfile(agent)}") do |result|
-      logger.notify result.stdout
-    end
-  end
-end
-
-# Evaluate the block, show logs if test assertions were triggered
-def show_pcp_logs_on_failure(&block)
-  yield
-rescue MiniTest::Assertion => exception
-  logger.notify "Assertion failed in test: #{exception}"
-  logger.notify "Fetching logs for inspection"
-  show_pcp_logs
-  raise exception
-end
-
 # Some helpers for working with a pcp-broker 'lein tk' instance
 def run_pcp_broker(host, instance=0)
   host[:pcp_broker_instance] = instance
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -5,6 +5,7 @@ require 'net/http'
 require 'openssl'
 require 'socket'
 require 'json'
+require 'securerandom'
 
 # This file contains general test helper methods for pxp-agent acceptance tests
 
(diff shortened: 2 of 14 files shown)


acceptance/config/aio/options.rb
```diff
@@ -16,4 +16,7 @@
     'setup/common/050_Setup_Broker.rb',
     'setup/common/060_Setup_PCP_Client.rb',
   ],
+  :post_suite => [
+    'setup/common/099_Archive_Logs.rb',
```
james-stocks: This perhaps should go into a different folder than 'setup'. I don't think there's a precedent for post-suite folder name in the puppet-agent repos. 'post_suite' perhaps.

acceptance/setup/common/099_Archive_Logs.rb
```diff
@@ -0,0 +1,9 @@
+step 'Archive files created during tests' do
+  ## Copy file(s) from master
+  archive_file_from(master, '/var/log/pcp-broker.log')
+  
+  ## Copy file(s) from agents
+  agents.each do |agent|
+    archive_file_from(agent, '/var/log/puppetlabs/pxp-agent/pxp-agent.log')
```
james-stocks: This won't work for Windows agents - there's a helper method to get the logfile for a given agent - https://github.com/puppetlabs/pxp-agent/blob/master/acceptance/lib/pxp-agent/test_helper.rb#L20

james-stocks:
@MikaelSmith - I was considering this problem with @RandellP last week. Some possible approaches:
1. Have some mechanism to trigger a log roll-over on-demand; then archive all the logs afterwards.
2. Do not wipe the log, and instead count the number of pattern matches before and after your test step. Fail if the count doesn't increment when expected. This approach has a low risk of the log rolling over mid-test. That risk may be acceptable for pxp-agent (we shouldn't ever reach a log rollover).
3. Parse the log timestamps so the test can assert a log entry exists after an expected time. Trickier to code; but this approach fits the paradigm of having the acceptance tests do what a manual tester would do.

Does addressing that block this ticket? I can work on this but can't start for a couple of days.

RandellP:
@MikaelSmith @james-stocks  I am missing a little bit of the context maybe.  Are you just trying not to lose the logs, or are you specifically looking to move them to the control machine.

If looking to move, I would think the log dir is where they should go... so that they get rotated per beaker run and kept with other run specific logs. 
If you are only concerned about losing them (which I agree with), I would make it look like a log rotation happened... which is basically copy to a name with the timestamp type extension.  For puppetserver there is a logback.xml that has format of the backup log file name. 

I am also pondering asking for the puppet server team to expose the logback methods for triggering a log rotation.  And then just calling them.  Agent might also be using logback, but I am not sure...  I think overall it would be good to have a way of doing this for any log file puppet produces.

MikaelSmith:
@james-stocks I'd probably opt to use a different log file for each test.

@RandellP Triggering logback rotation on pcp-broker would solve part of the problem. We restart the broker almost every test, so selecting a different log file name wouldn't be onerous.

MikaelSmith:
This seems to generally do what I want.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -5,24 +5,36 @@
...
 
 # This file contains general test helper methods for pxp-agent acceptance tests
 
 # The standard path for git checkouts is where pcp-broker will be
 GIT_CLONE_FOLDER = '/opt/puppet-git-repos'
 
 # Some helpers for working with the log file
-PXP_LOG_FILE_CYGPATH = '/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/var/log/pxp-agent.log'
-PXP_LOG_FILE_POSIX = '/var/log/puppetlabs/pxp-agent/pxp-agent.log'
+PXP_LOG_FILE_CYGPATH = '/cygdrive/c/ProgramData/PuppetLabs/pxp-agent/var/log'
```
james-stocks: <nitpick>These variable names should now be PXP_LOG_DIR_POSIX or PXP_LOG_PATH_POSIX. The FILE in the variable name was to indicate that they point to the file itself</nitpick>

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -5,24 +5,36 @@
...
+# @return The path to the log file on the host. For Windows, a cygpath is returned
+def logfile(host)
+  logfile_path(host)+'/pxp-agent.log'
+end
+
+def reset_logfile(host)
+  old_logfile = logfile(host)
+  new_logfile = old_logfile+'.'+SecureRandom.uuid
+  logger.notify "---- Saving pxp-agent logfile to #{new_logfile}"
+  retry_on(host, "mv #{old_logfile} #{new_logfile}")
```
james-stocks: I wonder if there's any potential test failure cases where old_logfile does not exist - that would cause a failure here.
I think the logfile can be missing if pxp-agent (the executable) exits very early.

MikaelSmith:
Updated.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -5,22 +5,34 @@
...
-    PXP_LOG_FILE_CYGPATH :
-    PXP_LOG_FILE_POSIX
+  logdir(host)+'/pxp-agent.log'
+end
+
+def reset_logfile(host)
+  old_logfile = logfile(host)
+  new_logfile = old_logfile+'.'+SecureRandom.uuid
+  logger.notify "---- Saving pxp-agent logfile to #{new_logfile}"
+  retry_on(host, "test -f #{old_logfile} && mv #{old_logfile} #{new_logfile}")
```
james-stocks: This still has the issue of erroring due to a non-zero exit code in cases where the log file doesn't exist.
It could be made safe for the case of the log file not existing by inverting the test like:

    retry_on(host, "test ! -f #{old_logfile} || mv #{old_logfile} #{new_logfile}")

MikaelSmith: Fixed.

PR #564, opened 2017-03-16
(PCP-729) Change default ping-interval to 3 minutes

MikaelSmith:
The default ping-interval of 15 seconds, combined with
allowed-keepalive-timeouts of 2, means that if the pcp-broker is
overloaded for more than 45 seconds the agent will disconnect and
reconnect. This can happen when several thousand agents attempt to
connect at once, causing reconnection to spiral out of control.

The ping only has 2 functions: prevent the server from disconnecting the
agent (currently the server has a 15 minute timeout) and detect if the
server has become unavailable without closing the TCP connection
cleanly. This 2nd one doesn't need to happen on such a quick interval.
Change the default ping-interval to 3 minutes to reduce chances of
disconnecting under startup load.


MikaelSmith:
I don't expect this change to be an issue in most scenarios. However, if we feel it's not something that should be changed then I can just do it in PE config.

MikaelSmith:
Actually, this has to clone configured on older agents anyway. I'll probably skip changing the default for now.

PR #565, opened 2017-03-21
(PCP-729) Shorten timeout for failover in tests

MikaelSmith:
After changing the default ping interval, failover tests need to reset
it to run in a reasonable time.


MikaelSmith:
@james-stocks mind looking at this? I wish I had a way to shutdown the process that allowed Jetty to close connections. I'm not sure why this is breaking them off so hard.

PR #567, opened 2017-03-22
(maint) Skip nonASCII test on japanese windows machines

MikaelSmith:
PUP-5756 should fix an issue in Puppet with UTF-8 characters in some
locales when writing to event log. Until that's fixed, skip the test.


MikaelSmith:
It's not tested in LTS.

PR #570, opened 2017-03-25
(PCP-729) Change default ping-interval to 2 minutes

MikaelSmith:
With PCP-734 fixed, pxp-agent will attempt to reconnect immediately
unless the broker disappears and the TCP connection is not closed. In
that case, it'll take 4-6 minutes to detect a failure.

This interval is chosen to minimize disconnects under reasonable loads
for pcp-broker, e.g. 10k nodes reconnecting.

Adjust the hard failover test to use a short ping-interval to minimize
test time.


MikaelSmith:
I might see about targeting this at stable.

PR #572, opened 2017-03-28
(maint) Remove timesync step

MikaelSmith:
It's no longer needed for tests to succeed, and is slow/unreliable and
already skipped on several platforms.

[skip ci]


james-stocks:
It would be great to get rid of this.
Was there a change to pcp-broker to move away from expiring based on a 'created' timestamp? I don't see any relevant recent commits on pcp-broker#master
There were some occasions where VMs in our test infrastructure had substantially (>20 second) different clock times so I think this change is only safe if we now don't care about synced clock times at all.

MikaelSmith:
pcp-broker 1.0 removed all message redelivery, and with it the concept of message expiration. So even though PCP v1 messages can still include an expires field, it's ignored by the broker (and was always ignored by agents). There should be nothing depending on relative system times.

MikaelSmith:
```
      Test Suite: post_suite @ 2017-03-28 13:23:09 -0700

      - Host Configuration Summary -


              - Test Case Summary for suite 'post_suite' -
       Total Suite Time: 0.98 seconds
      Average Test Time: 0.98 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```

PR #583, opened 2017-05-05
(maint) Add testing around exec via Puppet

MikaelSmith:
PE-14681 and MODULES-3125 identified issues running commands via Puppet
exec from other service agents. These issues led to PUP-6675: selinux
prevents processes from writing to files in /tmp if they don't have a
reason to do so. The Puppet service selinux policy appears to exempt
this, but other services that invoke Puppet - such as pxp-agent - aren't
configured that way and cause silent exec failures.

PUP-6675 fixes this issue. We add an acceptance test to ensure that it
doesn't recur in pxp-agent.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -291,11 +291,6 @@ def rpc_request(broker, targets,
     end
   end
 
-  message = PCP::Message.new({
-    :message_type => blocking ? 'http://puppetlabs.com/rpc_blocking_request' : 'http://puppetlabs.com/rpc_non_blocking_request',
-    :targets => targets
-  })
-
   message_data = {
     :transaction_id => transaction_id,
     :module         => pxp_module,
@@ -305,13 +300,22 @@ def rpc_request(broker, targets,
   if !blocking then
     message_data[:notify_outcome] = false
   end
-  message.data = message_data.to_json
 
-  message_expiry = 10 # Seconds for the PCP message to be considered failed
+  targets.each do |target|
+    message = PCP::Message.new({
+      :message_type => blocking ? 'http://puppetlabs.com/rpc_blocking_request' : 'http://puppetlabs.com/rpc_non_blocking_request',
+      :targets => [target]
+    })
+
+    message.data = message_data.to_json
+
+    message_expiry = 10 # Seconds for the PCP message to be considered failed
+    message.expires(message_expiry)
+
+    client.send(message)
+  end
+
   rpc_action_expiry = 60 # Seconds for the entire RPC action to be considered failed
-  message.expires(message_expiry)
-
-  client.send(message)
 
   begin
     Timeout::timeout(rpc_action_expiry) do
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Depends on https://github.com/puppetlabs/puppet/pull/5844 being fixed first.

MikaelSmith:
Should be ready to go with the latest build of puppet-agent.

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -305,13 +300,22 @@ def rpc_request(broker, targets,
...
+    message = PCP::Message.new({
+      :message_type => blocking ? 'http://puppetlabs.com/rpc_blocking_request' : 'http://puppetlabs.com/rpc_non_blocking_request',
+      :targets => [target]
+    })
 
-  client.send(message)
+    message.data = message_data.to_json
+
+    message_expiry = 10 # Seconds for the PCP message to be considered failed
+    message.expires(message_expiry)
```
mruzicka: Should we get rid of the message_expiry variable altogether?

acceptance/tests/pxp-module-puppet/run_puppet_killed_puppet.rb
```diff
@@ -75,6 +70,10 @@
       else
         on agent, "kill -9 #{pid}"
       end
+
+      # Also halt the sleep process here, as on Windows the parent process won't exit while the child process
+      # has an open handle shared with it (the stdout pipe).
+      stop_sleep_process(agents)
```
mruzicka: `agents` -> `agent`
otherwise there is an attempt to kill the sleep process on *all* agents on every iteration through the `agents` array

acceptance/lib/pxp-agent/test_helper.rb
MikaelSmith: It's no longer needed, so we could remove it.

acceptance/tests/pxp-module-puppet/run_puppet_killed_puppet.rb
MikaelSmith: Ah, that makes sense.

acceptance/lib/pxp-agent/test_helper.rb
MikaelSmith: Tempted to ignore it for now though.

acceptance/tests/pxp-module-puppet/run_puppet_exec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  env_name = test_file_name = File.basename(__FILE__, '.*')
+  environment_name = mk_tmp_environment(env_name)
+
+  step 'On master, create a new environment that execs hostname' do
+    site_manifest = "#{environmentpath}/#{environment_name}/manifests/site.pp"
+    create_remote_file(master, site_manifest, <<-SITEPP)
+node default {
+  exec { "hostname":
+    path => ["/bin", "/usr/bin", "C:/cygwin32/bin", "C:/cygwin64/bin"],
```
joshcooper: @MikaelSmith this is failing CI when running 32-bit puppet/ruby on 32-bit windows (using the `win-10-ent-i386` template), because the path needs to be `C:/cygwin/bin`. I think the `C:/cygwin32` would only happen when using 32-bit cygwin on 64-bit windows, which shouldn't happen in our current set of windows templates.

MikaelSmith: https://github.com/puppetlabs/pxp-agent/pull/605

PR #584, opened 2017-05-12
(PCP-745) Add manpage name to Solaris SMF manifest

abottchen:
Prior to this commit, the svcs -xv command would
return a libscf error when encountering the unset
name field in the manpage property for pxp-agent.
This commit adds a string to that setting to bring
it into alignment with other Puppet services.


MikaelSmith:
This is a good start. We should add a manpage as well, but that can probably be a separate ticket.

MikaelSmith:
Can you rebase this on the stable branch?

abottchen:
@MikaelSmith I think I have it rebased to stable now.

PR #587, opened 2017-05-22
(PCP-699) Avoid undefined variables when running puppet

nicklewis:
Because of the circuitous route to which the final version of this code was
arrived, as well as unit test coverage lacking at the outer levels, there are a
couple of places where arguments aren't properly set/used on the ModulePuppet
instance. This PR fixes those and improves the test coverage to exercise as
much of the module as possible.


jonathannewman:
👍

PR #589, opened 2017-05-26
(PCP-603) negative test case for bypassing whitelist

zreichert:


acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
```diff
@@ -0,0 +1,58 @@
+require 'pxp-agent/test_helper.rb'
+
+test_name 'C100245 - attempt bypass of flag whitelist with tab character' do
+
+  confine :except, :platform => ['windows']
```
james-stocks: Why not Windows? Would be useful to add a comment explaining the exclusion.

It could be valuable to see this test pass on all platforms - to ensure that the latest pxp-module-puppet script has been packaged

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
```diff
@@ -0,0 +1,58 @@
...
+    create_remote_file(@test_host, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
+    on @test_host, puppet('resource service pxp-agent ensure=running')
+
+    assert(is_associated?(master, "pcp://#{@test_host}/agent"),
+           "Agent #{@test_host} with PCP identity pcp://#{@test_host}/agent should be associated with pcp-broker")
+  end
+
+  step "Send an rpc_blocking_request to all agents" do
+    target_identities = []
+    target_identities << "pcp://#{@test_host}/agent"
```
james-stocks: You could save a couple of lines here and just pass ["pcp://#{@test_host}/agent"] directly to rpc_request

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
```diff
@@ -0,0 +1,58 @@
...
+    begin
+      rpc_request(master, target_identities,
+                                       'pxp-module-puppet', 'run',
+                                       {:env => [], :flags => ['--noop',
+                                                               '--onetime',
+                                                               '--no-daemonize',
+                                                               "\t--postrun_command #{@executable_path}"]
+                                       })
+    rescue => exception
+      fail("Exception occurred when trying to run Puppet on all agents: #{exception.message}")
```
james-stocks: This message could be more specific to this case (only one agent, and sending a deliberately dodgy flag).

Also a minor point - _fail_ can be confused with _fail_test_ , suggest changing it to either _fail_test_ or _raise_ depending on whether you intend a Failed or Errored test result (I think in this case you want _Errored_ )

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
```diff
@@ -0,0 +1,58 @@
...
+test_name 'C100245 - attempt bypass of flag whitelist with tab character' do
+
+  confine :except, :platform => ['windows']
+
+  step 'pick test host' do
+    @test_host = agents.first
+  end
+
+  step 'set up executable file' do
+    @executable_path = [@test_host.system_temp_path, 'run_me'].join('/')
```
mruzicka: I'd suggest using [`File::join`](https://ruby-doc.org/core-2.4.0/File.html#method-c-join) instead of `Array#join`.

mruzicka:
Looks good to me too. Although I agree with @james-stocks it would be worth making it work on windows too.

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
zreichert: Good point, this was copy pasta from another test.

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
zreichert: Agreed.  I limited my first attempt for simplicities sake.

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
zreichert: Fixed

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
zreichert: Fixed

PR #590, opened 2017-05-30
(MAINT) skip whitelist test on windows

zreichert:


acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
```diff
@@ -6,6 +6,7 @@
 
   step 'pick test host' do
     @test_host = agents.first
+    skip_test 'Test not windows compatible' if @test_host.platform =~ /win/
```
joshcooper: In puppet land we use `/windows/` so it doesn't partially match something like windriver

zreichert: fixed

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
```diff
@@ -6,6 +6,7 @@
 
   step 'pick test host' do
     @test_host = agents.first
+    skip_test 'Test not windows compatible' if @test_host.platform =~ /windows/
```
joshcooper: @zreichert so this is still failing because line https://github.com/zreichert/pxp-agent/blob/eb91202bddd6da394bd5f80fff19dc8debe8608c/acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb#L5 causes `@agents` array to be empty when `windows` is the only agent platform, so then `[].first` is nil.

james-stocks:
Also note this repo already has a windows?(host) helper in https://github.com/puppetlabs/pxp-agent/blob/master/acceptance/lib/pxp-agent/config_helper.rb#L12

I think ideally, that method belongs in beaker and not this repo; but for consistency with existing tests it might be preferable to use it

acceptance/tests/pxp-module-puppet/negative/attempt_puppet_flag_whitelist_bypass.rb
zreichert: @joshcooper Sorry for the churn.  Seems like the best solution is to remove the confine.  I'll make a new PR.

PR #592, opened 2017-06-01
(PCP-750) Add inventory retries to flaky tests

johnduarte:
This commit adds inventory retries to tests that have been intermittently
failing to get responses to inventory requests.

These test failures have not been consistent between runs or platforms. They
appear to be sensitive to starting or restarting the service.

### DIFF
--- a/acceptance/tests/invalid_ssl_config.rb
+++ b/acceptance/tests/invalid_ssl_config.rb
@@ -36,7 +36,8 @@ agents.each do |agent|
     create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
     on agent, puppet('resource service pxp-agent ensure=running')
 
-    assert(is_associated?(master, "pcp://#{agent}/agent"),
+    inventory_retries = 60
+    assert(is_associated?(master, "pcp://#{agent}/agent", inventory_retries),
            "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
   end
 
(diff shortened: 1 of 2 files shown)


acceptance/tests/invalid_ssl_config.rb
```diff
@@ -36,7 +36,8 @@
     create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_json_using_puppet_certs(master, agent).to_s)
     on agent, puppet('resource service pxp-agent ensure=running')
 
-    assert(is_associated?(master, "pcp://#{agent}/agent"),
+    inventory_retries = 60
+    assert(is_associated?(master, "pcp://#{agent}/agent", inventory_retries),
```
joshcooper: It looks like the default is already 60: https://github.com/puppetlabs/pxp-agent/blob/274b0265cbeb3824df1fbb96d9b35e34469dfdf1/acceptance/lib/pxp-agent/test_helper.rb#L204. Makes me think this test is subject to race conditions...

james-stocks:
@johnduarte This sounds like PCP-394 which was never resolved but disappeared from CI for some time.

If this issue is the same (or at least similar symptoms) then retrying the inventory check will not help - the one single chance for the broker to send the inventory response has failed, checking again a second later won't get a different result. 

We _could_ try a retry on stop/starting the pxp-agent service on the agent and then checking the inventory. A concern with that approach is that we may be glossing over a customer-affecting bug.

cc @mruzicka @MikaelSmith

acceptance/tests/invalid_ssl_config.rb
james-stocks: The retries here occur when the test receives an inventory response from the broker and that inventory doesn't contain the expected _pcp://#{agent}/agent_ identity.

See my comment below re. a workaround (and a concern with it)

johnduarte:
Closing PR since this is not the root cause of the issue.

PR #597, opened 2017-06-22
(PCP-754) Support stopping service when the binary path changes

MikaelSmith:
From pxp-agent 1.5.3 to 1.6.0, pxp-agent used a versioned binary with
pxp-agent symlinked to it. When upgrading, the binary used to start the
pxp-agent service would change. On SUSE, this prevented restarting the
service after upgrade because killproc would not kill the process
started with the previous binary. Use kill instead - with appropriate
guards to check it was the expected service process - to ensure we can
restart the service after upgrading from one of those versions.

### DIFF
--- a/ext/suse/pxp-agent.init
+++ b/ext/suse/pxp-agent.init
@@ -85,8 +85,24 @@ case "$1" in
         ## Stop daemon with killproc(8) and if this fails
         ## set echo the echo return value.
 
-        killproc -QUIT -p "${pidfile}" "${exec}" && rm -f "${lockfile}" "${pidfile}"
+        ## This accounts for the behavior of killproc on sles 11:
+        ## if the binary we're attempting to halt is a symlink to
+        ## a versioned binary, then it won't be halted if that
+        ## symlink is changed to point to a different file name.
+        ## Check to see if a running process matches the contents
+        ## of the pidfile, and if so kill it. This could fail if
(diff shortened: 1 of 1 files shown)


ext/suse/pxp-agent.init
```diff
@@ -85,8 +85,21 @@ case "$1" in
         ## Stop daemon with killproc(8) and if this fails
         ## set echo the echo return value.
 
-        killproc -QUIT -p "${pidfile}" "${exec}" && rm -f "${lockfile}" "${pidfile}"
+        ## This accounts for the behavior of killproc on sles 11.
+        ## Check to see if a running process matches the contents
+        ## of the pidfile, and if so kill it. This could fail if
+        ## multiple pxp-agent processes are running, in which case
+        ## killproc may succeed or manual intervention is required.
```
branan: Can we expand this comment a bit with why we need to special-case killproc for ourselves?

MikaelSmith: Yeah, I had more in the commit message but makes sense to put it in the comment.

MikaelSmith: Updated.

MikaelSmith:
I don't think this is actually fixing the upgrade problem :(

MikaelSmith:
I can't get this to actually work. Going back to a puppet-agent upgrade fix for now.

PR #598, opened 2017-06-22
(PCP-755) Support restarting service when the binary path changes

MikaelSmith:
From pxp-agent 1.5.3 to 1.6.0, pxp-agent used a versioned binary with
pxp-agent symlinked to it. When upgrading, the binary used to start the
pxp-agent service would change. On SUSE, this prevented restarting the
service after upgrade because checkproc would not detect the existing
process (needed by condrestart) and killproc would not kill the process
started with the previous binary. Use pgrep/kill instead - with
appropriate guards to check it was the expected service process - to
ensure we can restart the service after upgrading from one of those
versions.

[skip ci]

### DIFF
--- a/ext/suse/pxp-agent.init
+++ b/ext/suse/pxp-agent.init
@@ -85,8 +85,24 @@ case "$1" in
         ## Stop daemon with killproc(8) and if this fails
         ## set echo the echo return value.
 
-        killproc -QUIT -p "${pidfile}" "${exec}" && rm -f "${lockfile}" "${pidfile}"
+        ## This accounts for the behavior of killproc on sles 11:
+        ## if the binary we're attempting to halt is a symlink to
+        ## a versioned binary, then it won't be halted if that
+        ## symlink is changed to point to a different file name.
+        ## Check to see if a running process matches the contents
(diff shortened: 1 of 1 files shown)


geoffnichols:
This appears to be working as intended. 👍 

**Scenario 1 - Running pxp-agent service is restarted on upgrade.**
1. Built a test puppet-agent package with head of puppet-agent#master, using this changeset's ref for the pxp-agent component.
2. Installed puppet-enterprise-2017.2.2 with puppet-agent-1.10.4.
3. Confirmed pxp-agent was running.
4. Upgraded puppet-agent to the test 4.99 series  puppet-agent package.
5. After the upgrade, confirmed pxp-agent was running, using a new PID (so restarted successfully).

**Scenario 2 - Stopped pxp-agent service is not started on upgrade.**
Using same test puppet-agent package built with head of puppet-agent#master:
1. Installed puppet-enterprise-2017.2.2 with puppet-agent-1.10.4.
2. Confirmed pxp-agent was not running.
3. Upgraded puppet-agent to the test 4.99 puppet-agent package.
4. After the upgrade, confirmed pxp-agent still not running.

PR #599, opened 2017-06-23
(maint) Log version and level at startup

MikaelSmith:
Logs pxp-agent version and configured log level as soon as logging is
configured. Clarifies when versions change in logging, which makes
debugging issues easier.

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -3,6 +3,8 @@
 
 #include <pxp-agent/util/daemonize.hpp>
 
+#include "version-inl.hpp"
+
 #include <cpp-pcp-client/util/thread.hpp>
 #include <cpp-pcp-client/util/chrono.hpp>
 
@@ -182,8 +184,8 @@ int main(int argc, char *argv[]) {
     // Set up logging
 
     try {
-        Configuration::Instance().setupLogging();
-        LOG_DEBUG("pxp-agent logging has been initialized");
+        auto loglevel = Configuration::Instance().setupLogging();
+        LOG_INFO("pxp-agent {1} started at {2} level", PXP_AGENT_VERSION, loglevel);
     } catch (const Configuration::Error& e) {
         boost::nowide::cout << lth_loc::format("Failed to configure logging: {1}\n"
                                                "Cannot start pxp-agent", e.what());
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -189,7 +189,7 @@ static void validateLogDirPath(const std::string& logfile)
     }
 }
 
-void Configuration::setupLogging()
+std::string Configuration::setupLogging()
 {
     logfile_ = HW::GetFlag<std::string>("logfile");
     pcp_access_logfile_ = HW::GetFlag<std::string>("pcp-access-logfile");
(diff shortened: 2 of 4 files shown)


MikaelSmith:
Default output now is
```
2017-06-27 08:10:44.892718 INFO  puppetlabs.pxp_agent.configuration:271 - pxp-agent 1.6.0 started at info level
2017-06-27 08:10:44.899226 INFO  puppetlabs.pxp_agent.main:201 - pxp-agent configuration has been validated
2017-06-27 08:10:44.899321 INFO  puppetlabs.pxp_agent.agent:29 - Connecting using PCP v2
2017-06-27 08:10:44.899484 INFO  puppetlabs.cpp_pcp_client.client_metadata:125 - Retrieved common name from the certificate and determined the client URI: pcp://client02.example.com/agent
2017-06-27 08:10:44.899836 INFO  puppetlabs.pxp_agent.request_processor:767 - Loading external modules configuration from /etc/puppetlabs/pxp-agent/modules
2017-06-27 08:10:44.900017 INFO  puppetlabs.pxp_agent.request_processor:818 - Loading external modules from /Users/michaelsmith/puppetlabs/pxp-agent/modules
2017-06-27 08:10:46.351412 INFO  puppetlabs.pxp_agent.results_storage:221 - About to purge the results directories from '/Users/michaelsmith/puppetlabs/pxp-agent/dev-resources/spool'; TTL = 14d
2017-06-27 08:10:46.352860 INFO  puppetlabs.pxp_agent.results_storage:270 - Removed 0 directories from '/Users/michaelsmith/puppetlabs/pxp-agent/dev-resources/spool'
2017-06-27 08:10:46.353056 INFO  puppetlabs.pxp_agent.request_processor:908 - Starting the task for purging the spool directory every 24192 minutes; thread id 0x700006777000
2017-06-27 08:10:46.354018 INFO  puppetlabs.cpp_pcp_client.connection:395 - Establishing the WebSocket connection with 'wss://broker.example.com:8142/pcp2/agent' with a timeout of 5000 ms
```

The only weird thing is the first message comes from `puppetlabs.pxp_agent.configuration`, but plumbing the version and log level back to main would be a bit messier.

exe/main.cc
```diff
@@ -182,8 +182,9 @@ int main(int argc, char *argv[]) {
     // Set up logging
 
     try {
-        Configuration::Instance().setupLogging();
-        LOG_DEBUG("pxp-agent logging has been initialized");
+        std::string version, loglevel;
+        std::tie(version, loglevel) = Configuration::Instance().setupLogging();
+        LOG_INFO("pxp-agent {1} started at {2} level", version, loglevel);
```
mruzicka: I'd suggest returning only the log level from the `Configuration::Instance().setupLogging()` method and using the `PXP_AGENT_VERSION` here directly. The rationale being that the `PXP_AGENT_VERSION` define is relevant to all of the pxp-agent code not just logging, so it seems inappropriate to return its value from the logging setup function. Moreover by simplifying the `setupLogging()`'s return value we'll be able to avoid having to pack/unpack the two values to/from the `std::pair`.

lib/src/configuration.cc
```diff
@@ -268,14 +268,14 @@ void Configuration::setupLogging()
...
-    LOG_DEBUG("Logging configured");
-
     if (!log_on_stdout) {
         // Configure platform-specific things for file logging
         // NB: we do that after setting up lth_log in order to log in
         //     case of failure
         configure_platform_file_logging();
     }
+
+    return std::make_pair(PXP_AGENT_VERSION, loglevel);
```
mruzicka: As suggested above: Return just the log level from the method.

exe/main.cc
MikaelSmith: Updated.

PR #602, opened 2017-07-05
(PCP-758) A simple module wrapper in C++

MikaelSmith:


lib/src/action_response.cc
```diff
@@ -133,6 +135,36 @@ void ActionResponse::setValidResultsAndEnd(lth_jc::JsonContainer&& results,
...
+        }
+    }
+    return true;
+}
+
+void ActionResponse::setValidResultsAndEnd(std::string results,
+                                           const std::string& execution_error)
+{
+    // Confirm results is valid UTF-8. Throw a data_parse_error if not.
+    if (!isValidJson(results)) {
```
MikaelSmith: Need to restore some tests around this.

lib/tests/unit/external_module_test.cc
```diff
@@ -327,38 +279,13 @@ TEST_CASE("ExternalModule::executeAction", "[modules][output]") {
         ActionRequest request { RequestType::Blocking, convert_content };
 
         try {
-            REQUIRE(e_m.executeAction(request)
-                        .action_metadata
-                            .get<double>({ "results", "amount" }) == 1098.9);
+            auto result = e_m.executeAction(request);
+            CAPTURE(result.action_metadata.toString());
+            REQUIRE(result.action_metadata.get<std::string>("results") == "{\"amount\":1098.9}\n");
```
MikaelSmith: Account for `\r\n` on Windows.

MikaelSmith:
I think this is finally ready.

exe/task.cc
```diff
@@ -0,0 +1,174 @@
...
+    return "/opt/puppetlabs";
+}
+#endif
+
+#else
+#define SYSTEM_PREFIX test_prefix
+string test_prefix();
+#endif
+
+static string wrap_error(string err)
```
adreyer: Lets try to use the errors we've specced out so far here. ie

{"_error": { "kind", "msg", details}}

MikaelSmith: Updated. I didn't handle parsing stdout for an error object, not quite sure where we want to handle that.

exe/task.cc
```diff
@@ -0,0 +1,192 @@
...
+                    set_error(stdout_result, "output-encoding-error", _("Output cannot be represented as a JSON string"));
+                } else {
+                    stdout_result.set("output", exec.output);
+                }
+
+                stderr_str = exec.error;
+                exitcode = exec.exit_code;
+
+                if (exitcode != 0) {
+                    set_error(stdout_result, "task-error", _("The task errored with a code {1}", exitcode));
```
adreyer: I think we should handle this error in the service. Otherwise we might clobber better error output with this generic output

MikaelSmith: Oh yeah. I had a question about parsing for _error somewhere as well. Since we probably won't parse for it here, generating an _error key is somewhat redundant.

Although currently, what I'm returning is `{"output": "...", "_error": "..."}` which means nothing specifically gets clobbered.

adreyer: ah ok.

PR #603, opened 2017-07-06
(PCP-759) Enable EC2 master install

johnduarte:
This commit updates the acceptance install pre-suite to allow installing
an ec2 master dynamically on a el based system without intervention.
Due to the expectations of the nightly puppetserver repo, the nightly
puppet-agent is installed first and then over-written with the
puppet-agent version under test. The specification of a SERVER_VERSION
is also supported.

A `LEIN_PROFILE` constant has been added to switch between profiles
based on whether the SUT has access to the `puppetlabs.net` network
or not to enable the installation of pcp-broker on the master.

### DIFF
--- a/acceptance/setup/aio/010_Install.rb
+++ b/acceptance/setup/aio/010_Install.rb
@@ -18,30 +18,33 @@ step "Install puppet-agent..." do
   end
 end
 
-MASTER_PACKAGES = {
-  :redhat => [
-    'puppetserver',
-  ],
-  :debian => [
-    'puppetserver',
-  ],
-}
-
 step "Install puppetserver..." do
-  if ENV['SERVER_VERSION']
-    install_puppetlabs_dev_repo(master, 'puppetserver', ENV['SERVER_VERSION'])
-    install_puppetlabs_dev_repo(master, 'puppet-agent', ENV['SHA'])
-    master.install_package('puppetserver')
+  if master[:hypervisor] == 'ec2' && master[:platform].match(/(?:el|centos|oracle|redhat|scientific)/)
+    variant, version = master['platform'].to_array
+    server_num = ENV['SERVER_VERSION'].to_i
+    # nil or 'latest' will convert to 0
+    if server_num > 0 && server_num < 5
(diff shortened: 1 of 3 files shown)


acceptance/setup/aio/010_Install.rb
```diff
@@ -29,19 +29,30 @@
 
 step "Install puppetserver..." do
   if ENV['SERVER_VERSION']
-    install_puppetlabs_dev_repo(master, 'puppetserver', ENV['SERVER_VERSION'])
-    install_puppetlabs_dev_repo(master, 'puppet-agent', ENV['SHA'])
-    master.install_package('puppetserver')
+    server_version = ENV['SERVER_VERSION']
+    if master[:hypervisor] == 'ec2' && master[:platform].match(/(?:el|centos|oracle|redhat|scientific)/)
```
MikaelSmith: Why not just make this the code to use if we want nightly? i.e. SHA=nightly SERVER_VERSION=5 or something. Having it specific to the ec2 instance seems a bit odd.

acceptance/setup/aio/010_Install.rb
```diff
@@ -29,19 +29,30 @@
...
+    else
+      install_puppetlabs_dev_repo(master, 'puppetserver', ENV['SERVER_VERSION'])
+      install_puppetlabs_dev_repo(master, 'puppet-agent', ENV['SHA'])
+    end
   else
     # beaker can't install puppetserver from nightlies (BKR-673)
     repo_configs_dir = 'repo-configs'
     install_repos_on(master, 'puppetserver', 'nightly', repo_configs_dir)
     install_repos_on(master, 'puppet-agent', ENV['SHA'], repo_configs_dir)
-    install_packages_on(master, MASTER_PACKAGES)
```
MikaelSmith: MASTER_PACKAGES is unused, remove it?

acceptance/setup/aio/010_Install.rb
johnduarte: @MikaelSmith I have attempted to clarify this. Please let me know if it is still confusing.

PR #605, opened 2017-07-08
(maint) Fix the test on win-10-ent-i386

MikaelSmith:
Cygwin is installed at C:/cygwin, not C:/cygwin32.

[skip ci]


joshcooper:
Works as expected on 32-bit Windows 10.

PR #607, opened 2017-07-11
(PCP-750) Ensure failed connection attempts are closed

MikaelSmith:
Ensure failed attemption connects are closed. If not done, we can end up
with multiple connections, and an earlier attempt can supersede a later
one. When that happens, the connection being used for testing will fail
to receive expected responses.

[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -372,9 +372,12 @@ def connect_pcp_client(broker)
     })
     connected = client.connect(5)
     retries += 1
+
+    # If the connection was not established, close it. Otherwise we can end up with 2 successful connections
+    # and get an earlier attempt superseding later connections.
+    client.close if !connected
   end
   if !connected
-    client.close
     raise "Controller PCP client failed to connect with pcp-broker on #{broker} after #{max_retries} attempts: #{client.inspect}"
   end
   if !client.associated?


MikaelSmith:
Passed a Centos run: `be rake ci:test:aio SHA=1.10.4 SUITE_VERSION=1.10.4 SERVER_VERSION=2.7.2 TEST_TARGET=centos7-64a TEST=tests`
```
       Total Suite Time: 389.28 seconds
      Average Test Time: 17.69 seconds
              Attempted: 22
                 Passed: 21
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 22

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
  Test Case tests/multiple_pxp_agent_daemon.rb
Pending Tests Cases:
```

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -372,9 +372,12 @@ def connect_pcp_client(broker)
     })
     connected = client.connect(5)
     retries += 1
+
+    # If the connection was not established, close it. Otherwise we can end up with 2 successful connections
+    # and get an earlier attempt superseding later connections.
+    client.close if !connected
```
MikaelSmith: On inspecting https://github.com/puppetlabs/ruby-pcp-client/blob/master/lib/pcp/client.rb#L222, https://github.com/faye/faye-websocket-ruby/blob/master/lib/faye/websocket/api.rb#L79, and https://github.com/faye/websocket-driver-ruby/blob/master/lib/websocket/driver.rb#L126, I believe this will not throw an exception.

nicklewis: Should we have this for `!associated?` as well?

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -372,13 +372,15 @@ def connect_pcp_client(broker)
     })
     connected = client.connect(5)
     retries += 1
+
+    # If the connection was not established, close it. Otherwise we can end up with 2 successful connections
+    # and get an earlier attempt superseding later connections.
+    client.close if (!connected || !client.associated?)
```
nicklewis: Actually, it looks like `connect()` will return `nil` if association times out. So closing the connection on `!connected` alone seems sufficient.

PR #608, opened 2017-07-11
(PCP-758) Load .exe modules on Windows

MikaelSmith:


MikaelSmith:
A few tests are going to fail because `type -` doesn't work. Need to fix that up.

PR #609, opened 2017-07-12
(PCP-758) Support file exts on non-Windows, powershell/ruby on Windows

MikaelSmith:
Accepts file extensions on non-Windows platforms, except the reserved extensions `.json` or `.md`. That allows writing cross-platform tasks in scripting languages.

Adds support for cross-platform scripts via powershell and ruby.

### DIFF
--- a/exe/task.cc
+++ b/exe/task.cc
@@ -10,6 +10,7 @@
 #include <boost/nowide/iostream.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string/join.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 
 #include <rapidjson/rapidjson.h>
 
@@ -100,6 +101,15 @@ static string find_task(string taskname)
     }
 
     auto filename = filenames.front();
+
+#ifdef _WIN32
+    // Hard-code .ps1 and .rb as executable on Windows. On non-Windows, we still rely on permissions.
+    auto extension = fs::path(filename).extension();
+    if (extension == ".ps1" || extension == ".rb") {
+        return filename;
+    }
+#endif
+
     if (lth_exec::which(filename).empty()) {
         throw task_error("not-executable", _("Task file '{1}' is not executable", filename));
     }
@@ -198,11 +208,30 @@ int MAIN_IMPL(int argc, char** argv)
 
     try {
         auto filename = find_task(taskname);
+        auto arguments = vector<string>{};
         auto environment = generate_environment_from(input);
+
+#ifdef _WIN32
+        if (boost::algorithm::ends_with(filename, ".ps1")) {
+            arguments = vector<string>{
+                "-NoProfile",
+                "-NonInteractive",
+                "-NoLogo",
+                "-ExecutionPolicy",
+                "Bypass",
+                "-File",
+                filename
+            };
+            filename = "powershell";
+        } else if (boost::algorithm::ends_with(filename, ".rb")) {
+            arguments = vector<string>{filename};
+            filename = "ruby";
+        }
+#endif
 
         auto exec = lth_exec::execute(
             filename,
-            {},  // arguments
+            arguments,
             input.toString(),
             environment,
             0,  // timeout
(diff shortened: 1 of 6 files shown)


exe/task.cc
```diff
@@ -53,6 +55,67 @@ static string system_prefix()
...
+    string module, task = "init";
+    if (!lth_util::re_search(taskname, boost::regex("\\A(\\w+)\\z"), &module) &&
+        !lth_util::re_search(taskname, boost::regex("\\A(\\w+)::(\\w+)\\z"), &module, &task)) {
+        throw task_error("invalid-task", _("Invalid task name '{1}'", taskname));
+    }
+
+    auto filepath = SYSTEM_PREFIX()+"/pxp-agent/tasks/"+module+"/tasks";
+
+    // Search for any executable file starting with the task name and not matching a set of reserved extensions
+    auto reserved_extensions = vector<string>{".json", ".md"};
```
mruzicka: This seems to be unused.

exe/task.cc
```diff
@@ -142,41 +205,51 @@ int MAIN_IMPL(int argc, char** argv)
...
+            arguments,
+            input.toString(),
+            environment,
+            0,  // timeout
+            { lth_exec::execution_options::thread_safe,
+              lth_exec::execution_options::merge_environment,
+              lth_exec::execution_options::inherit_locale });
+
+        if (!isValidUTF8(exec.output)) {
+            set_error(stdout_result, "output-encoding-error", _("Output cannot be represented as a JSON string"));
```
mruzicka: Should we simply say the output is not valid UTF-8?

MikaelSmith:
I plan to add more testing, and make the shebangs in tests added on Windows as well (since they'll be there in cross-platform tasks).

exe/task.hpp
```diff
@@ -0,0 +1,44 @@
+#include <vector>
```
MikaelSmith: Needs `#pragma once`.

exe/task.cc
```diff
@@ -9,12 +11,11 @@
 
 #include <boost/nowide/iostream.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/algorithm/string/join.hpp>
+#include <boost/algorithm/string/predicate.hpp>
```
MikaelSmith: No longer used.

exe/task.cc
```diff
@@ -53,6 +55,56 @@ static string system_prefix()
...
+
+    // Search for any executable file starting with the task name and not matching a set of reserved extensions
+    vector<string> filenames;
+    string extension;
+
+    try {
+        fs::directory_iterator end;
+        for (auto f = fs::directory_iterator(filepath); f != end; ++f) {
+            if (!fs::is_directory(f->status()) && f->path().stem() == task) {
+                extension = f->path().extension().string();
```
MikaelSmith: Can be overwritten if json/md present.

exe/task.cc
```diff
@@ -53,6 +55,56 @@ static string system_prefix()
...
+    }
+
+    auto filename = filenames.front();
+
+    auto builtin = builtin_tasks_.find(extension);
+    if (builtin != builtin_tasks_.end()) {
+        return builtin->second(filename);
+    }
+
+    if (lth_exec::which(filename).empty()) {
```
adreyer: Is this always true on windows?

MikaelSmith: No. For .bat, .cmd, .exe, or .com it will return true. And seems to execute at least .bat and .exe correctly.

exe/task.hpp
```diff
@@ -0,0 +1,44 @@
...
+};
+
+struct task_runner
+{
+    task find_task(std::string taskname);
+
+private:
+    std::set<std::string> reserved_extensions_{".json", ".md"};
+    // Hard-code extensions as executable on Windows. On non-Windows, we still rely on permissions and #!
+    std::map<std::string, std::function<task(std::string)>> builtin_tasks_{
```
mruzicka: I wish there was a name for this field which would better correspond to its actual purpose, but I could not come up with anything which would be closer to the purpose yet short ...

nicklewis: I agree, this name was confusing when I first read it. Maybe `builtin_task_wrappers` or something?

MikaelSmith: Seems reasonable.

PR #611, opened 2017-07-15
(PCP-758) Add tests for shell, powershell, ruby, and puppet tasks

MikaelSmith:
Adds acceptance tests that exercise a basic shell task, a powershell
task on Windows, and ruby and puppet tasks.

Refactors code to check puppet results for re-use in run_task.

Task runner now accepts a `trace` argument to aid in debugging. Passing
it will cause logging to be configured to the `trace` level, so issues
can be dubugged locally using `cat input.json | task trace`.

### DIFF
--- a/acceptance/lib/pxp-agent/task_helper.rb
+++ b/acceptance/lib/pxp-agent/task_helper.rb
@@ -1,0 +1,46 @@
+require 'pxp-agent/test_helper.rb'
+require 'json'
+
+# Runs a task on targets, and passes the output to the block for validation
+# Block should expect a map parsed from the JSON output
+def run_task(broker, targets, task, input = {}, max_retries = 30, query_interval = 1, &block)
+  target_identities = targets.map {|agent| "pcp://#{agent}/agent"}
+
+  provisional_responses =
+    begin
+      rpc_non_blocking_request(broker, target_identities,
+                               'task', 'run',
+                               {:task => task, :input => input})
+    rescue => exception
+      fail("Exception occurred when trying to run task '#{task}' on all agents: #{exception.message}")
+    end
+
+  transaction_ids = target_identities.map do |identity|
+    assert_equal("http://puppetlabs.com/rpc_provisional_response",
+                 provisional_responses[identity][:envelope][:message_type],
+                 "Did not receive expected rpc_provisional_response in reply to non-blocking request")
+    provisional_responses[identity][:data]["transaction_id"]
+  end
+
+  target_identities.zip(transaction_ids).map do |identity, transaction_id|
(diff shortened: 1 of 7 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -520,25 +519,34 @@ def check_puppet_non_blocking_response(identity, transaction_id, max_retries, qu
         # allow a few unknowns before giving up
         unknown_count += 1
       else
-        puppet_run_result = JSON.parse(action_result['stdout'])['status']
-        puppet_run_environment = JSON.parse(action_result['stdout'])['environment']
+        run_result = JSON.parse(action_result['stdout'])
```
zreichert: Is there a case where the `action_result` would not have the key `'stdout'`?

acceptance/lib/pxp-agent/task_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  task_path = "#{tasks_path}/#{mod}/tasks"
+  on agent, "mkdir -p #{task_path}"
+
+  create_remote_file(agent, "#{task_path}/#{task}", body)
+  on agent, "chmod +x #{task_path}/#{task}"
+
+  teardown do
+    on agent, "rm -r #{task_path}"
```
MikaelSmith: Sure, I can use force.

acceptance/tests/tasks/run_echo.rb
```diff
@@ -0,0 +1,35 @@
...
+      else
+        task_name = 'init'
+        task_body = "#!/bin/sh\necho $PT_message"
+      end
+
+      create_task_on(agent, 'echo', task_name, task_body)
+    end
+  end
+
+  step 'Send an rpc_non_blocking_request to all agents' do
```
MikaelSmith: Ah, sure.

acceptance/lib/pxp-agent/test_helper.rb
MikaelSmith: Only if there was an error. I can add a test for it. I'm not really doing error testing on responses here, but unit tests cover them fairly well.

zreichert: Just curious

MikaelSmith: Oh, we do expect responses without the stdout key. Those are generally either error_message or a response that says the task is still running. We loop until we get a response with results and status of finished, or until we time out.

MikaelSmith:
We don't yet support metadata parameters. Pushed up fixes for the comments.

PR #614, opened 2017-07-26
(maint) Update puppet task test for hardware platforms

MikaelSmith:
AIX and Solaris 11 have similar shebang behavior to el-5 and Solaris 10.
Update test special casing for running a puppet manifest to include
them.

[skip ci]


Magisus:
I verified that this fixed the problem on both Solaris and AIX.

PR #615, opened 2017-08-01
(PCP-768) Add master-uris config variable

MikaelSmith:
Adds a master-uris config variable that takes an array of HTTPS uris.
These will be used for downloading tasks.


README.md
```diff
@@ -320,6 +320,13 @@ to connect to one it will try the next in the list, and repeat until a
 successful connection is made. In the event of a disconnect, the agent will
 retry that connection before trying a new broker.
 
+**master-uris**
```
adreyer: Does this need to be optional for backwards compatibility?
Should pxp agent still allow puppet runs when it's misconfigured so there is a good way to fix it remotely?

MikaelSmith: It should and is optional.

PR #616, opened 2017-08-01
(PCP-767) Create internal "task" module

MikaelSmith:
Subclasses external_module to create an internal task module. It
currently treats `task` as an external module, with the ability to run
code before running the external module.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -8,6 +8,8 @@ namespace Modules {
 
 namespace fs = boost::filesystem;
 
+// TODO: construct and pass metadata to a new ExternalModule constructor, rather than calling `task metadata`
+// TODO: how do we get the path to task.exe if it's co-located with pxp-agent, rather than in modules dir?
 Task::Task(const fs::path& modules_dir, const fs::path& spool_dir)
         : ExternalModule { (modules_dir/"task").string(), spool_dir.string() }
 {
@@ -18,6 +20,10 @@ ExternalModule::execArgs Task::getExecArgs(std::string action)
     return { path_, { action } };
 }
 
+// TODO: move environment generation for input to a new virtual method?
+
+// TODO: move local task lookup from task.cc to here
+//       handle returning an error if no module is found
 void Task::prepareAction(const ActionRequest& request)
 {
 }
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -1,0 +1,26 @@
+#include <pxp-agent/modules/task.hpp>
+
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.modules.task"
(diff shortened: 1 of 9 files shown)


MikaelSmith:
I've only done the easy parts so far. Can either continue in the morning, or hand it off.

lib/src/request_processor.cc
```diff
@@ -184,6 +185,8 @@ RequestProcessor::RequestProcessor(std::shared_ptr<PXPConnector> connector_ptr,
     loadInternalModules();
 
     if (!agent_configuration.modules_dir.empty()) {
+        // Tasks currently uses the modules_dir to avoid having to know pxp-agent's install location.
+        modules_["task"] = std::make_shared<Modules::Task>(agent_configuration.modules_dir, spool_dir_path_);
```
MikaelSmith: This seems to break a bunch unit tests because it can't find the `task` binary. Moving metadata into the internal module would address this, but would not enable unit testing the task module.

lib/src/modules/task.cc
```diff
@@ -0,0 +1,32 @@
...
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.modules.task"
+#include <leatherman/logging/logging.hpp>
+
+namespace PXPAgent {
+namespace Modules {
+
+namespace fs = boost::filesystem;
+
+// TODO: construct and pass metadata to a new ExternalModule constructor, rather than calling `task metadata`
+// TODO: how do we get the path to task.exe if it's co-located with pxp-agent, rather than in modules dir?
```
MikaelSmith: We probably need to do this to avoid the external module loader overwriting the module registration. It registers internal modules first, then external modules.

lib/src/modules/task.cc
```diff
@@ -0,0 +1,32 @@
...
+
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.modules.task"
+#include <leatherman/logging/logging.hpp>
+
+namespace PXPAgent {
+namespace Modules {
+
+namespace fs = boost::filesystem;
+
+// TODO: construct and pass metadata to a new ExternalModule constructor, rather than calling `task metadata`
```
MikaelSmith: I propose this, rather than overriding getModuleMetadata, because override functions aren't called during base class construction. So it didn't work.

lib/src/modules/task.cc
MikaelSmith: It might be easier to put it in the modules dir anyway, and instantiate it after other external modules. It'd get skipped if it doesn't respond to metadata. Finding the binary install location for pxp-agent is kind of a pain.

mruzicka:
I've submitted a PR with implementation based on the internal task: https://github.com/puppetlabs/pxp-agent/pull/619

PR #617, opened 2017-08-03
(PCP-769) Download tasks from master-uris

ekinanp:
This commit is a template for the logic described in PCP-769. I left a few questions about this ticket under some of the TODOs. Note that a few of the other TODOs are there just to make it easier for me to see the parts I added, they will be removed once an actual PR is ready.

I've tested out the calculate_sha256 and the download_task_file routines by running small programs that use them, so I know those are correct. I will start adding unit tests to this PR after I get some feedback. Also, note that the existing unit tests all pass.


exe/task.cc
```diff
@@ -198,10 +364,19 @@ int MAIN_IMPL(int argc, char** argv)
...
-
+ 
     try {
         task_runner runner;
-        auto task = runner.find_task(taskname);
+        task task;
+        if (!args.includes("files")) {
+            task = runner.find_task(taskname);
+        } else {
+            // TODO: Do I need to check if files has at least one element?
```
MikaelSmith: We'll need to check that somewhere, so we don't crash if it doesn't. Probably a PXP error again if it doesn't.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+// too for a higher level, more flexible interface if necessary.
+/*static*/ string calculate_sha256(const string &path) {
+    const streamsize CHUNK_SIZE = 2 << 14;
+    char buffer[CHUNK_SIZE];
+    unsigned char md[SHA256_DIGEST_LENGTH];
+  
+    SHA256_CTX ctx;
+    SHA256_Init(&ctx);
+    boost::nowide::ifstream ifs(path);
+    while (!ifs.eof()) {
```
MikaelSmith: Usually worth using `while (ifs)`, then check that `ifs.eof()` afterwards. It can also be unreadable for an error, and you probably want to handle that.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+    }
+    return string(md_ascii);
+}
+
+// TODO: We don't have access to the agent config. class here, so for now, we will
+// hard-code the master-uris. From PR #615, this will be a vector of strings that 
+// will be accessed via the get_master_uris method..
+/*static*/ vector<string> get_master_uris() {
+  // TODO: Make this a static file-hosting site for testing.
+    return { "https://dummy-master.com" };
```
MikaelSmith: Depends on #615.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+}
+
+// Callback for libcurl. First param. is the chunk of data that's been downloaded, next
+// two params are the size of a single data element and the # of data elements
+// respectively, while the last param. is the stream that the data will be written
+// to. See https://curl.haxx.se/libcurl/c/libcurl-tutorial.html for more details.
+// Note that the maximum size of the data in "ptr" is 16K.
+//
+// TODO: It is possible to modify this callback so that the SHA256 of the downloaded
```
MikaelSmith: This seems useful.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+//    (2) If the downloaded file's sha does not match the provided sha, then a PXP
+//        error is returned.
+//
+//    (3) If (1) and (2) both succeed, then the downloaded file is atomically
+//        renamed to tasks_cache_dir/<filename>
+//
+// The method returns the absolute path of the file. <-- The return value can be
+// removed, I put it there for convenience w/ MAIN_IMPL
+//
+// TODO: There is an inconsistency in orchestrator and the spec. of PCP-769. Orchestrator
```
MikaelSmith: That's true, I'll correct the Orchestrator implementation.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+    auto filename = file_obj.get<string>("filename");
+    auto sha256 = file_obj.get<string>("sha256");
+    auto filepath = tasks_cache_dir / filename;
+    // TODO: Also check permissions? Make sure it's executable and readable/writable by
+    // PXP-agent's owner/group
+    if (fs::exists(filepath) && sha256 == calculate_sha256(filepath.string())) {
+      return filepath.string();
+    }
+  
+    // TODO: Not sure how I'd use code-id.
```
MikaelSmith: You shouldn't need to use code-id at all. You should append all entries in "params" as query params on the URL. You'll have to URI escape them as well. https://curl.haxx.se/libcurl/c/curl_easy_escape.html

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+    if (fs::exists(filepath) && sha256 == calculate_sha256(filepath.string())) {
+      return filepath.string();
+    }
+  
+    // TODO: Not sure how I'd use code-id. 
+    // Would it be /etc/puppetlabs/code/<code-id OR environment>/<path-uri>?
+    // Not sure if URL is right here.
+    auto path_uri = file_obj.get<string>({"uri", "path"});
+    auto environment = file_obj.get<string>({"uri", "params", "environment"});
+    auto url = get_master_uris().front() + "/etc/puppetlabs/code/" + environment + "/" + path_uri;
```
MikaelSmith: This is definitely not the URL. The URL should only include get_master_uris, the path_uri, and query params.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+    // TODO: Also check permissions? Make sure it's executable and readable/writable by
+    // PXP-agent's owner/group
+    if (fs::exists(filepath) && sha256 == calculate_sha256(filepath.string())) {
+      return filepath.string();
+    }
+  
+    // TODO: Not sure how I'd use code-id. 
+    // Would it be /etc/puppetlabs/code/<code-id OR environment>/<path-uri>?
+    // Not sure if URL is right here.
+    auto path_uri = file_obj.get<string>({"uri", "path"});
```
MikaelSmith: You likely need some try/catch (or error_code) handling for all the boost::filesystem operations.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+      // TODO: Throw exception here containing the error message specified in "e". Could be
+      // directory was not created, its last modified time was not set, or the permissions
+      // setting failed. Should this be a PXP-error?
+    }
+    return tasks_cache_dir;
+}
+
+// Computes the sha256 of the file denoted by path. Assumes that
+// the file designated by "path" exists. Can switch to openssl/evp,
+// as described on https://www.openssl.org/docs/man1.1.0/crypto/EVP_DigestInit.html
```
ekinanp: Yeah, that's the only thing. Is it worth making the switch?

MikaelSmith: Seems like an easy change, so sure.

exe/task.cc
```diff
@@ -137,6 +148,161 @@ static map<string, string> generate_environment_from(lth_jc::JsonContainer &inpu
...
+// Creates the SYSTEM_PREFIX/pxp-agent/tasks-cache/<sha> directory,
+// and ensures that its permissions are only readable/writable by
+// the PXP agent owner/group. Returns the path to this directory.
+// Note that the last modified time of the directory is updated.
+/*static*/ fs::path create_tasks_cache_dir(const string &sha256) {
+    auto tasks_cache_dir = fs::path(SYSTEM_PREFIX()) / "pxp-agent" / "tasks-cache" / sha256;
+    try {
+      fs::create_directories(tasks_cache_dir);
+      fs::last_write_time(tasks_cache_dir, time(nullptr)); 
+      fs::permissions(tasks_cache_dir, TASK_FILE_RW_PERMS);
```
Iristyle: If we can get `tasks-cache` added to https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md that would be awesome.

Then we can get a good idea of how the parent directories are created / what perms are laid down.  That in turn will let us know if we can just inherit perms from parent directories or not.

MikaelSmith: https://github.com/puppetlabs/puppet-specifications/pull/97

MikaelSmith: It strikes me that we should be able to override this with a config option, as we allow with `spool-dir`.

ekinanp:
@MikaelSmith Put up an updated PR.  Main changes are:

(1)  Switching to EVP when calculating the SHA-256 of a file
(2) Generating the correct URL for libcurl to download from
(3) Returning _error objects where appropriate (in place of the original TODOs)
(4) Abstracting out filesystem_error objects and converting them to task_error objects instead (WIP, but would avoid a lot of try/catch clutter).

I have tested out each of the routines using small programs on my own except for update_task_file, and they work. I also have a test plan sketched out to add to the existing unit tests. Don't know if I can get all of the tests done tomorrow, but I'll try. These are the test cases I'm considering (rough outline):

(1) For a "FILES" entry w/ 0 elements, task_error exception is thrown.

(2) (This will be a part of the "happy case" test) Ensure that the "pxp-agent/tasks-cache/sha256" directory is created with the right permissions (read/write owner and group) and the last modified date (within a few seconds of the current time).

(3) If the task-file exists and its sha matches the provided sha, then ensure that (a) Nothing is downloaded using libcurl and (b) The task is executed.

(4) Ensure that when an error in the sha-computation occurs (i.e. "ifs.fail()" is triggered), the appropriate task_error object is thrown.

(5) Ensure that when a download error occurs, that the failed-task-download error object is returned, and the temp file is removed.

(6) Ensure that after download, if the downloaded sha's file does not match the provided sha, that the temp file is removed and the 'task-version-conflict' task error object is thrown.

(7) (Happy case) Task does not exist in tasks-cache directory, so it is downloaded, and executed correctly w/ the right permissions (group/owner execute, read and write)

(8) When a filesystem_error is thrown, it is converted to the expected task_error object.

Do these seem reasonable? Please let me know if you have any suggestions/fixes.

MikaelSmith:
I'd also take a look at the Travis failures, they're noting a few things: one catch block masking another, and some linting issues.

exe/task.cc
MikaelSmith: Inheritance does sound right. The right thing to do would then be to have packaging create the pxp-agent/spool and pxp-agent/tasks directories, and enforce more restricted permissions on them?

ekinanp:
Ok, I have tests for Cases (1), (3), (5) and (8) above. No obvious way to test (4) off the top of my head, at least not without having to declare the sha function in a header file. Cases (6), (7) and (2) require me to host some files on a static website, it was mentioned that we'd do this on Jetty?

MikaelSmith:
Hosting files would be using a static file host configured in Jetty as part of acceptance tests. Notes in PCP-771.

ekinanp:
Oh so does that mean I don't have to worry about those cases?

MikaelSmith:
Maybe highlight them explicitly in PCP-771.

MikaelSmith:
Also, CI's failing.

MikaelSmith:
I'll defer to @mruzicka on reviewing this if he's around next week.

ekinanp:
I'll look at the CI stuff now, wanted to put up something for you to look at before you left (hope you enjoy your PTO by the way).

MikaelSmith:
We'll want to rebase this on https://github.com/puppetlabs/pxp-agent/pull/619.

MikaelSmith:
Changes merged, should be able to rebase on master now.

MikaelSmith:
Looks like you're rebase wasn't resolved correctly. You might just try resetting to master, then cherry-pick your changes on top.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,163 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+      updateUrl(url, curl, params.get<std::string>(key));
+      url += "&";
+    }
+    return url;
+}
+
+// This method downloads the file at the specified url into the provided path
+// If something goes wrong during the download, a PXP-error should be returned.
+static void downloadTaskFile(const fs::path& path, const lth_jc::JsonContainer& uri) {
+    auto curl = curl_easy_init();
```
MikaelSmith: Would using https://github.com/puppetlabs/leatherman/blob/master/curl/inc/leatherman/curl/client.hpp simplify this?

ekinanp: Unfortunately I don't think it would, because that is meant for HTTP requests/responses. For our case, we can get a response but we would still have to write it to an external file anyways (plus responses can be huge, whereas the above code downloads the file in chunks). However, it wouldn't be a bad idea to extend some of the logic in that class to include a helper for downloading files, in the future (e.g. add a method download_file).

ekinanp:
Ok, I'm about to start local testing. Just updated the state of the PR for review.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
     return TaskCommand { task_executable.string(), { } };
 }
 
+// Converts boost filesystem errors to a task_error object.
+// TODO: Modify this to convert it to a PXP-error instead.
```
MikaelSmith: `response.setBadResultsAndEnd(execution_error);` is generally how we setup to return a PXP error. See https://github.com/puppetlabs/pxp-agent/blob/master/lib/src/request_processor.cc#L126-L152 for where the PXP error is sent.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    if (alg::contains(err_msg, "No such file or directory")) {
+      return Module::ProcessingError(lth_loc::format("No such file or directory: {1}", e.path1()));
+    } else if (alg::contains(err_msg, "File exists")) {
+      return Module::ProcessingError(lth_loc::format("A file matching the name of the provided sha already exists"));
+    }
+    return Module::ProcessingError(err_msg);
+}
+
+// Take out common permissions.
+static fs::perms TASK_FILE_RW_PERMS = fs::owner_read | fs::owner_write | fs::group_read | fs::group_write;
```
MikaelSmith: I'd make this owner_read/write, group_read. Drop group_write.

lib/src/modules/task.cc
ekinanp: I did that because executeAction will catch the exception and wrap it up in a PXP error: https://github.com/puppetlabs/pxp-agent/blob/master/lib/src/module.cc#L97. Should I include a separate catch block instead for filesystem errors?

MikaelSmith: Ah, that seems ok. You can probably remove the TODO then.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+
+// Creates the <task_cache_dir>/<sha256> directory for a single
+// task and ensures that its permissions are only readable/writable by
+// the PXP agent owner/group. Returns the path to this directory.
+// Note that the last modified time of the directory is updated.
+static fs::path createCacheDir(const fs::path& task_cache_dir, const std::string& sha256) {
+    auto cache_dir = fs::path(task_cache_dir) / sha256;
+    fs::create_directory(cache_dir);
+    fs::last_write_time(cache_dir, time(nullptr));
+    fs::permissions(cache_dir, TASK_FILE_RW_PERMS);
```
MikaelSmith: For a directory, you'll want to include the executable bit.

MikaelSmith: https://github.com/puppetlabs/pxp-agent/pull/623 has some examples for spool-dir and logfile.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    auto filepath = cache_dir / filename;
+    // TODO: Also check permissions? Make sure it's readable/writable/executable by
+    // PXP-agent's owner/group?
+    if (fs::exists(filepath) && sha256 == calculateSha256(filepath.string())) {
+      return filepath;
+    }
+
+    // TODO: Maybe place somewhere else? A task module without any master uris
+    // would need to have all its files hardcoded in its respective <sha> directory.
+    if (master_uris.empty()) {
```
MikaelSmith: This seems reasonable to do only if we actually need to download a file. That appears to be what you've done.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+// This method does the following. If the file matching the "filename" field of the
+// file_obj JSON does not exist OR if its hash does not match the sha value in the
+// "sha256" field of file_obj, then:
+//    (1) The file is downloaded using libcurl from the first entry in master-uris
+//        to a temporary directory. If this download fails, a PXP error is thrown.
+//
+//    (2) If the downloaded file's sha does not match the provided sha, then a PXP
+//        error is returned.
+//
+//    (3) If (1) and (2) both succeed, then the downloaded file is atomically
```
MikaelSmith: Should be atomically renamed to `cache_dir/<sha>/<filename>`.

MikaelSmith: Oh nevermind, you've defined cache_dir to include the sha.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    } else if (alg::contains(err_msg, "File exists")) {
+      return Module::ProcessingError(lth_loc::format("A file matching the name of the provided sha already exists"));
+    }
+    return Module::ProcessingError(err_msg);
+}
+
+// Take out common permissions.
+static fs::perms TASK_FILE_RW_PERMS = fs::owner_read | fs::owner_write | fs::group_read | fs::group_write;
+
+// Creates the <task_cache_dir>/<sha256> directory for a single
```
MikaelSmith: I'd note it will not fail if the directory already exists.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+//
+//    (3) If (1) and (2) both succeed, then the downloaded file is atomically
+//        renamed to cache_dir/<filename>
+static fs::path updateTaskFile(const std::vector<std::string>& master_uris,
+                               const fs::path& cache_dir,
+                               const lth_jc::JsonContainer& file) {
+    auto filename = file.get<std::string>("filename");
+    auto sha256 = file.get<std::string>("sha256");
+    auto filepath = cache_dir / filename;
+    // TODO: Also check permissions? Make sure it's readable/writable/executable by
```
MikaelSmith: That seems reasonable to enforce. We can relax it later if there's a reason, but making sure it's executable may avoid problems later.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    // PXP-agent's owner/group?
+    if (fs::exists(filepath) && sha256 == calculateSha256(filepath.string())) {
+      return filepath;
```
MikaelSmith: Indentation is inconsistent.

lib/src/modules/task.cc
```diff
@@ -128,11 +136,171 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+// If something goes wrong during the download, a PXP-error should be returned.
+static void downloadTaskFile(const std::vector<std::string>& master_uris,
+                             const fs::path& path,
+                             const lth_jc::JsonContainer& uri) {
+    auto curl = curl_easy_init();
+    if (!curl) {
+      throw Module::ProcessingError(lth_loc::format("Failed to initiate the task file download process"));
+    }
+    char err_buff[CURL_ERROR_SIZE];
+    auto fp = fopen(path.string().c_str(), "wb");
```
MikaelSmith: fopen might be a problem with unicode paths on Windows. Boost.Nowide should provide a UTF-8 capable version of this.

MikaelSmith: https://github.com/artyom-beilis/nowide/blob/master/include/boost/nowide/cstdio.hpp

lib/src/modules/task.cc
```diff
@@ -6,16 +6,22 @@
...
+#include <boost/algorithm/string/predicate.hpp>
 
 #include <rapidjson/rapidjson.h>
 
 #define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.modules.task"
 #include <leatherman/logging/logging.hpp>
 
+#include <openssl/evp.h>
+#include <curl/curl.h>
+#include <stdio.h>
```
MikaelSmith: Prefer `boost/nowide/cstdio` instead.

ekinanp:
I made some small updates to this PR. Main thing is inserting the SSL setup (note unlike leatherman, I don't check the return of easy_setopt and error if a failure occurs, as the examples I found online (e.g. the tutorial, https://curl.haxx.se/libcurl/c/simplessl.html) don't seem to do so, so I think we're safe.

The task configuration is getting rather large (6 items now). I am thinking it might be a good idea to create a separate TaskConfiguration struct, either in the task.hpp file or in configuration.hpp, as Configuration::Task. What do you think?

MikaelSmith:
None of the config is optional, so not sure there's much point in a separate struct.

ekinanp:
It would reduce the # of arguments to the helper routines for finding the task file (getCachedTask, updateTaskFile, downloadTaskFile, etc.). Maybe a struct inside task.hpp only (task_config)? The constructor for the Task module would take either the relevant fields from the Agent configuration or the entire agent configuration, and set the task_config fields using that.

MikaelSmith:
Ah, that's a reasonable argument for it.

lib/src/modules/task.cc
```diff
@@ -128,11 +147,189 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+                             const std::string& crt,
+                             const std::string& key,
+                             const fs::path& path,
+                             const lth_jc::JsonContainer& uri) {
+    auto curl = curl_easy_init();
+    if (!curl) {
+      throw Module::ProcessingError(lth_loc::format("Failed to initiate the task file download process"));
+    }
+    char err_buff[CURL_ERROR_SIZE];
+    auto fp = boost::nowide::fopen(path.string().c_str(), "wb");
```
MikaelSmith: This could fail, we should check that it succeeded.

MikaelSmith:
At this point, I'm ok getting this working independent of changes to Leatherman. We can refactor to use improved Leatherman utilities separately.

lib/src/modules/task.cc
```diff
@@ -128,11 +149,145 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
     return TaskCommand { task_executable.string(), { } };
 }
 
+// Converts boost filesystem errors to a task_error object.
+static Module::ProcessingError toModuleProcessingError(const fs::filesystem_error& e) {
+    std::string err_msg = e.what();
+    if (alg::contains(err_msg, "No such file or directory")) {
```
mruzicka: Could we use `e.code()` defined in the [system_error class](http://www.boost.org/doc/libs/1_64_0/libs/system/doc/reference.html#Class-system_error) rather than matching against the message text?

lib/src/modules/task.cc
```diff
@@ -128,11 +149,145 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+      return Module::ProcessingError(lth_loc::format("No such file or directory: {1}", e.path1()));
+    } else if (alg::contains(err_msg, "File exists")) {
+      return Module::ProcessingError(lth_loc::format("A file matching the name of the provided sha already exists"));
```
mruzicka: Again I'd prefer to use `e.code()`.

lib/src/modules/task.cc
```diff
@@ -128,11 +149,145 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+static fs::perms TASK_FILE_PERMS = fs::owner_read | fs::owner_write | fs::group_read;
+
+// Creates the <task_cache_dir>/<sha256> directory for a single
+// task, ensuring that its permissions are readable by
+// the PXP agent owner/group (for unix OSes), writable for the PXP agent owner,
+// and executable by both PXP agent owner and group. Returns the path to this directory.
+// Note that the last modified time of the directory is updated, and that this routine
+// will not fail if the directory already exists.
+static fs::path createCacheDir(const fs::path& task_cache_dir, const std::string& sha256) {
+    auto cache_dir = fs::path(task_cache_dir) / sha256;
```
mruzicka: I believe the `fs::path()` construction can be omitted here. Shortening the expression to:
```c++
auto cache_dir = task_cache_dir / sha256;
```

lib/src/modules/task.cc
```diff
@@ -128,11 +149,145 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    const std::streamsize CHUNK_SIZE = 2 << 14;
+    char buffer[CHUNK_SIZE];
+    unsigned char md_value[EVP_MAX_MD_SIZE];
+    unsigned int md_len;
+
+    auto md = EVP_sha256();
+    auto mdctx = EVP_MD_CTX_create();
+    EVP_DigestInit_ex(mdctx, md, nullptr);
+    boost::nowide::ifstream ifs(path);
+    while (ifs.read(buffer, CHUNK_SIZE)) {
```
mruzicka: If this read() method is the one described in [this documentation](http://en.cppreference.com/w/cpp/io/basic_istream/read) then this is an infinite loop.

mruzicka: I spoke too soon, apparently there is the implicit `(void *)` conversion in play here.

lib/src/modules/task.cc
```diff
@@ -128,11 +149,145 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+      EVP_DigestUpdate(mdctx, buffer, CHUNK_SIZE);
```
mruzicka: The `CHUNK_SIZE` needs to be replaced with [`ifs.gcount()`](http://en.cppreference.com/w/cpp/io/basic_istream/gcount) otherwise if the last chunk is not exactly CHUNK_SIZE long the checksum computation will process garbage (actually the end of the las but one chunk) at the end of the buffer.

mruzicka: Again too soon, now I see it's done below.

lib/src/modules/task.cc
```diff
@@ -128,11 +152,146 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
     return TaskCommand { task_executable.string(), { } };
 }
 
+// Converts boost filesystem errors to a task_error object.
+static Module::ProcessingError toModuleProcessingError(const fs::filesystem_error& e) {
+    auto err_code = e.code();
```
ekinanp: @mruzicka Is this what you had in mind?

ekinanp:
Okay quick update. Current form of this PR passes local testing. That is,
   (1) If the task file exists and it matches the provided sha, then the task is run w/ no download.
   (2) Otherwise the task is downloaded from the first master-uri w/ the right URL. If the shas of the downloaded file and the provided sha match, the task is set to rwx for user, rx for group and successfully ran. Otherwise a version conflict error is thrown and the temporary task file is removed.

The next step is to properly account for permissions. I think this might be better done after #623 is merged since the permissions defined there could be re-used here.

lib/src/modules/task.cc
MikaelSmith: Not void*. It returns an istream, which has a boolean() overload that returns true only if the stream is ready to have "read" called on it.

lib/src/modules/task.cc
```diff
@@ -108,6 +123,15 @@ Task::Task(const fs::path& exec_prefix,
 
     input_validator_.registerSchema(input_schema);
     results_validator_.registerSchema(output_schema);
+
+    // TODO: Remove this. Only put it here to test out the SSL stuff.
+    // ca_ = "/Users/enis.inan/GitHub/orchestrator/dev-resources/ssl/ca.pem";
```
MikaelSmith: Why testing this way, instead of creating a config file?

lib/src/modules/task.cc
MikaelSmith: That looks much better.

lib/src/modules/task.cc
ekinanp: Didn't think of it, the above was just quick and easy at the time. Thanks for the suggestion!

lib/src/modules/task.cc
```diff
@@ -128,11 +152,146 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+                             const fs::path& file_path,
+                             const lth_jc::JsonContainer& uri) {    
+    auto url = master_uris[0] + createUrlEndpoint(uri);
+    lth_curl::request req(url);
+    // timeout from connection after one minute, can configure 
+    req.connection_timeout(60000);
+    try {
+        client.download_file(req, file_path.string());
+    } catch (lth_curl::http_file_download_exception& e) {
+        // TODO: There's a chance that curl's temp file (not the temp. task file) might not be removed here.
```
MikaelSmith: I'm not sure I understand. I only see 1 temp file, the one Leatherman.Curl creates when downloading the file.

ekinanp: The way I was thinking with that one is that we have one "temp" file that we save the downloaded task to ("tempA") which is what's passed in to the downloadTask method. When Leatherman.curl downloads the task, it will create another temp "tempB" to write its contents in chunks and then save it to "tempA" after the download is complete. That's where the two temp files come from. I think it isn't a bad idea to have two, because once we have "tempA" we also want to check that the passed-in sha matches "tempA"'s sha for version control.

MikaelSmith: Ah, we've introduced two temp names. That's probably not bad. It's a little funny, but it somewhat simplifies error handling if multiple threads download the same file. The alternative would be download to the final location, and if it fails the SHA test delete it (accepting delete failure if the file is missing because another thread already deleted it). This solution seems ok.

MikaelSmith: I'd suggest adding some of that rationale to the code comments.

lib/src/modules/task.cc
```diff
@@ -128,11 +152,146 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+        return Module::ProcessingError(lth_loc::format("No such file or directory: {1}", e.path1()));
+    } else if (err_code ==  boost_error::file_exists) {
+        return Module::ProcessingError(lth_loc::format("A file matching the name of the provided sha already exists"));
+    } else {
+        return Module::ProcessingError(e.what());
+    }
+}
+
+// Take out common permissions.
+static fs::perms TASK_FILE_PERMS = fs::owner_read | fs::owner_write | fs::group_read;
```
MikaelSmith: I think we'll only want to use these on Linux. On Windows they don't make a lot of sense.

ekinanp: Do you want me to go ahead and add these for Linux and ignore them for Windows? I was waiting on #623 so that I could reuse the stuff that's there, e.g. the permissions defined in configuration.hpp.

lib/src/modules/task.cc
```diff
@@ -128,11 +152,146 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+//    (3) If (1) and (2) both succeed, then the downloaded file is atomically
+//        renamed to cache_dir/<filename>
+static fs::path updateTaskFile(const std::vector<std::string>& master_uris,
+                               lth_curl::client& client,
+                               const fs::path& cache_dir,
+                               const lth_jc::JsonContainer& file) {
+    auto filename = file.get<std::string>("filename");
+    auto sha256 = file.get<std::string>("sha256");
+    auto filepath = cache_dir / filename;
+    // TODO: Check permissions here, enforce executable part if linux!
```
MikaelSmith: I'd just re-apply the permissions we expect. Should be a no-op if they match.

lib/src/modules/task.cc
ekinanp: Am doing that now. What do you think of the periodic GC comment in the TODO, in case the removals both fail? Should I leave that there?

lib/src/modules/task.cc
MikaelSmith: Can wait on #623. Hopefully we can figure that out tomorrow.

lib/src/modules/task.cc
MikaelSmith: Why would curl's temp file not be removed? I think we handled all the scenarios where it needs to be. I'd rather not have a separate GC just in case we left a bug in Leatherman.Curl.

ekinanp: Well, one thought that comes to mind is some accidental change of permissions. Though that's unlikely, but the possibility is still there. In Leatherman.curl when the download files, we call removal only once and if that fails, then we pass in the path of the temp file in our http_file_download_exception.

Kind of an edge case, but I thought I'd bring it up anyways.

lib/src/modules/task.cc
ekinanp: Ah okay, I'll do that after #623. Thanks for the suggestion!

lib/src/modules/task.cc
```diff
@@ -128,11 +147,150 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+        auto escaped_key = std::string(
+           lth_curl::curl_escaped_string(curl_handle, key));
+        auto escaped_val = std::string(
+           lth_curl::curl_escaped_string(curl_handle, params.get<std::string>(key)));
+        url += escaped_key + "=" + escaped_val + "&";
+    }
+    return url;
+}
+
+// Downloads the file at the specified url into the provided path. If something goes
```
ekinanp: @MikaelSmith How is this for the documentation of the two temp-files case? I also refactored downloadTaskFile to remove the try catch block; the catch block for http_file_download_exception is in getCachedTask now.

MikaelSmith: Sure.

lib/src/modules/task.cc
mruzicka: @enis, yes, that's exactly it.

lib/src/modules/task.cc
mruzicka: The operator bool overload according to [this doc](http://en.cppreference.com/w/cpp/io/basic_ios/operator_bool) is marked as `explicit` which I thought would prohibit its use without an explicit `(bool)` cast?

MikaelSmith: Use in `while` might be considered an explicit cast? I'm not actually sure.

MikaelSmith:
https://github.com/puppetlabs/pxp-agent/pull/627 should fix the lint issues on master.

lib/src/modules/task.cc
MikaelSmith: Actually, ignore my comment about Windows. Easier to just apply it everywhere (and understand that it'll be a noop on Windows).

ekinanp: Okay, will do. Saw that #623 just got merged, so I'll start incorporating the changes there.

ekinanp:
@MikaelSmith @mruzicka I incorporated the permissions from #623 and also modified configuration.cc to point to the tasks-cache directory.

MikaelSmith:
Merged https://github.com/puppetlabs/pxp-agent/pull/627, should be able to rebase on it.

lib/src/configuration.cc
```diff
@@ -84,17 +84,15 @@ namespace lth_util = leatherman::util;
         }
     }();
 
-    // TODO(task) switch back to tasks-cache when downloading is implemented.
-    static const std::string DEFAULT_TASK_CACHE_DIR { (DATA_DIR / "tasks").string() };
+    static const std::string DEFAULT_TASK_CACHE_DIR { (DATA_DIR / "tasks-cache").string() };
```
MikaelSmith: I'd suggest making task_cache_dir configurable so we can run as non-root and override for local testing.

MikaelSmith: Also, going to rely on https://github.com/puppetlabs/puppet-agent/pull/1135.

ekinanp:
Just rebased, fixing some CI issues now then I'll change the commit message to something more descriptive.

lib/src/modules/task.cc
```diff
@@ -108,6 +123,10 @@ Task::Task(const fs::path& exec_prefix,
 
     input_validator_.registerSchema(input_schema);
     results_validator_.registerSchema(output_schema);
+
+    client_.set_ca_cert(ca);
```
MikaelSmith: I get an error trying to use this: `Failed to execute the task for the non blocking 'task run' request (transaction 142). Error: Downloading the task file init failed. Reason: SSL: Can't load the certificate "~/puppetlabs/orchestrator/dev-resources/ssl/cert.pem" and its private key: OSStatus -25299`.

Any ideas?

MikaelSmith: Oh, I might be picking up the wrong libcurl library.

lib/src/configuration.cc
ekinanp: Okay, working on that now

lib/src/modules/task.cc
MikaelSmith: I had to mess with how I build pxp-agent to get matching versions of OpenSSL and libcurl:
```
cmake -DCMAKE_PREFIX_PATH="/usr/local/opt/openssl@1.1;/usr/local/opt/curl" ..
```

MikaelSmith: We should add a note to the README about that.

ekinanp:
Ok I added configuration for the task-cache-dir and also a note to the README about the openssl and libcurl conflict. @MikaelSmith

README.md
```diff
@@ -114,6 +114,10 @@ Build
       cmake -DCMAKE_BUILD_TYPE=Debug -DDEV_LOG_COLOR=ON ..
       make
 
+  NOTE: If the versions of OpenSSL and libcurl conflict with each other, then invoke 
+  cmake with the following command:
+      cmake -DCMAKE_PREFIX_PATH="/usr/local/opt/openssl@1.1;/usr/local/opt/curl" ..
```
MikaelSmith: I'd update this to something like

```
NOTE: If the versions of OpenSSL and libcurl conflict with each other, curl may fail to load SSL files. On macOS this is common when using Homebrew. Invoke cmake with the following commands to use the Homebrew versions of OpenSSL and libcurl:
    cmake -DCMAKE_PREFIX_PATH="/usr/local/opt/openssl;/usr/local/opt/curl" ..
```

lib/src/configuration.cc
```diff
@@ -164,15 +162,10 @@ HW::ParseResult Configuration::parseOptions(int argc, char *argv[])
     // This is primarily for testing.
     master_uris_.clear();
 
-    // Remember the path to the pxp-agent executable used to start
+    // remember the path to the pxp-agent executable used to start
```
MikaelSmith: This looks like it rebased incorrectly.

lib/src/configuration.cc
```diff
@@ -228,7 +221,9 @@ std::string Configuration::setupLogging()
         // up logging before calling validateAndNormalizeConfiguration
         validateLogDirPath(logfile_);
         logfile_fstream_.open(logfile_.c_str(), std::ios_base::app);
+#ifndef _WIN32
```
MikaelSmith: Same here.

lib/src/modules/task.cc
```diff
@@ -128,11 +147,151 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    if (!ifs.eof()) {
+      throw Module::ProcessingError(lth_loc::format("Error while reading {1}", path));
+    }
+    EVP_DigestUpdate(mdctx, buffer, ifs.gcount());
+    EVP_DigestFinal_ex(mdctx, md_value, &md_len);
+    EVP_MD_CTX_destroy(mdctx);
+
+    char md_value_ascii[2*EVP_MAX_MD_SIZE+1] = {'\0'};
+    for (unsigned int i = 0; i < md_len; ++i) {
+      snprintf(md_value_ascii+2*i, 3, "%02x", md_value[i]);
```
MikaelSmith: There's a linter issue here. It's kind of an annoying one. It probably makes sense to disable that check, but I don't have a suggestion how yet.

ekinanp: Yeah, that's one that needs to be there unfortunately.

MikaelSmith: Honestly, I'd make the `3` a variable and call the variable. That satisfies the linter. The other things runtime/printf catches are useful.

MikaelSmith: Something like
```
+    constexpr size_t hexlen = 3;
     for (unsigned int i = 0; i < md_len; ++i) {
-      snprintf(md_value_ascii+2*i, 3, "%02x", md_value[i]);
+      snprintf(md_value_ascii+2*i, hexlen, "%02x", md_value[i]);
```

lib/tests/unit/modules/task_test.cc
```diff
@@ -206,6 +250,73 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
...
+            lth_jc::JsonContainer(task_txt),
+            {},
+            0 };
+        ActionRequest request { RequestType::Blocking, task_content };
+        auto response = e_m.executeAction(request);
+
+        REQUIRE_FALSE(response.action_metadata.includes("results"));
+        REQUIRE_FALSE(response.action_metadata.get<bool>("results_are_valid"));
+        REQUIRE(response.action_metadata.includes("execution_error"));
+        REQUIRE_THAT(response.action_metadata.get<std::string>("execution_error"), Catch::Contains("_mock_master_" + bad_path));
```
MikaelSmith: This appears to be inconsistent with different versions of curl. On my laptop it passes, in Travis it fails because the url is single-quoted. We should probably make the test simply search for the bad path in the output.

MikaelSmith: 👍

MikaelSmith:
Close on fixing AppVeyor in https://github.com/puppetlabs/pxp-agent/pull/629, but there's some issue with cpp-pcp-client and gcc 5.2 I'm still working on.

ekinanp:
Ok, I have updated my PR w/ the modified tasks acceptance tests if you want to look. Save for the windows one (which I think I'll need a separate agent build for), they all individually pass. I'm running the entire suite now.

MikaelSmith:
Ok, rebase this on master again and I think it's about ready.

acceptance/lib/pxp-agent/task_helper.rb
```diff
@@ -31,14 +31,14 @@ def run_task(broker, targets, task, filename, input = {}, max_retries = 30, quer
   end
 end
 
-def create_task_on(agent, mod, task, body)
```
MikaelSmith: Since we're passing body to the task, why not calculate SHA as part of create_task_on?

MikaelSmith: Oh, because you need it for run_task. Nevermind then.

MikaelSmith: I'd be tempted to have create_task_on return the sha...

lib/src/modules/task.cc
```diff
@@ -6,21 +6,31 @@
 #include <leatherman/file_util/file.hpp>
 
 #include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/nowide/cstdio.hpp>
+#include <boost/system/error_code.hpp>
+#include <boost/asio/error.hpp>
```
MikaelSmith: Why'd you add `boost/asio/error.hpp`?

exe/task_wrapper.cc
```diff
@@ -44,7 +45,9 @@ int main(int argc, char *argv[])
         // Avoid atomic update to allow testing against /dev/stderr. There should never be
         // multiple processes trying to write this output.
         boost::nowide::ofstream ofs { params.get<std::string>("stderr"), std::ios::binary };
+        #ifndef _WIN32
         fs::permissions(params.get<std::string>("stderr"), FILE_PERMS);
```
MikaelSmith: Also exclude at line 53/56, and line 16.

MikaelSmith: And stylistically, we tend to put the pragmas at the beginning of the line (not inline with code).

acceptance/tests/tasks/run_ps1.rb
```diff
@@ -20,15 +20,13 @@
 
   step 'Create powershell task on agent hosts' do
     windows_hosts.each do |agent|
-      create_task_on(agent, 'echo', 'init.ps1', <<-EOF)
-foreach ($i in $input) { Write-Output $i }
-Write-Output $env:PT_data
-EOF
+      task_body = "foreach ($i in $input) { Write-Output $i }\nWrite-Output $env:PT_data";
```
MikaelSmith: All of these could go back to using HEREDOC notation (assigning to @sha256 would be the only change).

lib/src/configuration.cc
```diff
@@ -911,6 +919,27 @@ void Configuration::validateAndNormalizeOtherSettings()
...
+    const auto task_cache_dir = HW::GetFlag<std::string>("task-cache-dir");
+
+    if (task_cache_dir.empty()) {
+        throw Configuration::Error {
+            lth_loc::translate("task-cache-dir must be defined") };
+    }
+
+    fs::path task_cache_dir_path { lth_file::tilde_expand(task_cache_dir) };
+    check_and_create_dir(task_cache_dir_path, "task-cache-dir", true);
+    try {
```
MikaelSmith: This should be ifdefed as well.

lib/src/modules/task.cc
```diff
@@ -128,11 +146,160 @@ static TaskCommand getTaskCommand(const fs::path &task_executable)
...
+    if (err_code == boost_error::no_such_file_or_directory) {
+        return Module::ProcessingError(lth_loc::format("No such file or directory: {1}", e.path1()));
+    } else if (err_code ==  boost_error::file_exists) {
+        return Module::ProcessingError(lth_loc::format("A file matching the name of the provided sha already exists"));
+    } else {
+        return Module::ProcessingError(e.what());
+    }
+}
+
+static const fs::perms NIX_TASK_FILE_PERMS = NIX_DIR_PERMS;
```
MikaelSmith: This needs to be ifdefed or we'll get compilation errors in puppet-agent builds. We might get it in AppVeyor too now that they're using the same GCC version.

acceptance/tests/tasks/run_ps1.rb
MikaelSmith: Nevermind, apparently that messes with the SHA calculation.

lib/tests/unit/modules/task_test.cc
```diff
@@ -206,6 +283,77 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
...
+        ActionRequest request { RequestType::Blocking, task_content };
+        auto response = e_m.executeAction(request);
+
+        REQUIRE_FALSE(response.action_metadata.includes("results"));
+        REQUIRE_FALSE(response.action_metadata.get<bool>("results_are_valid"));
+        REQUIRE(response.action_metadata.includes("execution_error"));
+        REQUIRE_THAT(response.action_metadata.get<std::string>("execution_error"), Catch::Contains("Downloading the task file some_file failed"));
+
+        // Make sure temp file is removed.
+        // TODO: Access denied error results when trying to clean out the temp directory on
```
MikaelSmith: Comment can go be removed.

lib/tests/unit/modules/task_test.cc
```diff
@@ -206,6 +283,77 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
...
+            lth_jc::JsonContainer(ENVELOPE_TXT),
+            lth_jc::JsonContainer(task_txt),
+            {},
+            0 };
+        ActionRequest request { RequestType::Blocking, task_content };
+        e_m.executeAction(request);
+
+        REQUIRE(fs::exists(cache));
+        REQUIRE(fs::is_directory(cache));
+        #ifndef _WIN32
```
MikaelSmith: This ifndef/endif should also be at the beginning of the line.

lib/src/modules/task.cc
MikaelSmith: Yup, errored.

exe/task_wrapper.cc
MikaelSmith: You can skip making changes to this file, shouldn't affect testing and it's more related to the spool-dir permissions (which I'll fix up separately).

PR #623, opened 2017-08-16
(PCP-766) Restrict permissions on spool files, logfile

MikaelSmith:
Applies permission 0640 to ensure only owner/group can read and only
owner can write sensitive spool files.

Ensure logfile is not world-readable. It may contain sensitive
information in the future.

Blocked on merge and release of https://github.com/puppetlabs/leatherman/pull/249.


lib/src/results_storage.cc
```diff
@@ -15,6 +16,8 @@
 
 #include <algorithm>  // std::find
 
+const uint16_t UMASK_FLAGS { 026 };
```
nicklewis: This should probably be `0026` for clarity.

mruzicka:
There is only a linting problem which we'll fix in a maint PR.

MikaelSmith:
Hopefully this passes CI. Leatherman 1.1.0 was released this morning.

MikaelSmith:
Failed lint check, I'll get up maint PR.

MikaelSmith:
Fixed in https://github.com/puppetlabs/pxp-agent/pull/627.

PR #625, opened 2017-08-18
(maint) Unblock acceptance tests on Japanese locale instances

MikaelSmith:
When output is UTF-8, but running on a Japanese-configured system with
Windows-31J, we get UndefinedConversionError trying to convert the
encoding to UTF-8. Handle this error.

Disable tests that fail with Japanese locales now that Puppet is
translated into Japanese. This should be reverted once PCP-776 is fixed.


MikaelSmith:
Also, https://github.com/puppetlabs/pxp-agent/blob/master/modules/pxp-module-puppet#L115-L121 is broken by translations.

PR #626, opened 2017-08-18
(PCP-776) Detect disabled/running Puppet via lock files

MikaelSmith:
`pxp-module-puppet` can't detect that a run failed - because Puppet was
disabled or because Puppet was already running - when Puppet's output is
translated. What we currently try to do is
- if Puppet's disabled, return an error message to that effect
- if Puppet's running, wait until it stops and start a new run

Switch to checking for the lockfiles that Puppet uses to determine
disabled or running, rather than parsing the output. This makes the
implementation language-agnostic.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -12,7 +12,6 @@ module Pxp
       NoPuppetBin = "no_puppet_bin"
       NoLastRunReport = "no_last_run_report"
       InvalidLastRunReport = "invalid_last_run_report"
-      AlreadyRunning = "agent_already_running"
       Disabled = "agent_disabled"
       FailedToStart = "agent_failed_to_start"
       NonZeroExit = "agent_exit_non_zero"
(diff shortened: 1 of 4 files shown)


MikaelSmith:
This has some shortcomings compared to the previous implementation. It's probably enough in most cases, and I can't think of ways to make it better without running Puppet in-memory.

modules/pxp-module-puppet
```diff
@@ -194,8 +209,8 @@ module Pxp
       return run_result
     end
 
-    # Wait for the lockfile to be removed. If it hasn't after 30 seconds, give up.
-    def wait_for_lockfile(lockfile = '', check_interval = 0.1, give_up_after = 30)
+    # Wait for the lockfile to be removed. If it hasn't after 10 minutes, give up.
```
MikaelSmith: Changes to a 10 minute limit to wait before giving up. Previously it would go until it had a successful run.

modules/pxp-module-puppet
```diff
@@ -214,58 +229,63 @@ module Pxp
...
-            return self.class.make_error_result(exitcode, Errors::AlreadyRunning,
-                                     "Puppet agent is already performing a run; it also appears to be" +
-                                     " misconfigured, agent_catalog_run_lockfile is an empty string")
+        # Check for a lockfile. If present, wait until it's removed and try running again.
+        # There's a chance that our run finished with a real error rather than because Puppet was
+        # already running, but another run started immediately after. Since we have no
+        # language-agnostic way to tell, we accept that we might run twice in that case.
+        # The run could also finish immediately after we tried, and the lockfile be absent.
+        # In that case we'll fail with poor error reporting.
+        lockfile = puppet_config['agent_catalog_run_lockfile'] || ''
```
mruzicka: We could probably do better if we used some more sophisticated locking mechanism in puppet-agent itself.

modules/pxp-module-puppet
```diff
@@ -214,58 +229,63 @@ module Pxp
...
-            # Can't identify lockfile, just report an error.
-            return self.class.make_error_result(exitcode, Errors::AlreadyRunning,
-                                     "Puppet agent is already performing a run; it also appears to be" +
-                                     " misconfigured, agent_catalog_run_lockfile is an empty string")
+        # Check for a lockfile. If present, wait until it's removed and try running again.
+        # There's a chance that our run finished with a real error rather than because Puppet was
+        # already running, but another run started immediately after. Since we have no
+        # language-agnostic way to tell, we accept that we might run twice in that case.
+        # The run could also finish immediately after we tried, and the lockfile be absent.
+        # In that case we'll fail with poor error reporting.
```
mruzicka: We could probably do better if we used some more sophisticated locking mechanism in puppet-agent itself.

PR #627, opened 2017-08-22
(maint) Drop ifdefs for linting, fix finding task_wrapper via PATH

MikaelSmith:
The code looks bad, linting doesn't like it, and it should have no effect.

If pxp-agent is invoked by finding it on PATH, finding the task_wrapper would search on the current working directory and likely fail. Fix so if pxp-agent was found via PATH, task_wrapper will also be searched via PATH.

### DIFF
--- a/exe/task_wrapper.cc
+++ b/exe/task_wrapper.cc
@@ -33,10 +33,8 @@ int main(int argc, char *argv[])
             {},       // environment
             nullptr,  // PID callback
             0,        // timeout
-#ifndef _WIN32
-            // Only override on non-Windows. On Windows we inherit directory ACLs.
+            // Has no effect on Windows. We instead rely on inherited directory ACLs.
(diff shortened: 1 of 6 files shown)


MikaelSmith:
Seems to work fine in local testing on Windows.

exe/task_wrapper.cc
```diff
@@ -33,10 +33,8 @@ int main(int argc, char *argv[])
             {},       // environment
             nullptr,  // PID callback
             0,        // timeout
-#ifndef _WIN32
             // Only override on non-Windows. On Windows we inherit directory ACLs.
```
mruzicka: This comment is obsolete.

MikaelSmith: Fixed. There was also one in configuration.hpp I updated as well.

PR #628, opened 2017-08-22
(maint) Don't check permissions for log_file when set to stdout

ekinanp:


lib/src/configuration.cc
```diff
@@ -241,7 +241,9 @@ std::string Configuration::setupLogging()
         pcp_access_fstream_ptr_.reset(
             new boost::nowide::ofstream(pcp_access_logfile_.c_str(),
                                         std::ios_base::app));
-        fs::permissions(logfile_, NIX_FILE_PERMS);
+        if (!log_on_stdout) {
+            fs::permissions(logfile_, NIX_FILE_PERMS);
```
MikaelSmith: Oh, the bug here was that it referred to logfile_ rather than pcp_access_logfile_. Can you fix that instead? We handle permissions of logfile_ at line 231.

ekinanp: @MikaelSmith Ah ok makes sense. Just incorporated the change.

PR #631, opened 2017-08-23
(PCP-773) Implement client-side failover for task downloads

ekinanp:
WIP, this is an initial skeleton. I left some TODOs there for a few questions I have about the behavior.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -18,6 +18,8 @@
 #include <openssl/evp.h>
 #include <curl/curl.h>
 
+#include <tuple>
+
 namespace PXPAgent {
 namespace Modules {
 
@@ -226,8 +228,7 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
     return url;
 }
 
-// Downloads the file at the specified url into the provided path. If something goes
-// wrong during the download, a PXP-error should be returned. Note that the provided
+// Downloads the file at the specified url into the provided path. Note that the provided
 // "file_path" argument is a temporary file, call it "tempA". Leatherman.curl during
 // the download method will create another temporary file, call it "tempB", to save
 // the downloaded task file's contents in chunks before renaming it to "tempA." The
@@ -240,30 +241,54 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
 //    the same task file.
 // The downloaded task file's permissions will be set to rwx for user and rx for
 // group for non-Windows OSes.
-static void downloadTaskFile(const std::vector<std::string>& master_uris,
-                             lth_curl::client& client,
-                             const fs::path& file_path,
-                             const lth_jc::JsonContainer& uri) {
-    auto url = master_uris[0] + createUrlEndpoint(uri);
-    lth_curl::request req(url);
-    // timeout from connection after one minute, can configure
-    req.connection_timeout(60000);
+//
+// The method returns a tuple (success, err_msg). success is true if the file was downloaded;
+// false otherwise. err_msg contains the most recent http_file_download_exception's error
+// message; it is initially empty.
+static std::tuple<bool, std::string> downloadTaskFile(const std::vector<std::string>& master_uris,
+                                                      lth_curl::client& client,
+                                                      const fs::path& file_path,
+                                                      const lth_jc::JsonContainer& uri) {
+    auto endpoint = createUrlEndpoint(uri);
+    std::tuple<bool, std::string> result = std::make_tuple(false, "");
+    for (auto& master_uri : master_uris) {
+        auto url = master_uri + endpoint;
+        lth_curl::request req(url);
+        // timeout from connection after one minute, can configure
+        req.connection_timeout(60000);
+        try {
 #ifndef _WIN32
-    client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
+            client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
 #else
-    client.download_file(req, file_path.string());
+            client.download_file(req, file_path.string());
 #endif
+        } catch (lth_curl::http_file_download_exception& e) {
+            // Server-side error, do nothing here -- we want to try the next master-uri.
+            LOG_WARNING("Downloading the task file from the master-uri '{1}' failed. Reason: {2}", master_uri, e.what());
+            std::get<1>(result) = e.what();
+        } catch (lth_curl::http_request_exception& e) {
+            // For http_curl_setup and http_file_operation exceptions
+            throw Module::ProcessingError(lth_loc::format("Downloading the task file failed. Reason: {1}", e.what()));
+        }
+
+        if (fs::exists(file_path)) {
+            std::get<0>(result) = true;
+            return result;
+        }
+    }
+
+    return result;
 }
 
 // This method does the following. If the file matching the "filename" field of the
 // file_obj JSON does not exist OR if its hash does not match the sha value in the
 // "sha256" field of file_obj, then:
-//    (1) The file is downloaded using Leatherman.curl from the first entry in
-//    master-uris to a temporary directory. If this download fails, a PXP error
-//    is thrown.
+//    (1) The file is downloaded using Leatherman.curl by trying each of the master_uris
+//    until one of them succeeds. If this download fails, a PXP error is thrown.
 //
 //    (2) If the downloaded file's sha does not match the provided sha, then a PXP
-//        error is returned.
+//        error is returned. TODO: Now that we are trying all the master_uris for
+//        download, should we try another master_uri if the shas do not match?
 //
 //    (3) If (1) and (2) both succeed, then the downloaded file is atomically
 //        renamed to cache_dir/<filename>
@@ -285,8 +310,16 @@ static fs::path updateTaskFile(const std::vector<std::string>& master_uris,
     if (master_uris.empty()) {
         throw Module::ProcessingError(lth_loc::format("Cannot download task. No master-uris were provided"));
     }
+
     auto tempname = cache_dir / fs::unique_path("temp_task_%%%%-%%%%-%%%%-%%%%");
-    downloadTaskFile(master_uris, client, tempname, file.get<lth_jc::JsonContainer>("uri"));
+    auto download_result = downloadTaskFile(master_uris, client, tempname, file.get<lth_jc::JsonContainer>("uri"));
+    if (!std::get<0>(download_result)) {
+        throw Module::ProcessingError(lth_loc::format(
+              "Downloading the task file {1} failed after trying all the available master-uris. Most recent error message: {2}",
+              file.get<std::string>("filename"),
+              std::get<1>(download_result)));
+    }
+
     if (sha256 != calculateSha256(tempname.string())) {
       fs::remove(tempname);
       throw Module::ProcessingError(lth_loc::format("The downloaded {1}'s sha differs from the provided sha", filename));
@@ -313,8 +346,6 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
     try {
         auto cache_dir = createCacheDir(task_cache_dir, file.get<std::string>("sha256"));
         return updateTaskFile(master_uris, client, cache_dir, file);
-    } catch (lth_curl::http_file_download_exception& e) {
-        throw Module::ProcessingError(lth_loc::format("Downloading the task file {1} failed. Reason: {2}", file.get<std::string>("filename"), e.what()));
     } catch (fs::filesystem_error& e) {
         throw toModuleProcessingError(e);
     }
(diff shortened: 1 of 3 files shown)


lib/src/modules/task.cc
```diff
@@ -240,30 +239,43 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
 #else
-    client.download_file(req, file_path.string());
+            client.download_file(req, file_path.string());
 #endif
+            break;
+        } catch (lth_curl::http_file_download_exception& e) {
+            // Do nothing here -- we want to try the next master-uri.
+
+            // TODO: Should we collect all the exception messages here to pass in to the overall
+            // Module::ProcessingError if all master_uris fail? Might not be a bad idea.
```
mruzicka: Yes, we should do that.

lib/src/modules/task.cc
```diff
@@ -240,30 +239,43 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
-//    (1) The file is downloaded using Leatherman.curl from the first entry in
-//    master-uris to a temporary directory. If this download fails, a PXP error
-//    is thrown.
+//    (1) The file is downloaded using Leatherman.curl by trying each of the master_uris
+//    until one of them succeeds. If this download fails, a PXP error is thrown.
 //
 //    (2) If the downloaded file's sha does not match the provided sha, then a PXP
-//        error is returned.
+//        error is returned. TODO: Now that we are trying all the master_uris for
+//        download, should we try another master_uri if the shas do not match?
```
mruzicka: IMO we should.

MikaelSmith:
I wouldn't bother rebasing this, as we'll need to do a clean revert via https://github.com/puppetlabs/pxp-agent/pull/634 before we can merge this.

lib/src/modules/task.cc
MikaelSmith: I'd like to distinguish between problems connecting to the server and problems writing the file to disk. That might require a change to Leatherman to make possible (so we can distinguish between types of failures).

lib/src/modules/task.cc
MikaelSmith: Not sure I agree. If we get an incorrect sha, presumably that's because a newer version has rolled out. In normally operating infrastructure, we'd expect the certainty of that newer version being rolled out to all masters to increase as time goes on, so I'd expect it to almost always fail on subsequent masters.

lib/src/modules/task.cc
ekinanp: Writing to the disk sounds like a filesystem_error, while server stuff is a CURL error. Because I think we really write to the disk as in write to our main file at the last fs::rename part in download_file. But then there's the question of what if things fail at the write_file callback? According to https://curl.haxx.se/libcurl/c/libcurl-errors.html, a CURL-WRITE-ERROR is returned for that case. I think the question is "What would be considered a 'problem writing to disk' vs. connecting to the server?"

MikaelSmith: Ah... I would consider CURL-WRITE-ERROR a 'problem writing to disk'.

ekinanp: What about the rename call at the end?

MikaelSmith: Also filesystem issue. I think our failure modes are: connectivity failure, file operation failure.

ekinanp: Okay, so I'm thinking something along the following lines for the Leatherman changes:

(1) If a filesystem_error is thrown, then that is a file operation failure (writing to disk)

(2) If curl_easy_perform returns CURL_WRITE_ERROR, that is also a file operation failure.

(3) Otherwise, it's a server error.

It seems like (1) and (2) require a new exception class representing file operation failures, unless we want to use fs::filesystem_error. (3) is an http_file_download_exception.

MikaelSmith: 👍 probably a new exception is better

lib/src/modules/task.cc
mruzicka: What about the opposite case: Obtaining a SHA of the most recent version but downloading an older file?
In that case it would make sense to try downloading from other masters precisely for the reason you describe.
But can that happen in a normally operating infrastructure? I guess that depends on the order of the URIs in the `master-uris` configuration parameter.

MikaelSmith: That's fair. In practice, multiple masters will generally be used for MoM/replicas, or load balancers. In the 1st case, the MoM should be up-to-date. With load balancers, you might get an out-of-date master first, then an up-to-date master. I still think it seems odd to retry just because you didn't get the answer you expected.

MikaelSmith:
This needs an update for LTH-142.

ekinanp:
Updated the PR with the new leatherman changes. There are still two lingering questions:

(1) In the catch-block for http_file_download exceptions, should I store all of the exception messages and then later print them all out line by line if the entire download fails? 

(2) The ongoing sha discussion.

lib/src/modules/task.cc
```diff
@@ -240,30 +239,49 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
+            client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
 #else
-    client.download_file(req, file_path.string());
+            client.download_file(req, file_path.string());
 #endif
+        } catch (lth_curl::http_file_download_exception& e) {
+            // Server-side error, do nothing here -- we want to try the next master-uri.
+
+            // TODO: Should we collect all the exception messages here to pass in to the overall
+            // Module::ProcessingError if all master_uris fail? Might not be a bad idea.
```
MikaelSmith: I would log something, and throw the last one.

lib/src/modules/task.cc
```diff
@@ -240,30 +239,49 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
+        req.connection_timeout(60000);
+        try {
 #ifndef _WIN32
-    client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
+            client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
 #else
-    client.download_file(req, file_path.string());
+            client.download_file(req, file_path.string());
 #endif
+        } catch (lth_curl::http_file_download_exception& e) {
```
MikaelSmith: That's not true, this needs to exclude http_file_operation_exception.

Also, Leatherman 1.1.2 is tagged, so should be usable.

ekinanp: Thanks for that catch, you're right! Making the fix.

MikaelSmith: Ok, this should be addressed by Leatherman 1.1.3.

lib/src/modules/task.cc
MikaelSmith: I still don't think we should try unless we get feedback saying that's an issue. It's more complicated and not necessarily the right fix.

lib/src/modules/task.cc
```diff
@@ -240,30 +241,54 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
+            throw Module::ProcessingError(lth_loc::format("Downloading the task file failed. Reason: {1}", e.what()));
+        }
+
+        if (fs::exists(file_path)) {
+            boost::get<0>(result) = true;
+            return result;
+        }
+    }
+
+    return result;;
```
MikaelSmith: Only need 1 semicolon.

lib/src/modules/task.cc
```diff
@@ -240,30 +241,54 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
-                             const lth_jc::JsonContainer& uri) {
-    auto url = master_uris[0] + createUrlEndpoint(uri);
-    lth_curl::request req(url);
-    // timeout from connection after one minute, can configure
-    req.connection_timeout(60000);
+//
+// The method returns a tuple (success, err_msg). success is true if the file was downloaded;
+// false otherwise. err_msg contains the most recent http_file_download_exception's error
+// message; it is initially empty.
+static boost::tuple<bool, std::string> downloadTaskFile(const std::vector<std::string>& master_uris,
```
MikaelSmith: You probably actually want `std::tuple` (that's the header you included.

lib/src/modules/task.cc
```diff
@@ -240,30 +241,54 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
+                                                      lth_curl::client& client,
+                                                      const fs::path& file_path,
+                                                      const lth_jc::JsonContainer& uri) {
+    auto endpoint = createUrlEndpoint(uri);
+    boost::tuple<bool, std::string> result = boost::make_tuple(false, "");
+    for (auto& master_uri : master_uris) {
+        auto url = master_uri + endpoint;
+        lth_curl::request req(url);
+        // timeout from connection after one minute, can configure
```
MikaelSmith: I'd make use of the existing `connection-timeout` setting. I believe curl's connection timeout only applies to the initial connection, not a timeout for completing the download.

lib/src/modules/task.cc
```diff
@@ -240,30 +241,54 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
+//
+// The method returns a tuple (success, err_msg). success is true if the file was downloaded;
+// false otherwise. err_msg contains the most recent http_file_download_exception's error
+// message; it is initially empty.
+static boost::tuple<bool, std::string> downloadTaskFile(const std::vector<std::string>& master_uris,
+                                                      lth_curl::client& client,
+                                                      const fs::path& file_path,
+                                                      const lth_jc::JsonContainer& uri) {
+    auto endpoint = createUrlEndpoint(uri);
+    boost::tuple<bool, std::string> result = boost::make_tuple(false, "");
```
MikaelSmith: Use `auto result = std::make_tuple(false, std::string{})`.

lib/src/modules/task.cc
```diff
@@ -240,30 +241,54 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
...
-    client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
+            client.download_file(req, file_path.string(), NIX_TASK_FILE_PERMS);
 #else
-    client.download_file(req, file_path.string());
+            client.download_file(req, file_path.string());
 #endif
+        } catch (lth_curl::http_file_download_exception& e) {
+            // Server-side error, do nothing here -- we want to try the next master-uri.
+            LOG_WARNING("Downloading the task file from the master-uri '{1}' failed. Reason: {2}", master_uri, e.what());
+            boost::get<1>(result) = e.what();
```
MikaelSmith: With `std::tuple`, you'd use `std::get`.

lib/src/modules/task.cc
ekinanp: Will do, I wasn't sure if we should use boost or std there b/c we use boost more often (i.e. not sure of when to use boost vs. std).

lib/src/modules/task.cc
ekinanp: Unfortunately I can't, b/c auto makes the result a constant tuple so I can't modify it.

lib/src/modules/task.cc
ekinanp: Yeah, you're right. Their default for the initial connection is five minutes; should we keep that?

MikaelSmith: pxp-agent has a config option `connection-timeout` that defaults to 5 seconds. I think that's what we should use.

MikaelSmith: I guess leave it as-is for now, we'll address that separately.

ekinanp:
Alright, updated my PR including the commit message as it's closer to the final state. Did local testing on my machine for the following cases:
  (1) Ensure that pxp-agent falls over to other master-uris.
  (2) Ensure that pxp-agent only throws a download error when all master-uris fail.

The existing functionality from PCP-769 works. Also modified an existing unit test; depending on how CI goes, the modification might have to be turned off for Windows only.

lib/tests/unit/modules/task_test.cc
```diff
@@ -305,7 +306,7 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
         REQUIRE(boost::contains(response.action_metadata.get<std::string>("execution_error"), "Cannot download task. No master-uris were provided"));
     }
 
-    SECTION("errors when a download error occurs and removes the temporary file") {
+    SECTION("if a master-uri has a server-side error, then it proceeds to try the next master-uri. if they all fail, it errors on download and removes all temporary files") {
```
MikaelSmith: We should also cover the previous error case.

MikaelSmith: I'll add it post-hoc.

MikaelSmith: I guess this test pretty much covers it, just slightly change in the error message.

PR #637, opened 2017-08-28
(maint) Update task blocking response to return results directly

MikaelSmith:
The output from a blocking response was wrapped in a key to make it a
valid json object. That was a hack that's unnecessary, fix it by
allowing the task module to return a string, and returning the result
string directly.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -120,7 +120,7 @@ Task::Task(const fs::path& exec_prefix,
     actions.push_back(TASK_RUN_ACTION);
 
     PCPClient::Schema input_schema { TASK_RUN_ACTION, lth_jc::JsonContainer { TASK_RUN_ACTION_INPUT_SCHEMA } };
-    PCPClient::Schema output_schema { TASK_RUN_ACTION };
+    PCPClient::Schema output_schema { TASK_RUN_ACTION, PCPClient::TypeConstraint::String };
 
     input_validator_.registerSchema(input_schema);
     results_validator_.registerSchema(output_schema);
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Might be able to fix this so it doesn't need to add the `_output` key.

MikaelSmith:
This makes it consistent with the result from a non-blocking response.

lib/src/modules/task.cc
```diff
@@ -510,12 +510,14 @@ void Task::processOutputAndUpdateMetadata(ActionResponse& response)
...
 
     if (isValidUTF8(output)) {
+        // Use a temporary object to create a JSON string. The JsonContainer API doesn't
+        // provide a way to set the root to a string unless that string is already quoted.
         lth_jc::JsonContainer stdout_result;
-        stdout_result.set("output", output);
+        stdout_result.set("_output", output);
 
-        response.setValidResultsAndEnd(std::move(stdout_result));
+        response.setValidResultsAndEnd(stdout_result.get<lth_jc::JsonContainer>("_output"));
```
nicklewis: Can we not use the `_output` key here, since the semantics of this key are very different from the orchestrator's `_output` key?

MikaelSmith: It's completely omitted in the results, but I can use something that's obvious not a real key.

lib/src/modules/task.cc
```diff
@@ -510,12 +510,20 @@ void Task::processOutputAndUpdateMetadata(ActionResponse& response)
     std::string &output = response.output.std_out;
 
     if (isValidUTF8(output)) {
-        lth_jc::JsonContainer stdout_result;
-        stdout_result.set("output", output);
+        // Use a temporary object to create a JSON string. The JsonContainer API doesn't
+        // provide a way to set the root to a string unless that string is already quoted.
+        lth_jc::JsonContainer result;
```
MikaelSmith: This avoids the mistake we made with external modules, by returning all the relevant results in the non-blocking response.

PR #638, opened 2017-08-29
(maint) Improve test debugibility and failure handling

MikaelSmith:
When an action finishes, it always returns an exitcode. stdout and
stderr are only included if they have contents. Identify completion by
the presence of exitcode so we can finish early, rather than timing out
if a failure means stdout is empty.

Also include debug logging when Beaker is configured to use debug
output.

[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -504,21 +504,21 @@ def start_puppet_non_blocking_request(broker, targets, environment = 'production
 end
 
 def check_non_blocking_response(broker, identity, transaction_id, max_retries, query_interval, &block)
-  run_result = nil
+  run_result = rpc_action_status = nil
   query_attempts = 0
-  until (query_attempts == max_retries || run_result) do
+  loop do
     query_responses = rpc_blocking_request(broker, [identity],
                                            'status', 'query', {:transaction_id => transaction_id})
     action_result = query_responses[identity][:data]['results']
     assert(action_result, "Response to status query was an error: #{query_responses[identity][:data]}")
-    if (action_result.has_key?('stdout') && (action_result['stdout'] != ""))
+    if action_result.has_key?('exitcode')
       rpc_action_status = action_result['status']
       run_result = action_result['stdout']
+      break
     end
     query_attempts += 1
-    if (!run_result)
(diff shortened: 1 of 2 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -504,14 +504,14 @@ def start_puppet_non_blocking_request(broker, targets, environment = 'production
...
-  until (query_attempts == max_retries || run_result) do
+  until (query_attempts == max_retries || rpc_action_status) do
     query_responses = rpc_blocking_request(broker, [identity],
                                            'status', 'query', {:transaction_id => transaction_id})
     action_result = query_responses[identity][:data]['results']
     assert(action_result, "Response to status query was an error: #{query_responses[identity][:data]}")
-    if (action_result.has_key?('stdout') && (action_result['stdout'] != ""))
+    if action_result.has_key?('exitcode')
       rpc_action_status = action_result['status']
       run_result = action_result['stdout']
```
mruzicka: It seems we could just `break` out of the loop here, which would allow us to drop the testing of the `rpc_action_status` in the condition of the loop and the `if` condition (which still tests the possibly empty stdout) around the sleep below.

mruzicka: Actually I'd discard the loop condition altogether in favor of:
```ruby
loop do
    ...
    if action_result.has_key?('exitcode')
        rpc_action_status = action_result['status']
        run_result = action_result['stdout']
        break
    end
    query_attempts += 1
    break if query_attempts == max_retries
    sleep query_interval
end
```

MikaelSmith: Looks better, updated.

MikaelSmith:
```
       Total Suite Time: 376.77 seconds
      Average Test Time: 13.46 seconds
              Attempted: 28
                 Passed: 26
                 Failed: 0
                Errored: 0
                Skipped: 2
                Pending: 0
                  Total: 28

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
  Test Case tests/multiple_pxp_agent_daemon.rb
  Test Case tests/tasks/run_ps1.rb
Pending Tests Cases:
```

PR #639, opened 2017-08-29
(maint) Update task blocking response to return all results

MikaelSmith:
In the task response, return exitcode and stderr as well as stdout. This
allows a consumer to skip querying status, as all the data they need for
a task is in the response.

The status query used to retrieve task results returns stdout, stderr,
and exitcode. This makes the response to running a task - which is
returned immediately upon completion, rather than in response to a
status request - contain the same information so a controller can avoid
a follow-up request to get it.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -510,12 +510,19 @@ void Task::processOutputAndUpdateMetadata(ActionResponse& response)
     std::string &output = response.output.std_out;
 
     if (isValidUTF8(output)) {
-        lth_jc::JsonContainer stdout_result;
-        stdout_result.set("output", output);
+        // Return all relevant results: exitcode, stdout, stderr.
+        lth_jc::JsonContainer result;
+        result.set("exitcode", response.output.exitcode);
+        if (!output.empty()) {
+            result.set("stdout", output);
+        }
+        if (!response.output.std_err.empty()) {
+            result.set("stderr", response.output.std_err);
+        }
 
-        response.setValidResultsAndEnd(std::move(stdout_result));
+        response.setValidResultsAndEnd(std::move(result));
     } else {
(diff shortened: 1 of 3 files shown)


MikaelSmith:
```
2017-08-29 14:03:29,002 TRACE [qtp1067320520-19] [p.p.b.shared] Sending PCP message to pcp://puppet:8143/server: {:id "c8b1c15d-0301-44dd-8b65-d86b483d74f6", :message_type "http://puppetlabs.com/rpc_non_blocking_response", :target "pcp://puppet:8143/server", :sender "pcp://10-32-167-127.rfc1918.puppetlabs.net/agent", :data {:transaction_id "202", :results {:exitcode 0, :stdout "{\"message\":\"hello\"}\n"}}}
```

MikaelSmith:
@nicklewis did you have any more thoughts on this?

nicklewis:
No, I think it makes sense to me.

lib/src/modules/task.cc
```diff
@@ -510,12 +510,19 @@ void Task::processOutputAndUpdateMetadata(ActionResponse& response)
...
-        lth_jc::JsonContainer stdout_result;
-        stdout_result.set("output", output);
+        // Return all relevant results: exitcode, stdout, stderr.
+        lth_jc::JsonContainer result;
+        result.set("exitcode", response.output.exitcode);
+        if (!output.empty()) {
+            result.set("stdout", output);
+        }
+        if (!response.output.std_err.empty()) {
+            result.set("stderr", response.output.std_err);
```
mruzicka: Should we sanitize this to ensure it is valid UTF-8?

MikaelSmith: Would would also want to do that in the else clause. Maybe, but I'm tempted to say it's unlikely we'll get binary data on stderr.

mruzicka: Fair enough.

PR #641, opened 2017-08-29
(PCP-769) Fix permissions hacks

MikaelSmith:
Applying permissions on Windows and using fs::permissions on Solaris
were both hacked around due to a faulty assumption that they were broken
on those platforms. The root cause of both issues was a static
initialization ordering issue.

Fix it by using a macro to alias the variable name, rather than
assigning from a global variable in another object. Unwind Windows and
Solaris hacks.

### DIFF
--- a/lib/inc/pxp-agent/configuration.hpp
+++ b/lib/inc/pxp-agent/configuration.hpp
@@ -24,11 +24,9 @@ namespace PXPAgent {
 
 extern const std::string DEFAULT_SPOOL_DIR;     // used by unit tests
 
-#ifndef _WIN32
-// Not used on Windows. We instead rely on inherited directory ACLs.
+// Note that on Windows we rely on inherited directory ACLs.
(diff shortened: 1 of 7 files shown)


MikaelSmith:
Tests running at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent_init-van-int_suite-daily-ad-hoc/107/.

MikaelSmith:
pxp-agent tests passed on redhat, ubuntu, windows, solaris. I'll see about Windows versions of the permissions tests.

MikaelSmith:
Tests updated.

mruzicka:
So the problem was the functions making use of the global variable were called _before_ the variable was initialized?

MikaelSmith:
Not quite. The global variable was copying from other global variable before that one was initialized.

mruzicka:
Ah, I see!

PR #642, opened 2017-08-30
(PCP-782) Add port 8140 to master-uris if absent

MikaelSmith:
If master-uris includes a host with no port specified, add port 8140.
This ensures it can use the same values as Puppet's server_list and
should successfully connect to the master.


MikaelSmith:
Will require a release of https://github.com/puppetlabs/leatherman/pull/258.

lib/src/configuration.cc
```diff
@@ -841,8 +835,15 @@ void Configuration::validateAndNormalizeWebsocketSettings()
         }
     }
 
-    for (auto const& uri : master_uris_) {
-        validate_https(uri, "master-uris");
+    for (auto &uri : master_uris_) {
+        auto parsed = lth_util::uri(uri);
+        if (parsed.protocol.empty()) {
+            parsed.protocol = "https";
```
adreyer: should we reject http:// ?

MikaelSmith: Hmm, I guess curl is going to explicitly use HTTPS. I'll restore throwing if we specify something besides HTTPS.

MikaelSmith:
Updated and running through CI.

lib/src/configuration.cc
```diff
@@ -841,8 +835,21 @@ void Configuration::validateAndNormalizeWebsocketSettings()
...
+        }
+
+        if (parsed.protocol != "https") {
```
mruzicka:
```C++
} else if () {
```
?

PR #649, opened 2017-09-06
(PCP-786) Send error response to status if found

MikaelSmith:
If an execution error is found in the metadata of a non-blocking action,
send a response with `{"_error": {"kind":
"puppetlabs.pxp-agent/task-error", "msg": <msg>, "details" {}}}` on
stdout. This should only happen if an error occurs while invoking an
action that would result in sending an `rpc_error_message`, and is done
to make sure the same error message can be retrieved by a status request
if the `rpc_error_message` is lost.

### DIFF
--- a/lib/src/action_response.cc
+++ b/lib/src/action_response.cc
@@ -97,7 +97,7 @@ ActionResponse::ActionResponse(ModuleType module_type_,
           request_type { request.type() },
           output {},
           action_metadata { getMetadataFromRequest(request) },
-          status_query_transaction { status_query_transaction_ }
+          status_query_transaction { std::move(status_query_transaction_) }
 {
 }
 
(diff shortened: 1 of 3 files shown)


MikaelSmith:
I'm a little wary of fixing this, because it's somewhat unlikely and the right way to fix it is probably to add a new PXP message type. An alternative fitting within the current PXP specification would be to send an `rpc_error_message` in response to the status request, but that seems somewhat misleading as it would imply the status request itself failed.

lib/src/action_response.cc
```diff
@@ -252,10 +252,25 @@ lth_jc::JsonContainer ActionResponse::toJSON(R_T response_type) const
                     ACTION_STATUS_NAMES.at(ActionStatus::Unknown));
             }
 
-            if (!output.std_out.empty())
+            // If an execution error exists, report that instead of any results.
+            if (action_metadata.includes(EXECUTION_ERROR)) {
+                auto exec_err = action_metadata.get<std::string>(EXECUTION_ERROR);
+                if (!exec_err.empty()) {
+                    lth_jc::JsonContainer err_obj;
+                    err_obj.set("kind", "puppetlabs.pxp-agent/action-error");
```
MikaelSmith: This could also be `execution-error`, seems consistent with the rest of naming.

One caveat with this change is that new agents with old Orchestrator could - potentially - result in sending a message Orchestrator doesn't understand (the _error key) in response to a pxp-module-puppet run that fails due to https://github.com/puppetlabs/pxp-agent/blob/master/lib/src/external_module.cc#L172-L184.

MikaelSmith: @nicklewis an alternative that would be more backward compatible with Orchestrator is to send an rpc_error_message, and add a kind field to that. It'd violate the [PXP spec](https://github.com/puppetlabs/pcp-specifications/blob/master/pxp/versions/1.0/request_response.md#error-message), but nothing except pxp-agent would care.

nicklewis: Invalid JSON on stdout can't really happen, can it? That would only be the case for a broken module. So execution-error seems fine, assuming orchestrator doesn't have any automatic weird handling of that.

MikaelSmith: Thin I think this is ready.

PR #656, opened 2017-09-08
(packaging) Update beaker and PA install method

melissa:
The latest beaker release includes new install utilities that enable
less reliance on correctly guessing the path to artifacts staged on
builds.delivery.puppetlabs.net. This commit updates the puppet tests to
use these new install utilities.


MikaelSmith:
Does this remove the new for SUITE_VERSION?

melissa:
Was that a typo? I'm not quite sure, but it seems you're asking if it removed the "need" for SUITE_VERSION? That I'm not actually sure. I don't know what SUITE_VERSION was previously used for, or how it differs from SHA. I may have grabbed the wrong variable for my changes, I'm not sure. I do know tests were green though

MikaelSmith:
Yup, meant need. It was sometimes used for SHA, but SUITE_VERSION was required on platforms where the file name used a `git describe`.

acceptance/setup/aio/010_Install.rb
```diff
@@ -31,10 +31,7 @@ def install_from_local(hosts, package)
       :puppet_agent_sha     => ENV['SHA'],
       :puppet_agent_version => ENV['SUITE_VERSION'] || ENV['SHA']
     }
-    agents.each do |agent|
-      next if agent == master # Avoid SERVER-528
-      install_puppet_agent_dev_repo_on(agent, opts)
-    end
+    install_from_build_data_url('puppet-agent', "http://builds.delivery.puppetlabs.net/puppet-agent/#{opts[:puppet_agent_sha]}/artifacts/#{opts[:puppet_agent_sha]}.yaml")
```
MikaelSmith: This removes the need for the `opts` object completely. Just use `ENV['SHA']` directly.

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -162,7 +162,7 @@ def install_repos_on(host, project, sha, repo_configs_dir)
                     :puppet_agent_version => ENV['SUITE_VERSION'] || ENV['SHA']
                 }
                 # this installs puppet-agent on windows (msi) and osx (dmg)
-                install_puppet_agent_dev_repo_on(agent, opts)
+                install_from_build_data_url('puppet-agent', "http://builds.delivery.puppetlabs.net/puppet-agent/#{opts[:puppet_agent_sha]}/artifacts/#{opts[:puppet_agent_sha]}.yaml")
```
MikaelSmith: This removes the need for the `opts` object completely. Just use `ENV['SHA']` directly.

MikaelSmith:
It looks like there are some helpers I can probably remove, but I can handle that separately.

acceptance/lib/puppet/acceptance/install_utils.rb
MikaelSmith: Actually, `install_repos_on` is never used, so we could remove it completely.

MikaelSmith: Same is true of everything else in this file except `install_packages_on`.

acceptance/setup/aio/010_Install.rb
MikaelSmith: What this does preclude is anyone outside Puppet running these tests.

MikaelSmith: Also, this applies to all hosts. That means this setup needs to be slightly restructured.

MikaelSmith:
I'm working on updates around this.

melissa:
@MikaelSmith you can see more conversation around these changes at https://github.com/puppetlabs/puppet/pull/6187

I think Kevin and I are going to update the new install method to take a host or array of hosts as an argument. This will let us keep more of the current functionality.

The last question we need to address is how much knowledge of the config url that gets passed in has a default in beaker vs in the project. I definitely agree we should not be hard coding `builds.delivery.puppetlabs.com`, but I also think we need to be able to overwrite the url itself. Anyway, we're still hashing out the details.

acceptance/lib/puppet/acceptance/install_utils.rb
melissa: Should I just clean up this file?

MikaelSmith:
I think i’ve done some of that cleanup already. Try a rebase.

melissa:
I've retargeted this work at the 1.5.x branch so that it'll be available on the LTS moving forward. Is that going to mess things up?

MikaelSmith:
Ah, it may make sense to cherry-pick some of the changes I made back. I’ll take a look.

MikaelSmith:
I would cherry-pick 876ff76ae29f4794318781793e2db87e42bca1a4, then rebase your code on that.

melissa:
@MikaelSmith updated, but I want to run this through tests before it gets merged

acceptance/setup/aio/010_Install.rb
```diff
@@ -5,15 +5,8 @@
...
-    :puppet_collection    => 'PC1',
-    :puppet_agent_sha     => ENV['SHA'],
-    :puppet_agent_version => ENV['SUITE_VERSION'] || ENV['SHA']
-  }
-  agents.each do |agent|
-    next if agent == master # Avoid SERVER-528
-    install_puppet_agent_dev_repo_on(agent, opts)
-  end
+  dev_builds_url = ENV['DEV_BUILDS_URL'] || 'http://builds.delivery.puppetlabs.net'
+  install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{ENV['SHA']}/artifacts/#{ENV['SHA']}.yaml", agents)
```
MikaelSmith: Is this what we came to on the Puppet PR? I'd prefer to see this overridable in a way that you could use a publicly released artifact directly from puppet.com.

acceptance/setup/aio/010_Install.rb
```diff
@@ -29,17 +22,16 @@
       on(master, "rpm -Uvh https://yum.puppetlabs.com/puppet5-release-el-#{version}.noarch.rpm")
     end
   else
+    dev_builds_url = ENV['DEV_BUILDS_URL'] || 'http://builds.delivery.puppetlabs.net'
```
MikaelSmith: Could be set once outside the steps and used in both places.

acceptance/setup/aio/010_Install.rb
MikaelSmith: Also, `agents` doesn't apply to the master. So you probably want `hosts` to install on the master as well.

melissa: Oh right, sorry, I left this with agents because we were also installing the agent below, but I forgot I took that out

MikaelSmith: I take it hosts is the default if you don't pass an argument then?

MikaelSmith:
I'd suggest squashing your changes before merge, and use `[skip ci]` to disable AppVeyor/Travis.

acceptance/setup/aio/010_Install.rb
melissa: The previous way were were installing this, with `install_puppet_agent_dev_repo_on`, was also very specific to our infrastructure
```
        # Install development repo of the puppet-agent on the given host(s).  Downloaded from
        # location of the form DEV_BUILDS_URL/puppet-agent/AGENT_VERSION/repos
        #
        # @param [Host, Array<Host>, String, Symbol] hosts    One or more hosts to act upon,
        #                            or a role (String or Symbol) that identifies one or more hosts.
        # @param [Hash{Symbol=>String}] opts An options hash
        # @option opts [String] :puppet_agent_version The version of puppet-agent to install. This
        #                       parameter is used by puppet with the +SUITE_VERSION+ environment
        #                       variable to provide a `git describe` value to beaker to create a
        #                       build server URL. Note that +puppet_agent_sha+ will still be used
        #                       instead of this if a value is provided for that option
        # @option opts [String] :puppet_agent_sha The sha of puppet-agent to install, defaults to provided
        #                       puppet_agent_version
        # @option opts [String] :copy_base_local Directory where puppet-agent artifact
        #                       will be stored locally
        #                       (default: 'tmp/repo_configs')
        # @option opts [String] :copy_dir_external Directory where puppet-agent
        #                       artifact will be pushed to on the external machine
        #                       (default: '/root')
        # @option opts [String] :puppet_collection Defaults to 'PC1'
        # @option opts [String] :dev_builds_url Base URL to pull artifacts from
        # @option opts [String] :copy_base_local Directory where puppet-agent artifact
        #                       will be stored locally
        #                       (default: 'tmp/repo_configs')
        # @option opts [String] :copy_dir_external Directory where puppet-agent
        #                       artifact will be pushed to on the external machine
        #                       (default: '/root')
        #
        # @note on windows, the +:ruby_arch+ host parameter can determine in addition
        # to other settings whether the 32 or 64bit install is used
        #
        # @example
        #   install_puppet_agent_dev_repo_on(host, { :puppet_agent_sha => 'd3377feaeac173aada3a2c2cedd141eb610960a7', :puppet_agent_version => '1.1.1.225.gd3377fe'  })
        #
        # @return nil
        def install_puppet_agent_dev_repo_on( hosts, opts )

          opts[:puppet_agent_version] ||= opts[:version] #backward compatability
          if not opts[:puppet_agent_version]
            raise "must provide :puppet_agent_version (puppet-agent version) for install_puppet_agent_dev_repo_on"
          end
          # TODO consolidate these values as they serve no purpose from beaker's side
          # you could provide any values you could to one to the other
          puppet_agent_version = opts[:puppet_agent_sha] || opts[:puppet_agent_version]

          opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
          opts[:download_url] = "#{opts[:dev_builds_url]}/puppet-agent/#{ puppet_agent_version }/repos/"
          opts[:copy_base_local]    ||= File.join('tmp', 'repo_configs')
          opts[:puppet_collection]  ||= 'PC1'
          release_path = opts[:download_url]

          block_on hosts do |host|
            variant, version, arch, codename = host['platform'].to_array
            add_role(host, 'aio') #we are installing agent, so we want aio role
            copy_dir_local = File.join(opts[:copy_base_local], variant)
            onhost_copy_base = opts[:copy_dir_external] || host.external_copy_base

            case variant
            when /^(fedora|el|centos|debian|ubuntu|cumulus|huaweios|cisco_nexus|cisco_ios_xr)$/
              if arch == 's390x'
                logger.trace("#install_puppet_agent_dev_repo_on: s390x arch detected for host #{host}. using dev package")
              else
                opts[:dev_builds_repos] ||= [ opts[:puppet_collection] ]
                install_puppetlabs_dev_repo( host, 'puppet-agent', puppet_agent_version, nil, opts )
                host.install_package('puppet-agent')
                logger.trace("#install_puppet_agent_dev_repo_on: install_puppetlabs_dev_repo finished")
                next
              end
            when /^(eos|osx|windows|solaris|sles|aix)$/
              # Download installer package file & run install manually.
              # Done below, so that el hosts with s390x arch can use this
              # workflow as well
            else
              raise "No repository installation step for #{variant} yet..."
            end

            release_path_end, release_file = host.puppet_agent_dev_package_info(
              opts[:puppet_collection], opts[:puppet_agent_version], opts )
            release_path << release_path_end
            logger.trace("#install_puppet_agent_dev_repo_on: dev_package_info, continuing...")

            if variant =~ /eos/
              host.get_remote_file( "#{release_path}/#{release_file}" )
            else
              onhost_copied_file = File.join(onhost_copy_base, release_file)
              fetch_http_file( release_path, release_file, copy_dir_local)
              scp_to host, File.join(copy_dir_local, release_file), onhost_copy_base
            end

            case variant
            when /^eos/
              host.install_from_file( release_file )
            when /^(sles|aix|el)$/
              # NOTE: AIX does not support repo management. This block assumes
              # that the desired rpm has been mirrored to the 'repos' location.
              on host, "rpm -ivh #{onhost_copied_file}"
            when /^windows$/
              result = on host, "echo #{onhost_copied_file}"
              onhost_copied_file = result.raw_output.chomp
              msi_opts = { :debug => host[:pe_debug] || opts[:pe_debug] }
              install_msi_on(host, onhost_copied_file, {}, msi_opts)
            when /^osx$/
              host.install_package("puppet-agent-#{opts[:puppet_agent_version]}*")
            when /^solaris$/
              host.solaris_install_local_package( release_file, onhost_copy_base )
            end
            configure_type_defaults_on( host )
          end
        end
        alias_method :install_puppetagent_dev_repo, :install_puppet_agent_dev_repo_on
```

The methods you took out, to install repos and packages, allowed users to install from nightlies though

melissa: yes, it defaults to `hosts` if no argument is passed

MikaelSmith: Ok. I guess I'll deal with that if anyone complains.

MikaelSmith:
Test run on Centos 7 passed for me.

acceptance/setup/aio/010_Install.rb
```diff
@@ -7,7 +7,14 @@
 dev_builds_url = ENV['DEV_BUILDS_URL'] || 'http://builds.delivery.puppetlabs.net'
 
 step "Install puppet-agent..." do
-  install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{ENV['SHA']}/artifacts/#{ENV['SHA']}.yaml")
+  if dev_builds_url == 'http://nightlies.puppetlabs.com'
+    hosts.each do |host|
+      install_puppetlabs_dev_repo(host, 'puppet-agent', ENV['SHA'], nil, :dev_builds_url => dev_builds_url)
```
MikaelSmith: I like that, thanks.

acceptance/setup/aio/010_Install.rb
```diff
@@ -23,11 +30,11 @@
       on(master, "rpm -Uvh https://yum.puppetlabs.com/puppet5-release-el-#{version}.noarch.rpm")
     end
   else
-    if ENV['SERVER_VERSION']
-      install_from_build_data_url('puppetserver', "#{dev_builds_url}/puppetserver/#{ENV['SERVER_VERSION']}/artifacts/#{ENV['SERVER_VERSION']}.yaml", master)
-    else
-      install_puppetlabs_dev_repo(master, 'puppetserver', 'latest', nil, :dev_builds_url => 'http://nightlies.puppetlabs.com')
+    if dev_builds_url == 'http://nightlies.puppetlabs.com'
```
MikaelSmith: Will this work in existing CI? I'm not sure what we set for AWS or Jenkins test environments.

acceptance/setup/aio/010_Install.rb
```diff
@@ -7,7 +7,14 @@
 dev_builds_url = ENV['DEV_BUILDS_URL'] || 'http://builds.delivery.puppetlabs.net'
 
 step "Install puppet-agent..." do
-  install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{ENV['SHA']}/artifacts/#{ENV['SHA']}.yaml")
+  if dev_builds_url == 'http://nightlies.puppetlabs.com'
```
MikaelSmith: I'd support either http or https. We should prefer people use https.

acceptance/setup/aio/010_Install.rb
```diff
@@ -1,20 +1,19 @@
...
-  agents.each do |agent|
-    next if agent == master # Avoid SERVER-528
-    install_puppet_agent_dev_repo_on(agent, opts)
+  if dev_builds_url == 'http://builds.delivery.puppetlabs.net'
+    install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{ENV['SHA']}/artifacts/#{ENV['SHA']}.yaml")
+  else
+    hosts.each do |host|
+      install_puppetlabs_dev_repo(host, 'puppet-agent', ENV['SHA'], nil, :dev_builds_url => dev_builds_url)
+      host.install_package('puppet-agent')
+    end
```
melissa: @MikaelSmith better?

acceptance/setup/aio/010_Install.rb
```diff
@@ -31,17 +30,22 @@
       on(master, "rpm -Uvh https://yum.puppetlabs.com/puppet5-release-el-#{version}.noarch.rpm")
     end
   else
-    if ENV['SERVER_VERSION']
-      install_puppetlabs_dev_repo(master, 'puppetserver', ENV['SERVER_VERSION'])
-      install_puppetlabs_dev_repo(master, 'puppet-agent', ENV['SHA'])
+    if ENV['SERVER_VERSION'] && dev_builds_url == 'http://builds.delivery.puppetlabs.net'
```
melissa: @MikaelSmith better?

acceptance/setup/aio/010_Install.rb
MikaelSmith: Ah, sure.

acceptance/setup/aio/010_Install.rb
MikaelSmith: Seems reasonable.

acceptance/setup/aio/010_Install.rb
```diff
@@ -31,17 +30,20 @@
...
-      install_repos_on(master, 'puppetserver', 'nightly', repo_configs_dir)
-      install_repos_on(master, 'puppet-agent', ENV['SHA'], repo_configs_dir)
+      if ENV['SERVER_VERSION'].nil? || ENV['SERVER_VERSION'] == 'latest'
+        server_version = ''
+        project = 'puppetserver-latest'
+      else
+        server_version = ENV['SERVER_VERSION']
+        project = 'puppetserver'
+      end
+      install_puppetlabs_dev_repo(master, project, server_version, nil, :dev_builds_url => dev_builds_url)
```
melissa: If `ENV['SERVER_VERSION']` isn't set, and `dev_builds_url` is set to `http://builds.delivery.puppetlabs.net`, this will fail

MikaelSmith: Ah, that's fun.

melissa: Do you want me to add in extra checks to catch that case?

MikaelSmith: In Jenkins, we set SERVER_VERSION to `latest`. We should default to that in general if not set.

MikaelSmith: That would also allow you to simplify the project variable above.

acceptance/setup/aio/010_Install.rb
```diff
@@ -31,17 +30,27 @@
...
+      if dev_builds_url == 'http://builds.delivery.puppetlabs.net'
+        dev_builds_url = 'http://nightlies.puppetlabs.com'
+      end
+    else
+      server_version = ENV['SERVER_VERSION']
+      project = 'puppetserver'
+    end
+
+    if dev_builds_url == 'http://builds.delivery.puppetlabs.net'
+      install_from_build_data_url('puppetserver', "#{dev_builds_url}/puppetserver/#{server_version}/artifacts/#{server_version}.yaml", master)
```
MikaelSmith: `project` is never used.

acceptance/setup/aio/010_Install.rb
```diff
@@ -31,17 +30,27 @@
       on(master, "rpm -Uvh https://yum.puppetlabs.com/puppet5-release-el-#{version}.noarch.rpm")
     end
   else
-    if ENV['SERVER_VERSION']
-      install_puppetlabs_dev_repo(master, 'puppetserver', ENV['SERVER_VERSION'])
-      install_puppetlabs_dev_repo(master, 'puppet-agent', ENV['SHA'])
+    if ENV['SERVER_VERSION'].nil? || ENV['SERVER_VERSION'] == 'latest'
```
MikaelSmith: How about
```
if ENV['SERVER_VERSION'] && ENV['SERVER_VERSION'] != 'latest' && dev_builds_url == 'http://builds.delivery.puppetlabs.net'
  install_from_build_data_url('puppetserver', "#{dev_builds_url}/puppetserver/#{server_version}/artifacts/#{server_version}.yaml", master)
else
  server_version = ENV['SERVER_VERSION'] || 'latest'

  # builds.delivery.puppetlabs.net can't handle an empty version or a
  # latest version, so we should default to nightlies in those cases
  if dev_builds_url == 'http://builds.delivery.puppetlabs.net'
    dev_builds_url = 'http://nightlies.puppetlabs.com'
  end

  install_puppetlabs_dev_repo(master, project, server_version, nil, :dev_builds_url => dev_builds_url)
  master.install_package('puppetserver')
end

MikaelSmith: I think the current code would still fail if you tried to run with DEV_BUILDS_URL set but SERVER_VERSION not.

melissa: If SERVER_VERSION isn't set, or is set to 'latest', I'm overwriting dev_builds_url if it's set to builds, so it shouldn't fail. But I'll update this

MikaelSmith: But people who set it to something else, i.e. nightlies.puppetlabs.com, should be able to use latest. I guess just setting latest gets what they want anyway. But mine seems a bit more compact.

MikaelSmith:
Ok, seems reasonable.

PR #660, opened 2017-09-14
(PCP-511, PCP-792) Use hocon, support non-root agent config

MikaelSmith:
Use hocon to load config

Add non-root agent  configuration. When running as non-root, use defaults based on `$HOME/.puppetlabs/pxp-agent`. This is consistent with other tools in the Puppet ecosystem.

Also ensure all directories are generated if missing, with owner/group
permissions applied to them.


MikaelSmith:
Depends on https://github.com/puppetlabs/cpp-hocon/pull/93.

lib/tests/unit/configuration_test.cc
```diff
@@ -435,6 +437,30 @@ TEST_CASE("Configuration::parseOptions duplicate broker-ws-uris", "[configuratio
     }
 }
 
+TEST_CASE("Configuration::parseOptions invalid config-file name", "[configuration]") {
+    const char* altArgv[] = {
+    "test-command",
+    "--config-file", INVALID_CONFIG_NAME.data(),
```
mruzicka: We should use `c_str()` instead of `data()` as it guarantees nul terminated string which is what we want here IMO.

lib/tests/unit/configuration_test.cc
```diff
@@ -372,6 +376,30 @@ TEST_CASE("Configuration::validate", "[configuration]") {
     }
 }
 
+TEST_CASE("Configuration::validate with unknown config options", "[configuration]") {
+    const char* altArgv[] = {
+    "test-command",
+    "--config-file", UNKNOWN_CONFIG.data(),
```
mruzicka: We should use `c_str()` instead of `data()` as it guarantees nul terminated string which is what we want here IMO.

cmake/FindCPPHOCON.cmake
```diff
@@ -0,0 +1,6 @@
+include(FindDependency)
+find_dependency(CPPHOCON DISPLAY "cpp-hocon" HEADERS "hocon/config.hpp" LIBRARIES "libcpp-hocon.a")
+
+include(FeatureSummary)
+set_package_properties(CPPHOCON PROPERTIES DESCRIPTION "A C++ parser for the HOCON configuration language" URL "https://github.com/puppetlabs/cpp-hocon")
+set_package_properties(CPPHOCON PROPERTIES TYPE REQUIRED PURPOSE "Allows parsing of the Facter config file.")
```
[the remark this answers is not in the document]
MikaelSmith: Yes, I'll update.

lib/src/configuration.cc
```diff
@@ -48,21 +53,22 @@ namespace lth_util = leatherman::util;
...
+            try {
+                return fs::path(lth_w::file_util::get_programdata_dir()) / "PuppetLabs" / "pxp-agent";
+            } catch (lth_w::file_util::unknown_folder_exception &e) {
+                throw Configuration::Error {
+                    lth_loc::format("failure getting Windows AppData directory: {1}", e.what()) };
+            }
+        } else {
+            auto home = lth_w::user::home_dir();
+            if (!home.empty()) {
+                return fs::path(home) / ".puppetlabs" / "pxp-agent";
```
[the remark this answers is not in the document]
MikaelSmith: Consistent with https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppet-agent-non-root. Which I need to put a PR against.

lib/tests/unit/configuration_test.cc
MikaelSmith: Ok.

lib/tests/unit/configuration_test.cc
MikaelSmith: Ok.

lib/src/configuration.cc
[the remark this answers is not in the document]
mruzicka: Ok.

lib/src/configuration.cc
```diff
@@ -703,54 +757,65 @@ void Configuration::setDefaultValues()
...
+    hocon::shared_config config;
     try {
-        config_json = lth_jc::JsonContainer(lth_file::read(config_file_));
-    } catch (lth_jc::data_parse_error& e) {
+        LOG_TRACE("Parsing {1}", config_file_);
+        config = hocon::config::parse_file_any_syntax(config_file_)->resolve();
+    } catch (hocon::config_exception& e) {
         throw Configuration::Error {
-            lth_loc::translate("cannot parse config file; invalid JSON") };
+            _("cannot parse config file: {1}", e.what()) };
```
mruzicka: What is the difference between `_()` and `lth_loc::format()`?

mruzicka: I see both of them bing used throughout the file.

MikaelSmith: One's an alias for the other. I can revert to format to be consistent.

mruzicka: Ha, not much!
https://github.com/puppetlabs/leatherman/blob/master/locale/inc/leatherman/locale/locale.hpp#L159-L163

mruzicka: I don't mind using either but it's less confusing when only one is used consistently.
So I'm +1 for consistency.

lib/src/configuration.cc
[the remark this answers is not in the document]
MikaelSmith: Actually, these are slightly inconsistent with puppet/mcollective/facter, which recreate the Linux dir structure. i.e. `~/.puppetlabs/etc/pxp-agent`, not `~/.puppetlabs/pxp-agent/etc`. I'll work on updates tomorrow.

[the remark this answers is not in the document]
MikaelSmith: Proposed paths at https://github.com/puppetlabs/puppet-specifications/pull/99.

PR #661, opened 2017-09-19
(maint) Prepare for rapidjson 1.1.0

MikaelSmith:
Update to support both rapidjson 1.0.2 and 1.1.0.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -11,6 +11,10 @@
 #include <boost/system/error_code.hpp>
 
 #include <rapidjson/rapidjson.h>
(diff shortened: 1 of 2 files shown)


lib/src/modules/task.cc
```diff
@@ -11,6 +11,10 @@
 #include <boost/system/error_code.hpp>
 
 #include <rapidjson/rapidjson.h>
+#if RAPIDJSON_MAJOR_VERSION >= 1 && RAPIDJSON_MINOR_VERSION >= 1
```
mruzicka: I'd suggest changing the condition to:
```c++
#if RAPIDJSON_MAJOR_VERSION > 1 || RAPIDJSON_MAJOR_VERSION == 1 && RAPIDJSON_MINOR_VERSION >= 1
```
as the original condition is not satisfied for x.0 (2.0, 3.0, etc.).

MikaelSmith: Fixed.

PR #665, opened 2017-09-22
(maint) Ensure EntryBase is correctly destroyed

MikaelSmith:
Previously members of EntryBase would not be correcty destructed when an
Entry is destroyed. Add a virtual destructor to ensure that happens.

### DIFF
--- a/lib/inc/pxp-agent/configuration.hpp
+++ b/lib/inc/pxp-agent/configuration.hpp
@@ -54,6 +54,8 @@ struct EntryBase
               help { std::move(_help) },
               type { std::move(_type) } {
     }
+
+    virtual ~EntryBase() = default;
 };
 
 template <typename T>


lib/inc/pxp-agent/configuration.hpp
```diff
@@ -67,6 +69,8 @@ struct Entry : EntryBase
             : EntryBase { std::move(_name), std::move(_aliases), std::move(_help), std::move(_type) },
               value { std::move(_value) } {
     }
+
+    ~Entry() override = default;
```
mruzicka: According to [my testing](https://gist.github.com/mruzicka/c527df43e92a578bc56e8bee7788ae58) this does not seem to be necessary as that is the what the compiler does by default. I.e. it generates a default destructor and since the destructor is declared virtual in the base class it must override that from the base class.

MikaelSmith: Ah cool, I'll remove it.

MikaelSmith: Removed.

PR #666, opened 2017-09-22
(maint) Fix acceptance teardown on Cisco NX

MikaelSmith:
On Cisco NX we don't login as root. However, pxp-agent runs as root and
restricts permissions on the logdir to root owner and group.

Ensure logdir is readable by login user before attempting to copy it
back.


McdonaldSeanp:
👍

PR #667, opened 2017-09-26
(PCP-511) Include locale for cpp-hocon

MikaelSmith:
If cpp-hocon is built with Boost.Locale support, we need to also use
the locale library when linking it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,8 +51,18 @@ if (WIN32)
 endif()
 
 find_package(Leatherman REQUIRED COMPONENTS ${LEATHERMAN_COMPONENTS})
-find_package(Boost 1.54 REQUIRED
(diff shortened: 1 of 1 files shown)


Magisus:
Isn't Boost.Locale not available on some OSes (e.g. Solaris)? Will this still build on those?

MikaelSmith:
Probably safer to exclude it; I think it's available but would be unused.

MikaelSmith:
Updated.

MikaelSmith:
Not clear this fixes it yet. My Debian build failed.

MikaelSmith:
The CMake config is behaving strangely.

MikaelSmith:
Works now.

Magisus:
Ah yeah, moving the `include` makes sense.

PR #670, opened 2017-09-26
(PCP-511) Add iconv dependency for libpxp-agent

MikaelSmith:
libpxp-agent now uses libcpp-hocon, which introduces a dependency on
Boost.Locale. On macOS that requires linking iconv. Include it as a
public dependency of libpxp-agent so it's incorporated into binaries
that link against it (pxp-agent and pxp-agent-unittests).


MikaelSmith:
pxp-agent built successfully on macOS 10.12 with this change.

PR #675, opened 2017-10-06
(PCP-792) acceptance test for pxp-agent as non-root user

zreichert:


zreichert:
@MikaelSmith  
When I try to start pxp-agent here is what I see 
`=> #<Beaker::Result:0x007f9d9d72aff0
 @cmd="su foo -c \"/opt/puppetlabs/puppet/bin/pxp-agent\"",
 @exit_code=3,
 @host="redhat7-64-2",
 @output="",
 @raw_output="",
 @raw_stderr="",
 @raw_stdout="",
 @stderr="",
 @stdout="">`

Here is the configuration file that is in the users home directory 
`[foo@o12po6xafk9gfv9 home]$ cat /home/foo/.puppetlabs/etc/pxp-agent/modules/pxp-agent.conf
{"broker-ws-uri":"wss://m9py0m9hjggpj1x.delivery.puppetlabs.net:8142/pcp2/","pcp-version":"2","loglevel":"debug","ssl-key":"/home/foo/.puppetlabs/etc/puppet/ssl/private_keys/o12po6xafk9gfv9.delivery.puppetlabs.net.pem","ssl-ca-cert":"/home/foo/.puppetlabs/etc/puppet/ssl/certs/ca.pem","ssl-cert":"/home/foo/.puppetlabs/etc/puppet/ssl/certs/o12po6xafk9gfv9.delivery.puppetlabs.net.pem"}`

MikaelSmith:
Looks like it needs an update for Michal's move to HOCON.

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
```diff
@@ -0,0 +1,59 @@
...
+  step 'create non-root user on all nodes' do
+    on(agents, puppet("resource user foo ensure=present managehome=true"))
+
+    #TODO fix this only one agent ???
+    @user_home_dir = YAML.load(on(agents.first, puppet("resource user foo --to_yaml")).stdout)['user']['foo']['home']
+    @user_puppetlabs_dir = "#{@user_home_dir}/.puppetlabs"
+    @user_puppet_dir = "#{@user_puppetlabs_dir}/etc/puppet"
+    @user_pxp_module_dir = "#{@user_puppetlabs_dir}/etc/pxp-agent/modules"
+    on(agents, "mkdir -p #{@user_puppet_dir}")
+    on(agents, "mkdir -p #{@user_pxp_module_dir}")
```
MikaelSmith: You shouldn't need to create this directory.

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
```diff
@@ -0,0 +1,59 @@
...
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+
+      ssl_config = {
+          :ssl_key        => "#{@user_puppet_dir}/ssl/private_keys/#{agent}.pem",
+          :ssl_ca_cert    => "#{@user_puppet_dir}/ssl/certs/ca.pem",
+          :ssl_cert       => "#{@user_puppet_dir}/ssl/certs/#{agent}.pem"
+      }
+
+      create_remote_file(agent, "#{@user_pxp_module_dir}/pxp-agent.conf", pxp_config_json(master, agent, ssl_config).to_s)
```
MikaelSmith: Config should go in `~/.puppetlabs/etc/pxp-agent/pxp-agent.conf`. Right now you're instead putting it in the modules subdirectory.

zreichert:
@MikaelSmith  Thanks for the feedback, test passes now.

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
```diff
@@ -0,0 +1,59 @@
...
+    @user_home_dir = YAML.load(on(agents.first, puppet("resource user foo --to_yaml")).stdout)['user']['foo']['home']
+    @user_puppetlabs_dir = "#{@user_home_dir}/.puppetlabs"
+    @user_puppet_dir = "#{@user_puppetlabs_dir}/etc/puppet"
+    @user_pxp_dir = "#{@user_puppetlabs_dir}/etc/pxp-agent"
+    on(agents, "mkdir -p #{@user_puppet_dir}")
+    on(agents, "mkdir -p #{@user_pxp_dir}")
+  end
+
+  teardown do
+    on(agents, "kill -9 `pidof pxp-agent`", :accept_all_exit_codes => true)
```
MikaelSmith: Will this work everywhere?

MikaelSmith: At the least, pidof is not present on AIX.

MikaelSmith: Can test against AIX 7.1 by:
- requesting an nspooler instance
- using it in `be rake ci:test:aio SHA=... TEST_TARGET=aix71-POWERa{hypervisor=none,hostname=<hostname>}`

MikaelSmith: `get_process_pids` can probably be used as a replacement. It returns an array of matching PIDs.

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
```diff
@@ -0,0 +1,59 @@
...
+
+test_name  'Run puppet agent as non-root' do
+
+  platform = agents.first.platform
+  skip_test "Test is not compatible with #{platform}" if platform =~ /windows/
+
+  step 'create non-root user on all nodes' do
+    on(agents, puppet("resource user foo ensure=present managehome=true"))
+
+    #TODO fix this only one agent ???
```
MikaelSmith: Plan to address this?

zreichert: It seems to be a pattern that we use in all the other tests, however I didn't see anywhere that we specify more than one agent in a test configuration.

MikaelSmith: We don’t. It was broken for awhile and no one noticed.

zreichert: Sorry @MikaelSmith Im not sure I understand, are you saying that we don't ever use more than one agent?

I can make the test loop through all the agents if you feel its valuable, but if not we can just address the target with the helper method `agent`

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
zreichert: fixed

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
MikaelSmith: I was saying we don't ever test with more than one agent in practice. It's supported by Beaker and the test suite, but never exercised outside people doing it from their laptops.

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
MikaelSmith: Let me know when you've pushed the fix...

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
zreichert: In that case I'll put the loop in.

acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
zreichert: pushed

PR #676, opened 2017-10-09
(PCP-805) Enable testing system openssl linked PA in CI pipelines on rhel7

jsane:
Openssl versions on RHEL7 images being used by CI/VMpooler have an old version
which causes failures when running agent setup and test scripts. They need to be
upgraded to a desired version that is >= that on Centos7 used to build PA.

Related to PA-1580 (part of epic PA-1608). Has been test in PA CI - logs would be attached to PR for PA-1264 (enables linking PA against system openssl).
This does not impact regular agent build using vendored openssl which disregard system
openssl.

### DIFF
--- a/acceptance/setup/aio/010_Install.rb
+++ b/acceptance/setup/aio/010_Install.rb
@@ -12,6 +12,15 @@ def install_from_local(hosts, package)
       scp_to(agent, package, tmpdir)
     end
 
+    if agent[:platform].match(/(?:el-7|redhat-7)/)
+      step "Upgrade openssl package...(" + agent[:platform] + ")" do
+      end
+      on(agent, "yum -y install openssl-1.0.1e-51.el7_2.4.x86_64")
+    else
+      step "Skipping upgrade of openssl package...(" + agent[:platform] + ")" do
+      end
+    end
+
     step "Install package on #{agent}" do
       agent.install_package("#{tmpdir}/#{File.basename(package)}")
     end
@@ -31,6 +40,16 @@ step "Install puppet-agent..." do
     }
     agents.each do |agent|
       next if agent == master # Avoid SERVER-528
+
+      if agent[:platform].match(/(?:el-7|redhat-7)/)
+        step "Upgrade openssl package...(" + agent[:platform] + ")" do
+        end
+        on(agent, "yum -y install openssl-1.0.1e-51.el7_2.4.x86_64")
+      else
+        step "Skipping upgrade of openssl package...(" + agent[:platform] + ")" do
+        end
+      end
+
       install_puppet_agent_dev_repo_on(agent, opts)
     end
   end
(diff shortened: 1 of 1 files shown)


acceptance/setup/aio/010_Install.rb
```diff
@@ -12,6 +12,15 @@ def install_from_local(hosts, package)
       scp_to(agent, package, tmpdir)
     end
 
+    if agent[:platform].match(/(?:el-7|redhat-7)/)
+      step "Upgrade openssl package...(" + agent[:platform] + ")" do
```
MikaelSmith: Your use of blocks here is strange. I'd suggest putting the `yum install` inside the block.

acceptance/setup/aio/010_Install.rb
```diff
@@ -12,6 +12,15 @@ def install_from_local(hosts, package)
...
+      end
+      on(agent, "yum -y install openssl-1.0.1e-51.el7_2.4.x86_64")
+    else
+      step "Skipping upgrade of openssl package...(" + agent[:platform] + ")" do
```
MikaelSmith: No need for a block here, you can leave off do/end.

acceptance/setup/aio/010_Install.rb
MikaelSmith: Same below as well.

acceptance/setup/aio/010_Install.rb
```diff
@@ -12,6 +12,15 @@ def install_from_local(hosts, package)
       scp_to(agent, package, tmpdir)
     end
 
+    if agent[:platform].match(/(?:el-7|redhat-7)/)
+      step "Upgrade openssl package...(" + agent[:platform] + ")" do
+      end
+      on(agent, "yum -y install openssl-1.0.1e-51.el7_2.4.x86_64")
```
MikaelSmith: Is there any way we can avoid hard-coding this? Is it an upgrade or downgrade?

At least make it overridable with an environment variable.

jsane: Yes that has been my intent/preference as well. I just wanted to get it to run thru CI successfully once and that time I was not sure if any env variables set at the time of launching the Step1 would be visible to any downstream steps (which needed some experimentation). In any case will check and do it if it possible.

acceptance/setup/aio/010_Install.rb
```diff
@@ -12,6 +12,20 @@ def install_from_local(hosts, package)
       scp_to(agent, package, tmpdir)
     end
 
+   # Move the openssl libs package to a newer version on redhat platforms
+   use_system_openssl = ENV['USE_SYSTEM_OPENSSL']
+
+   if use_system_openssl && agent[:platform].match(/(?:el-7|redhat-7)/)
```
MikaelSmith: Repeated 3 times, seems worth pulling this whole thing out into a helper function.

MikaelSmith:
If you want to get this merged and do some of the refactoring as a separate PR, I could try to do that.

jsane:
Sorry I should have had closed this PR earlier as I indicated in the ticket. Basically once we have proper support for FIPS platforms in CI and VMpooler etc., for which there is consensus, this would not be required.

PR #677, opened 2017-10-12
Revert "(PCP-792) acceptance test for pxp-agent as non-root user"

MikaelSmith:
Reverts puppetlabs/pxp-agent#675


MikaelSmith:
Still broke on a number of platforms, so reverting. /cc @zreichert

PR #678, opened 2017-10-13
(PCP-792) test for pcp as non root agent

zreichert:
- Address failures on solaris and OSX
- Address test hang on debian 7 redhat 5 redhat 6 ubuntu 14.04

### DIFF
--- a/acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
+++ b/acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
@@ -1,0 +1,63 @@
+require 'pxp-agent/test_helper.rb'
+require 'yaml'
+
+test_name  'Run puppet agent as non-root' do
+
+  agents.each do |agent|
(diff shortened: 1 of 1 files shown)


acceptance/tests/pxp-module-puppet/run_puppet_agent_non_root.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    step 'Copy certs and keys to new users home directory' do
+      puppet_ssldir = on(agent, puppet('config print ssldir')).stdout.chomp
+      #ssl-key ssl-ca-cert ssl-cert
+      on(agent, "cp -R #{puppet_ssldir} #{@user_puppet_dir}")
+    end
+
+    step 'Attempt start of pxp-agent as non-root user' do
+      on( agent, puppet('resource service pxp-agent ensure=stopped'))
```
MikaelSmith: Why stop it twice? line 30 does the same thing.

PR #683, opened 2017-11-08
(PCP-809) Purge old cached tasks

MikaelSmith:
Refactor code to reduce duplication, fix error message. Update README to include task-cache-dir. Switch to format_n for pluralization.

Switch to using min of 1h or spool-dir-purge-ttl. Add task-cache-dir-purge-ttl configuration option.

Purge cached tasks older than task-cache-dir-purge-ttl using a thread similar to spool dir purging. The purging thread uses a mutex to check for shutdown.

A separate mutex is used to ensure a directory is not purged while it is also being used - i.e. ensuring the cache directory exists and has an updated last modified time happens under the mutex. After the mutex releases, purge should be prevented by the updated last modified time. The mutex will block downloading tasks, so it's acquired only while creating+updating the directory or checking+deleting it.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -1017,17 +1017,11 @@ void Configuration::validateAndNormalizeOtherSettings()
             lth_loc::format("invalid spool-dir-purge-ttl: {1}", e.what()) };
     }
 
-    if (HW::GetFlag<int>("association-timeout") < 0)
-        throw Configuration::Error {
-            lth_loc::translate("association-timeout must be positive") };
-
-    if (HW::GetFlag<int>("association-request-ttl") < 0)
-        throw Configuration::Error {
-                lth_loc::translate("association-request-ttl must be positive") };
-
-    if (HW::GetFlag<int>("pcp-message-ttl") < 0)
-        throw Configuration::Error {
-            lth_loc::translate("association-request-ttl must be positive") };
+    for (auto msg_ttl : {"association-timeout", "association-request-ttl", "pcp-message-ttl"}) {
+        if (HW::GetFlag<int>(msg_ttl) < 0)
+            throw Configuration::Error {
+                lth_loc::format("{1} must be positive", msg_ttl) };
+    }
 }
 
 const Options::iterator Configuration::getDefaultIndex(const std::string& flagname)
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -138,7 +138,7 @@ static const std::string DEFAULT_PCP_ACCESS_FILE { (DEFAULT_LOG_DIR / "pcp-acces
 static const std::string DEFAULT_MODULES_CONF_DIR { (DEFAULT_CONF_DIR / "modules").string() };
 static const std::string DEFAULT_CONFIG_FILE { (DEFAULT_CONF_DIR / "pxp-agent.conf").string() };
 static const std::string DEFAULT_PCP_VERSION { "1" };
-static const std::string DEFAULT_SPOOL_DIR_PURGE_TTL { "14d" };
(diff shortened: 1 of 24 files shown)


lib/src/modules/task.cc
```diff
@@ -543,5 +574,84 @@ void Task::processOutputAndUpdateMetadata(ActionResponse& response)
...
+        "Removed {1} directories from '{2}'",
+        num_purged_dirs, num_purged_dirs, task_cache_dir_));
+    return num_purged_dirs;
+}
+
+//
+// Task cache directory purge task (private interface)
+//
+
+void Task::taskCacheDirPurgeTask()
```
MikaelSmith: With some distance, I may look at pulling these tasks and threading into a separate class that both Task and RequestProcessor pull in.

An alternative is to have RequestProcessor have a separate list of things that require purging (such as the Task) class and run through that list as part of the purge task there.

MikaelSmith: The purge functionality has a fair amount of duplication as well.

lib/src/request_processor.cc
```diff
@@ -915,38 +923,48 @@ void RequestProcessor::logLoadedModules() const
...
-                 "minutes; thread id {2}",
-                 num_minutes, pcp_util::this_thread::get_id());
+        return init;
     }
+}
+
+void RequestProcessor::purgeTask()
+{
+    // Use min of 1h and gcd of purgeable TTLs.
+    auto num_minutes = std::min(60u, minutes_gcd(purgeables_));
```
MikaelSmith: Using gcd might be overly aggressive here. Poor choice of TTLs could result in checking every minute, and we probably don't need to be that precise.

lib/src/modules/task.cc
```diff
@@ -355,7 +363,10 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
     LOG_DEBUG("Verifying task file based on {1}", file.toString());
 
     try {
-        auto cache_dir = createCacheDir(task_cache_dir, file.get<std::string>("sha256"));
+        auto cache_dir = [&]() {
+            pcp_util::lock_guard<pcp_util::mutex> the_lock { task_cache_dir_mutex };
+            return createCacheDir(task_cache_dir, file.get<std::string>("sha256"));
+        }();
         return updateTaskFile(master_uris, client, cache_dir, file);
```
nicklewis: Does this line need to be part of the critical section above? It seems possible for the purge and this code to run concurrently. In that case, the following sequence of operations could happen:

A) The purge function notices the task file needs to be deleted
B) The running task operation checks that the task file exists and sees it does
C) The purge function deletes the task file
D) The running task operation presumably explodes

MikaelSmith: Purge should only operate on directories. A side-effect of createCacheDir is to update mtime of the directory. We then download files into that directory. The purge should only: check whether the directory is old, and if so remove the directory. So I believe it behaves correctly.

lib/src/configuration.cc
```diff
@@ -1010,24 +1022,21 @@ void Configuration::validateAndNormalizeOtherSettings()
...
-        throw Configuration::Error {
-                lth_loc::translate("association-request-ttl must be positive") };
-
-    if (HW::GetFlag<int>("pcp-message-ttl") < 0)
-        throw Configuration::Error {
-            lth_loc::translate("association-request-ttl must be positive") };
+    for (auto msg_ttl : {"association-timeout", "association-request-ttl", "pcp-message-ttl"}) {
+        if (HW::GetFlag<int>(msg_ttl) < 0)
+            throw Configuration::Error {
+                lth_loc::translate("{1} must be positive", msg_ttl) };
```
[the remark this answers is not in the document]
MikaelSmith: Good catch, thanks.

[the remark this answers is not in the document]

PR #684, opened 2017-11-14
(PCP-822) Allow stdin to be ignored

MikaelSmith:
Allow tasks to ignore stdin. The default mode is to pass arguments on
stdin and as environment variables, and tasks can choose to use either.
Ensure that if they choose not to use stdin, execution still succeeds if
the task exits before we write stdin to the appropriate pipe (that has
been closed).

Requires Leatherman 1.2.2 or 1.4+.


MikaelSmith:
I'm running this in a loop with `tests/tasks/run_echo.rb` to check for failures. Going to call it good after an hour with no failure.

MikaelSmith:
Still passing after 2 hours. 👍

PR #686, opened 2017-11-21
(PCP-801) Use kill -15 rather than kill -s TERM for broader support

MikaelSmith:
The default shell on Solaris 10 (SPARC) doesn't support `kill -s TERM`,
so use `kill -15` instead. This is more consistent with other `kill`
calls used in acceptance tests as well.

[skip ci]

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -83,6 +83,7 @@ def kill_all_pcp_brokers(host)
   on(host, "ps -C java -f | grep pcp-broker | sed 's/[^0-9]*//' | cut -d\\  -f1") do |result|
     pids = result.stdout.chomp.split("\n")
     pids.each do |pid|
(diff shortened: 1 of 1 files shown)


acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -475,7 +475,7 @@ def stop_sleep_process(targets, seconds_to_sleep, accept_no_pid_found = false)
     pids.each do |pid|
       target['platform'] =~ /win/ ?
         on(target, "taskkill /F /pid #{pid}") :
-        on(target, "kill -s TERM #{pid}")
+        on(target, "kill -15 #{pid}")
```
nicklewis: Can you add a comment explaining `-15` is `TERM`?

MikaelSmith: Done.

ScottGarman:
:+1:

MikaelSmith:
@nicklewis updated, waiting on your 👍

PR #687, opened 2017-11-29
(PCP-763) clear task-cache

zreichert:


zreichert:
Tests failing with the following setup:

export SHA=4bcaaea11510530543f138b8ed575d2e687b13da
export TESTS=tests/tasks/remove_old_task_from_cache.rb
export BEAKER_HOSTS=hosts.yaml
bundle exec beaker-hostgenerator redhat7-64m-redhat7-64a --hypervisor abs --disable-default-role --osinfo-version 1
bundle exec rake ci:test:aio

@adreyer @nicklewis @mruzicka  any thoughts?

nicklewis:
@zreichert Do you have output for what is failing?

zreichert:
@nicklewis the assert 
`assert_equal(result.exit_code, 1, 'The cached file was not purged on restart of pxp-agent')`
is failing because the exit code is 0 from the command `test -f <path_to_file>`

I expected that file would be removed.

acceptance/tests/tasks/task_purged_from_cache.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  step 'Update access time on file and restart pxp-agent to trigger cache purge' do
+    agents.each do |agent|
+      date = DateTime.now - 4
+      on(agent, "ls -l $(find #{cache_path} -type f )") do |result|
+        @target_file_path = result.stdout.match(/#{Regexp.escape(cache_path)}\S*/)[0]
+      end
+      on(agent, "touch --date #{date} #{@target_file_path}")
```
nicklewis: The cleanup check is based on directory mtime, not file. So that's what we should be touching.

PR #694, opened 2018-01-09
(maint) Remove brittle test

MikaelSmith:
Remove restart_host_run_puppet test. It takes a long time, is brittle,
and has never exposed actual errors.

[skip ci]


james-stocks:
There was PCP-585 affecting Fedora; but this test was subsequently confined from running on Fedora. I guess that bug can still be fix-tested if/when needed by cherry-picking this test from git on a test setup.

PR #696, opened 2018-01-18
(RE-9936) Migrate from stickler to artifactory

sarameisburger:


sarameisburger:
@MikaelSmith new PR pointing at 1.5.x

PR #700, opened 2018-02-09
(RE-9880) Update acceptance tests that reference nightlies

mwaggett:
This commit changes the name of the nightlies server to the actual hostname so that when we remove the 'nightlies' cname (to use for *new* nightlies), tests won't break. THIS SHOULD NOT BE PERMANENT (see RE-10231).


MikaelSmith:
This seems ok. @McdonaldSeanp anything I should consider in recent changes to puppet-agent CI?

MikaelSmith:
Oh, we should probably target 1.5.x, that's still an active branch.

mwaggett:
😊 yup, started there and didn't find anything relevant to this

PR #705, opened 2018-02-18
(FACT-1772) Use CMAKE_INSTALL_LIBDIR for multilib install

ekinanp:
This allows the libpxp-agent library to be installed in e.g.
/usr/lib, /usr/lib64, etc. depending on the platform that it
is being built on.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,9 +38,11 @@ if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
 endif()
 
 # Set RPATH if not installing to a system library directory
-list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" INSTALL_IS_SYSTEM_DIR)
+# TODO: Once a new version of Leatherman is released, update this
+# to use the set_cmake_install_rpath macro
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -38,9 +38,9 @@ if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
 endif()
 
 # Set RPATH if not installing to a system library directory
-list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" INSTALL_IS_SYSTEM_DIR)
+list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" INSTALL_IS_SYSTEM_DIR)
```
ekinanp: This pattern is used in several places. Here, in https://github.com/puppetlabs/cpp-pcp-client/pull/234/files, and https://github.com/puppetlabs/facter/pull/1698/files. I wonder if it would be useful to make it a macro in leatherman, something like `leatherman_set_cmake_install_rpath`?

MikaelSmith: Yeah, would be useful.

ekinanp:
Blocked on https://github.com/puppetlabs/leatherman/pull/273

PR #708, opened 2018-04-04
Maint/1.8.x/merge up from 1.5.x

melissa:


melissa:
@MikaelSmith as far as I can tell, the only functionality we're loosing is https://github.com/puppetlabs/pxp-agent/commit/570ee58f1227064faad611ba016b5bd05f98d575 

Is that something you're using in the ci pipeline? Would that functionality be something you'd like to see for all components?

MikaelSmith:
It hasn't been heavily used, I'm ok with dropping it for now.

PR #709, opened 2018-04-04
(maint) Remove unused files and functions

MikaelSmith:


MikaelSmith:
jenkins please test this

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -6,32 +6,5 @@ def ruby_command(host)
       end
       module_function :ruby_command
```
melissa: we have ruby_command defined in beaker-puppet. We might be able to use the method defined there rather than this one

https://github.com/puppetlabs/beaker-puppet/blob/master/lib/beaker-puppet/install_utils/foss_utils.rb#L1406

MikaelSmith: Great, I was wondering about that.

MikaelSmith: Done.

MikaelSmith:
jenkins please test this

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -546,6 +545,6 @@ def check_puppet_non_blocking_response(identity, transaction_id, max_retries, qu
 end
 
 def get_mode(host, path)
-  ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+  ruby = ruby_command(host)
```
melissa: Does this work? When I use this command in beaker-puppet, I had to include `extend Beaker::DSL::InstallUtils::FOSSUtils` in the file

MikaelSmith: It seemed to. I'm not exactly sure why.

MikaelSmith: Also, why isn't jenkins testing?

MikaelSmith:
Test run passes locally for me.

acceptance/lib/pxp-agent/test_helper.rb
melissa: This totally works locally, even though I have no idea where the method is getting loaded. It's definitely calling the right one from beaker-puppet though, so I suppose this is good

PR #711, opened 2018-04-23
(PCP-797) Add task-download-connect-timeout, task-download-timeout

MikaelSmith:
Configure task download timeouts based on new options `task-download-connect-timeout` and `task-download-timeout`.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -120,12 +120,16 @@ Task::Task(const fs::path& exec_prefix,
            const std::string& ca,
            const std::string& crt,
            const std::string& key,
+           uint32_t task_download_connect_timeout,
+           uint32_t task_download_timeout,
            std::shared_ptr<ResultsStorage> storage) :
     Purgeable { task_cache_dir_purge_ttl },
     storage_ { std::move(storage) },
     task_cache_dir_ { task_cache_dir },
     exec_prefix_ { exec_prefix },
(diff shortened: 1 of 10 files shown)


MikaelSmith:
I'm looking into how to test this. I can probably setup something that doesn't respond to connection attempts or stops sending after the connection handshake, but not sure what will be easy to work into tests yet.

MikaelSmith:
One option for testing was to add 1s delay to all traffic to the master node via
```
tc qdisc add dev eth0 root netem delay 1000ms
```
(reset via `tc qdisc del dev eth0 root netem`). Got
```
2018-04-23 16:27:25.238692 WARN  puppetlabs.pxp_agent.modules.task:301 - Downloading the task file from the master-uri 'https://puppet:8140' failed. Reason: File download server side error: Connection timed out after 1006 milliseconds
```
with both `task-download-connect-timeout: 1` and `task-download-timeout: 1`. That's probably not a good thing to put in tests though.

lib/src/modules/task.cc
```diff
@@ -245,6 +249,11 @@ static std::string createUrlEndpoint(const lth_jc::JsonContainer& uri) {
     return url;
 }
 
+struct DownloadConfig {
```
nicklewis: This struct isn't used.

nicklewis:
Seems like this works in my testing.

PR #714, opened 2018-06-21
(PA-2055) Update boost math header to 'integer'

McdonaldSeanp:
The boost/math/common_factor_rt header has changed to
boost/integer/common_factor_rt in boost 1.67.0. This commit updates pxp-agent
to reference the new header and namespace

### DIFF
--- a/lib/src/request_processor.cc
+++ b/lib/src/request_processor.cc
@@ -28,7 +28,7 @@
 
 #include <boost/filesystem/operations.hpp>
 #include <boost/format.hpp>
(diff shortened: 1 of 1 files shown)


MikaelSmith:
This is going to be a fun change...

Another PR will be updating how Travis CI builds. We can look at making sure we have compatible versions of Boost for AppVeyor and Travis when that gets merged.

MikaelSmith:
Looks like Boost 1.58 will be new enough, so should be easy after we update travis builds.

MikaelSmith:
Planning to update this?

PR #716, opened 2018-06-28
(maint) Increase timeout for task_purged_from_cache test

caseywilliams:
Doubles the timeout for task_purged_from_cache to allow for slower
systems in CI


caseywilliams:
This is intended to address timeouts we're seeing in puppet-agent#master, so PR testing won't really be testing the right things, but [here](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/192/) is an ad-hoc run of master with this change cherry picked in pxp-agent#master

PR #718, opened 2018-08-14
(PCP-835) Add support for task implementations

MikaelSmith:
Select implementation that matches known features, initially only `puppet-agent`. Uses first file in `files` if no implementations are provided. Errors if implementations are provided but features do not satisfy any of their requirements.

### DIFF
--- a/acceptance/tests/tasks/task_implementations.rb
+++ b/acceptance/tests/tasks/task_implementations.rb
@@ -1,0 +1,65 @@
+require 'pxp-agent/task_helper.rb'
+require 'pxp-agent/config_helper.rb'
+require 'puppet/acceptance/environment_utils.rb'
+
+test_name 'task downloads correct implementation' do
+
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  PUPPETSERVER_CONFIG_FILE = windows?(master) ?
+    '/cygdrive/c/ProgramData/PuppetLabs/puppetserver/etc/conf.d/webserver.conf'
+  : '/etc/puppetlabs/puppetserver/conf.d/webserver.conf'
+
+  step 'Ensure each agent host has pxp-agent running and associated' do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_hocon_using_puppet_certs(master, agent))
+      on agent, puppet('resource service pxp-agent ensure=running')
+
+      assert(is_associated?(master, "pcp://#{agent}/agent"),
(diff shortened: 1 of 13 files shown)


MikaelSmith:
I'll plan to add an acceptance test around making sure we download the right files.

MikaelSmith:
jenkins, please test this

MikaelSmith:
jenkins please test this

MikaelSmith:
jenkins please test this

MikaelSmith:
jenkins please test this on aix71-POWER,osx1012-64,redhat7-64,sles12-64,solaris10-32,solaris11-64,windows2012r2-64

MikaelSmith:
jenkins please test this on aix61-POWER,osx1012-64,redhat7-64,sles12-64,solaris10-32,solaris11-64,windows2012r2-64

MikaelSmith:
jenkins please test this on redhat7-64,windows2012r2-64

MikaelSmith:
I can't reproduce the run_ps1 failure, I don't think it's caused by my changes. Encountering errors starting the service with Puppet.

MikaelSmith:
jenkins please test this on redhat7-64a,windows2012r2-64a

MikaelSmith:
So turns out the `test on` logic is very literal. If we don't include agents, must of the tests do nothing. The run_ps1 test was poorly written, so it tries to run pxp-agent on non-agent Windows targets. That works generally because Windows is never a master platform, but fails if the windows hosts aren't agents.

MikaelSmith:
Acceptance tests passed.

acceptance/tests/tasks/task_implementations.rb
```diff
@@ -0,0 +1,65 @@
...
+    # Make the static content path
+    env_name = File.basename(__FILE__, '.*')
+    @static_content_path = File.join(environmentpath, mk_tmp_environment(env_name))
+
+    # Create the task
+    task_body = "#!/opt/puppetlabs/puppet/bin/ruby\nputs ENV['PT_message']"
+    taskpath = "#{@static_content_path}/#{filename}"
+    create_remote_file(master, taskpath, task_body)
+    on master, "chmod 1777 #{taskpath}"
+    @sha256 = Digest::SHA256.hexdigest(task_body + "\n")
```
nicklewis: Where does the `\n` come from in the task body?

MikaelSmith: I think that's automatically created on the target with `create_remote_file`. This is just copied around at this point.

nicklewis: :+1:

PR #719, opened 2018-08-16
(maint) Fix tests that only worked when agents are all windows or not

MikaelSmith:
Fix tests so they work with a mixed set of agents - the default when
running `rake ci:test:aio` - rather than only working when agents are
all windows or all non-windows.

### DIFF
--- a/acceptance/tests/proxy_connection.rb
+++ b/acceptance/tests/proxy_connection.rb
@@ -87,6 +87,7 @@ test_name 'Connect via proxy' do
       assert(is_associated?(master, "pcp://#{agent}/agent"),
              "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
     end
+    clear_squid_log(master)
   end
 
   step 'Download and run the task on agent hosts via proxy' do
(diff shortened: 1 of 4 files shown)


MikaelSmith:
jenkins please test this on redhat7-64a,windows2012r2-64a

acceptance/tests/proxy_connection.rb
```diff
@@ -87,6 +87,7 @@
       assert(is_associated?(master, "pcp://#{agent}/agent"),
              "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
     end
+    clear_squid_log(master)
```
donoghuc: So that is why there were always two entries in the log! Yes this makes sense.

MikaelSmith: Maybe we should also be verifying there's an entry here (for the PCP connection).

MikaelSmith: I guess we do that as a separate test above.

MikaelSmith:
jenkins please test this

donoghuc:
jenkins please test this on redhat7-64a,windows2012r2-64a

MikaelSmith:
jenkins please test this on redhat7-64a,windows2012r2-64a

MikaelSmith:
This won't work until tests are updated to beaker 4.

PR #720, opened 2018-09-05
(PCP-849) Set _task metaparameter to the task name

nicklewis:
This metaparameter allows a single executable to be used to provide many
tasks, enabling easier code reuse between similar tasks.

This also updates the Powershell shim to only pass along parameters which are understood by the task, to avoid failing when a task doesn't expect the new implicit *_task* parameter.

### DIFF
--- a/exe/PowershellShim.ps1
+++ b/exe/PowershellShim.ps1
@@ -5,17 +5,21 @@ try {
   $here = (Split-Path -Parent $MyInvocation.MyCommand.Path)
   . $here\PowershellShim-Helper.ps1
 
-  $scriptArgsHash = Get-ContentAsJson ($input -join "")
+  $private:tempArgs = Get-ContentAsJson ($input -join "")
 
   Write-Debug "`nTask-shim deserialized arguments to Hashtable:`n"
 
-  if ($scriptArgsHash) {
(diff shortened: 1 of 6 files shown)


nicklewis:
This PR is just so I can test in appveyor, for now.

nicklewis:
This should be ready to go now.

exe/PowershellShim.ps1
```diff
@@ -5,17 +5,21 @@ try {
...
       Write-Debug "* $($_.Key) ($($_.Value.GetType())):`n$($_.Value | ConvertTo-String)"
     }
   }
 
-  & $args[0] @scriptArgsHash
+  $allowedArgs = (Get-Command $args[0]).Parameters.Keys
+  $private:taskArgs = @{}
+  $private:tempArgs.Keys | ? { $allowedArgs -contains $_ } | % { $private:taskArgs[$_] = $private:tempArgs[$_] }
+
+  & $args[0] @taskArgs
```
MikaelSmith: Can we look at adding a Pester test? See https://github.com/puppetlabs/pxp-agent/blob/master/exe/README.md.

nicklewis: Updated with my best swing at a Pester test.

PR #721, opened 2018-09-06
(PCP-836) Add powershell and shell features

MikaelSmith:
Add the shell feature on non-Windows hosts, and add the powershell
feature if a powershell executable is found.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -80,6 +80,12 @@ static const std::string TASK_RUN_ACTION_INPUT_SCHEMA { R"(
         }
       }
     },
+    "features": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
     "files": {
       "type": "array",
       "items": {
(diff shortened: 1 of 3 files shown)


lib/src/modules/task.cc
```diff
@@ -174,6 +180,13 @@ Task::Task(const fs::path& exec_prefix,
     client_.set_client_cert(crt, key);
     client_.set_supported_protocols(CURLPROTO_HTTPS);
     client_.set_proxy(proxy);
+
+#ifndef _WIN32
+    features_.insert("shell");
+#endif
+    if (!lth_exec::which("powershell").empty()) {
```
nicklewis: Do powershell tasks work on non-windows systems? I'm unclear whether the "built-in task interpreter" for ps1 on windows does anything that would be necessary elsewhere. Specifically, it passes `-NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass`.

MikaelSmith: I'm not sure those arguments are needed with powershell on other platforms. They're present in the powershell 6 beta, but that might just be so scripts don't break.

MikaelSmith: Hard-coded it so powershell is only matched on Windows.

MikaelSmith:
jenkins please test this

MikaelSmith:
Acceptance tests are still running off the 2nd commit, and I expect to be sufficient to validate the changes.

MikaelSmith:
Everything's green @nicklewis .

PR #722, opened 2018-09-07
(PCP-851) Download files required by task

donoghuc:
Add support for multi file tasks in PXP-agent. PXP-agent should find all files required for the chosen implementation of the task and download them from puppet-server into the correct directory structure. The `_installdir` pointing to the supporting files should be passed to the task executable as task metadata.


lib/src/modules/task.cc
```diff
@@ -416,9 +428,26 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
     }
 }
 
+static fs::path getMultiFiles(lth_jc::JsonContainer& meta_files, lth_jc::JsonContainer& impl_files, lth_jc::JsonContainer& files) {
+    // meta_files: download from the metadata level files array
```
donoghuc: @MikaelSmith does this logic make sense based on what we discussed?

lib/src/modules/task.cc
```diff
@@ -373,7 +385,7 @@ static fs::path updateTaskFile(const std::vector<std::string>& master_uris,
                                lth_curl::client& client,
                                const fs::path& cache_dir,
                                const lth_jc::JsonContainer& file) {
-    auto filename = file.get<std::string>("filename");
+    auto filename = fs::basename(fs::path(file.get<std::string>("filename")));
```
donoghuc: We chose to keep the cache dir "flat" and name files based on their base name. This will overwrite duplicate files but the complexity of maintaining a nested cache dir is not worth the minor performance issue of re-downloading and overwriting cached files.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,93 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
     }
 }
 
+// build a spool dir for the supporting library files requested by a multifile task
+static fs::path createInstallDir(const fs::path& spool_dir, std::set<std::string> download_set) {
```
donoghuc: It appears that the executable for a single file task is being run from it's location in the cache. I noticed that spool dir is configurable so I am assuming we want to use that. That will likely be a new parameter in the `Task` constructor. I want to verify that is the correct use of spool dir before altering the constructor.

lib/src/modules/task.cc
```diff
@@ -583,6 +680,14 @@ ActionResponse Task::callAction(const ActionRequest& request)
                                        client_,
                                        file);
 
+    auto meta_files = task_metadata.getWithDefault<std::vector<lth_jc::JsonContainer>>("files", {});
+    auto lib_files = getMultiFiles(meta_files, implementation.files, files);
+    if (lib_files.size() > 0) {
+        install_dir = downloadMultiFile(files, download_set);
+    }
+
+    // TODO pass install_dir as metaparameter _installdir for task execution
```
donoghuc: Will need to pass this location as _installd_dir to the task. Will also need to delete the install_dir after execution (including if the task fails).

donoghuc:
This commit is somewhat psuedo code at the moment. I just wanted to get the structure of the proposed update up while it is fresh in my head. I think it illustrates the general intent of what we want to do, but it for sure still needs work.

lib/src/modules/task.cc
MikaelSmith: It shouldn't overwrite duplicate files. If the file has the same hash, we shouldn't redownload it in the first place.

lib/src/modules/task.cc
MikaelSmith: I think the spool dir will already have been created. The directory for `_installdir` should be a subdirectory under whatever spooldir will be used for the execution of this action.

lib/src/modules/task.cc
MikaelSmith: We don't need to delete the install dir after execution. spooldir's normal cleanup will handle that.

lib/src/modules/task.cc
MikaelSmith: The `if` block at line 380/392 will take care of that.

lib/src/modules/task.cc
MikaelSmith: I think creating the install dir from downloaded files could happen outside of the `Task` class.

I think in `callAction`, you can get the results dir (which is what I mean by spool_dir) from the request object.

MikaelSmith: One twist is that we only create a resultsDir for a`RequestType::Blocking`. For non-blocking requests, we probably just want to create a tempdir for `_installdir`.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,93 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+std::set<std::string> getMultiFiles(std::vector<std::string> meta_files, std::vector<std::string> impl_files, lth_jc::JsonContainer& files) {
+    // build set of unique files to download and exclude the selected task executable
+    std::set<std::string> download_set(meta_files.begin(), meta_files.end());
+    download_set.insert(impl_files.begin(), impl_files.end());
+
+    // expand files from directories specified by trailing '/'
+    // this should replace ^^ with de-duplicated filenames for download
+    for(auto f_name : download_set) {
+        if(f_name[f_name.length() - 1] == '/') {
+            download_set.erase(f_name);
```
MikaelSmith: I don't think this is safe, as it will invalidate the iterator the range-based for-loop is using. See https://stackoverflow.com/questions/10360461/removing-item-from-vector-while-in-c11-range-for-loop.

I think I'd handle this by creating a new set that has all the files and returning that.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,93 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+            download_set.insert(f_name);
+        }
+    }
+
+    return download_set;
+}
+
+// iterate over unique set of filenames to support task
+// copy file from cache to install_dir
+static fs::path downloadMultiFile( lth_jc::JsonContainer& files, std::set<std::string> download_set) {
```
MikaelSmith: Looks like `downloadMultiFile` needs to take `task_spool_dir` as an argument (based on `request.resultsDir()`, with fallback to a temp directory).

lib/src/modules/task.cc
```diff
@@ -453,7 +549,8 @@ static Implementation selectImplementation(std::vector<lth_jc::JsonContainer> co
 
     if (impl != implementations.cend()) {
         auto input_method = impl->getWithDefault<std::string>("input_method", {});
-        return Implementation { impl->get<std::string>("name"), input_method };
+        auto files = impl->getWithDefault<std::string>("files", {});
```
MikaelSmith: This should be getting a vector of strings, i.e. `impl->getWithDefault<std::vector<std::string>>("files", {});`.

lib/src/modules/task.cc
MikaelSmith: The most performant option might be to extract a list of directories from download_set, then iterate over the new list to erase them from download_set and add anything from `files` that matches to download_set.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,93 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+static fs::path downloadMultiFile( lth_jc::JsonContainer& files, std::set<std::string> download_set) {
+    auto install_dir =  createInstallDir(task_spool_dir, download_set);
+    lth_jc::JsonContainer file_object;
+    fs::path lib_file;
+    for(auto file_name : download_set) {
+        // get file object info based on name
+        file_object = std::find_if(files.cbegin(), files.cend(),
+                                    [&](lth_jc::JsonContainer const& file) {
+                                        auto filename = file.get<std::string>("filename");
+                                        return filename == f;
```
MikaelSmith: I don't see an `f` here, must mean `file`.

lib/inc/pxp-agent/modules/task.hpp
```diff
@@ -73,6 +73,10 @@ class Task : public PXPAgent::Module, public PXPAgent::Util::Purgeable {
 
     leatherman::curl::client client_;
 
+    boost::filesystem::path downloadMultiFile(std::vector<leatherman::json_container::JsonContainer> const& files, std::set<std::string> const& download_set);
```
donoghuc: We have discussed removing this from the Task class.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,111 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+
+    fs::create_directories(install_dir);
+    fs::last_write_time(install_dir, time(nullptr));
+    fs::permissions(install_dir, NIX_DIR_PERMS);
+
+    // this should generate a unique collection of directory paths to append to install_dir
+    // for example:
+    // if files had the paths: /one/two/three.txt, /one/two/one.txt, /one/two.txt
+    // dir_structure set would contain: /one/two /one
+    std::set<fs::path> dir_structure;
```
donoghuc: This is currently not working. I think I need a different data structure bc set is un-ordered. When rebuilding path the expected order is not preserved.

lib/tests/unit/modules/task_test.cc
```diff
@@ -207,6 +207,32 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
...
+        REQUIRE(output == "{\"message\":\"hello\",\"_task\":\"test\"}");
+    }
+
+    SECTION("builds installdir") {
+        auto echo_txt =
+#ifndef _WIN32
+            (DATA_FORMAT % "\"0632\""
+                         % "\"task\""
+                         % "\"run\""
+                         % "{\"task\": \"test\", \"input_method\": \"environment\",  \"input\":{\"message\":\"hello\"}, \"metadata\": {\"files\": [\"file1.txt\", \"dir/\"]}, \"files\" : [{\"sha256\": \"c9b0792365dddb71eca56e722aa8913a8de1721c2b3ee5fdbd697a9da988ab3f\", \"filename\": \"init.sh\"}, {\"sha256\": \"67ee5478eaadb034ba59944eb977797b49ca6aa8d3574587f36ebcbeeb65f70e\", \"filename\": \"dir/file2.txt\"}, {\"sha256\": \"ecdc5536f73bdae8816f0ea40726ef5e9b810d914493075903bb90623d97b1d8\", \"filename\": \"file1.txt\"}]}").str();
```
donoghuc: Will clean up formatting and account for windows. 
I am adding the required resources in the src/test/resources/task-cache directory saved under their hashes. Will update once i get the trailing directory support added.

donoghuc:
jenkins please test this on redhat6-64,windows2012r2-64

acceptance/tests/tasks/run_multifile.rb
```diff
@@ -0,0 +1,43 @@
...
+      @sha256_file1 = create_task_on(agent, 'multi_file', 'file1.txt', file1_body)
+      @sha256_file2 = create_task_on(agent, 'multi_file', 'file2.txt', file2_body)
+      @sha256_file3 = create_task_on(agent, 'multi_file', 'file3.txt', file3_body)
+    end
+  end
+
+  step 'Run ruby task on agent hosts' do
+    files = [file_entry('init.rb', @sha256), file_entry('file1.txt', @sha256_file1), file_entry('dir/file2.txt', @sha256_file2), file_entry('dir/sub_dir/file3.txt', @sha256_file3)]
+    metadata = { implementations: [
+      { name: 'init.rb', requirements: ['puppet-agent'], files: ['file1.txt', 'dir/file2.txt', 'dir/sub_dir/file3.txt']}]}
```
MikaelSmith: Let's move one of these out to a metadata-level `files` entry.

acceptance/tests/tasks/run_multifile.rb
```diff
@@ -0,0 +1,43 @@
...
+    run_successful_task(master, agents, 'multi_file', files, metadata: ) do |stdout|
+      json, data = stdout.delete("\r").split("\n")
```
MikaelSmith: This doesn't seem quite right. Wouldn't `split` result in more than two entries?

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+// Get a unique list of "lib" files that the task has specified in metadata
+// this involves enumerating files when directories are requested
+std::set<std::string> getMultiFiles(std::vector<std::string> const& meta_files, std::vector<std::string>  const& impl_files, std::vector<lth_jc::JsonContainer> const& files) {
+    // build set of unique files to download and exclude the selected task executable
+    std::set<std::string> download_set(meta_files.begin(), meta_files.end());
+    download_set.insert(impl_files.begin(), impl_files.end());
+    std::vector<std::string> directories;
+
+    // expand files from directories specified by trailing '/'
+    for (auto f_name : download_set) {
```
MikaelSmith: I'm not sure what you'll end up iterating over when you're also inserting items. This may do more work than needed, which is why I suggested the first pass just add to `directories`, then do a pass over directories to add items to `download_set`.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+        fs::copy_file(lib_file, install_dir / fs::path(file_name));
+    }
+
+    return install_dir;
+}
+
+lth_jc::JsonContainer Task::selectLibFile(std::vector<lth_jc::JsonContainer> const& files,
+                                            std::string const& file_name)
+{
+    // Select file based on impl.
```
MikaelSmith: `based on impl` is unclear to me. What's the impl?

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+    auto file = std::find_if(files.cbegin(), files.cend(),
+        [&](lth_jc::JsonContainer const& file) {
+            auto filename = file.get<std::string>("filename");
+            return filename == file_name;
+        });
+
+    if (file == files.cend()) {
+        throw Module::ProcessingError {
+            lth_loc::format("'{1}' file requested by implementation not found", file_name) };
```
MikaelSmith: This error message doesn't seem very clear either. It can work, but implementation isn't obviously related in my mind.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+        auto lib_file = getCachedTaskFile(task_cache_dir_,
+                                       task_cache_dir_mutex_,
+                                       master_uris_,
+                                       task_download_connect_timeout_,
+                                       task_download_timeout_,
+                                       client_,
+                                       file_object);
+
+        // copy to expected location in install_dir
+        auto t = install_dir / fs::path(file_name);
```
MikaelSmith: You don't technically need to wrap `file_name` in `fs::path`, `fs::path`'s `/` operator will do that implicitly. You also create `t`, then don't use it. Seems like this line could be omitted.

MikaelSmith: Or do we need special handling because `file_name` can actually be a multisegment path?

lib/src/modules/task.cc
```diff
@@ -592,6 +706,20 @@ ActionResponse Task::callAction(const ActionRequest& request)
 
     task_params.set<std::string>("_task", task_name);
 
+    std::vector<std::string> meta_files;
+    // if metadata::files key not available catch error
+    try {
+         meta_files = task_metadata.getWithDefault<std::vector<std::string>>("files", {});
```
MikaelSmith: Decided not to rework this part? Maybe instead guard it with `task_metadata != task_execution_params`.

MikaelSmith: The comment doesn't feel clear to me either.

lib/tests/resources/tasks-cache/8fdd86d9be77a27b296d5b6f2d2d984fd7d671bc8bb1ffc86f6c6b0492ec4d01/init.bat
```diff
@@ -0,0 +1 @@
+type %PT__installdir\file1.txt% %PT__installdir\dir\file2.txt% %PT__installdir\dir\sub_dir\file3.txt% 2>nul
```
MikaelSmith: Oh, the problem here is that an environment variable in bat is accessed with `%envvar%`. So it should be `type %PT__installdir%\file1.txt`.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
     }
 }
 
+// build a spool dir for the supporting library files requested by a multifile task
+static fs::path createInstallDir(const fs::path& spool_dir, const std::set<std::string>& download_set) {
```
MikaelSmith: I think I'd switch this to `task_cache_dir` to be consistent with what we're actually passing. Also, the comment isn't really accurate. The spool dir is already created.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+      download_set.erase(f_name);
+    }
+
+    return download_set;
+}
+
+// iterate over unique set of filenames to support task
+// copy file from cache to install_dir
+fs::path Task::downloadMultiFile(std::vector<lth_jc::JsonContainer> const& files, std::set<std::string> const& download_set) {
+    auto install_dir =  createInstallDir(task_cache_dir_, download_set);
```
MikaelSmith: This isn't where we talked about putting files. It can work, but we don't have anything making sure they're cleaned up afterwards.

MikaelSmith: Oh, I guess purge below will probably work for it.

lib/src/modules/task.cc
MikaelSmith: When we talk about a spool dir, that's the location results of a run are put. This is something else.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
     }
 }
 
+// build a spool dir for the supporting library files requested by a multifile task
+static fs::path createInstallDir(const fs::path& spool_dir, const std::set<std::string>& download_set) {
+    auto install_dir = spool_dir / fs::unique_path("temp_task_%%%%-%%%%-%%%%-%%%%");
```
MikaelSmith: We should add some debug logging about this to make it easy to find the installdir for a past run (in the logs).

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+    fs::create_directories(install_dir);
+    fs::last_write_time(install_dir, time(nullptr));
```
MikaelSmith: This should be unnecessary because we're always creating a new directory. It's only done elsewhere when we reuse an existing directory.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+        dir_structure.insert(fs::path(f_name).parent_path());
+    }
+
+    // now create the directories to copy the required files into
+    for (auto dir_path : dir_structure) {
+        fs::path tmp = install_dir;
+        for (const fs::path &dir : dir_path) {
+            tmp /= dir;
+            fs::create_directories(tmp);
+            fs::last_write_time(tmp, time(nullptr));
```
MikaelSmith: Same here, shouldn't need to update last_write_time.

lib/src/modules/task.cc
```diff
@@ -426,9 +438,110 @@ static fs::path getCachedTaskFile(const fs::path& task_cache_dir,
...
+            fs::permissions(tmp, NIX_DIR_PERMS);
```
MikaelSmith: This seems concerning. Won't this update permissions starting at the root, like `/`, then `/opt`, etc?

acceptance/tests/tasks/run_multifile.rb
donoghuc: Will do once i get acceptance test working. Looks like it will require addition to `task_helper.rb`.

lib/src/modules/task.cc
donoghuc: Ahh, Yes, I see now (originally got rid of the `erase` calls but neglected to remove the `insert`) I understand that modifying the # of elements DS you are iterating over is not good... Updated.

lib/src/modules/task.cc
donoghuc: Copy paste error. Removed

lib/src/modules/task.cc
donoghuc: Agreed, this is unclear. Updated.

lib/src/modules/task.cc
donoghuc: I was using `t` for debugging by printing name to `std::cout`. Removed. I was not sure if we could depend on the implicit handling of `file_name`. I will experiment with removing it once I get acceptance tests working.

lib/src/modules/task.cc
donoghuc: I am planning on re-working this. I need to understand what information is included in `request.params()` to avoid defaulting to `task_execution_params` when initializing `task_metadata`.

lib/tests/resources/tasks-cache/8fdd86d9be77a27b296d5b6f2d2d984fd7d671bc8bb1ffc86f6c6b0492ec4d01/init.bat
donoghuc: That is one problem. The error message in appveyor shows the expanded paths, it just does not appear to be printing the file contents cleanly. Will keep looking at this, trying `@echo OFF` on this iteration.

lib/src/modules/task.cc
donoghuc: I was using `task_cache_dir` for debugging. Still planning on using spool/results dir.

lib/src/modules/task.cc
donoghuc: `LOG_INFO()`? or `TRACE`? or ?

MikaelSmith: LOG_DEBUG

donoghuc:
jenkins please test this on redhat6-64,windows2012r2-64

acceptance/tests/tasks/run_multifile.rb
donoghuc: Nevermind. No change needed to `task_helper` I can just add it to the metadata as you suggested.

donoghuc:
jenkins please test this on windows2012r2-64

donoghuc:
jenkins please test this on windows2012r2-64

donoghuc:
jenkins please test this on windows2012r2-64

lib/src/modules/task.cc
donoghuc: Settled on using `request.resultDir()` with a fallback to `task_cache_dir_` when not available (Should only happen in tests when `RequestType::Blocking` request is made).

donoghuc:
jenkins please test this on redhat6-64,solaris10-SPARC,solaris11-64,ubuntu1604-64,windows2008r2-64,windows2012r2-64

lib/src/modules/task.cc
MikaelSmith: This doesn't appear to be removed.

lib/src/modules/task.cc
```diff
@@ -592,6 +713,20 @@ ActionResponse Task::callAction(const ActionRequest& request)
 
     task_params.set<std::string>("_task", task_name);
 
+    std::vector<std::string> meta_files;
+    // Only get files array from top level of metatdata instead of puppetserver supplied file data
```
MikaelSmith: This still looks wrong.

lib/src/modules/task.cc
donoghuc: Right, re-added it for debugging and forgot to take it out again (that is a prob with debugging with std::cout). Thanks.

lib/src/modules/task.cc
MikaelSmith: I guess it works.

donoghuc: I experimented a bit with this idea. In most of the unit test examples the files key will not be present. This avoids the leatherman data_type_error exception but leaves an empty object that will have to be guarded against before calling `getMultiFiles` (because it is not of vector<string>). Also I was not finding an easy way of comparing `leatherman::JsonContainer`. How would you compare those? would they end up being the same memory address?

nicklewis:
Notes about install dir purging: If a task is run 100x a day on a node, and has 10 supplementary files (seems fairly high) which are 10k each (definitely large), then during the default 14 day purge time, this would create ~133mb of waste. That doesn't seem too bad, and we can make future improvements to clean it up afterward.

donoghuc:
jenkins please test this on redhat6-64,solaris10-SPARC,solaris11-64,ubuntu1604-64,windows2008r2-64,windows2012r2-64

lib/src/modules/task.cc
MikaelSmith: Ah, don't necessarily have an answer. Let's look at it later.

donoghuc:
jenkins please test this on jenkins please test this on windows2008r2-64a,windows2012r2-64a

donoghuc:
jenkins please test this on windows2008r2-64a,windows2012r2-64a

donoghuc:
@nicklewis did some manual end to end testing based on the passing commit. Multi file task support in PXP-agent seems to be working as expected.

PR #724, opened 2018-09-14
(PCP-856) Update tests to use 'puppetserver ca'

MikaelSmith:
Puppet 6 is removing `puppet cert`. Update tests to use `puppetserver
ca`. For the `invalid_ssl_config` test, this requires a small change in
implementation to generate the alternate CA on the master and copy certs
around, because it's no longer simple to generate it on each agent.


MikaelSmith:
jenkins please test this

PR #725, opened 2018-09-14
(PCP-856) Update invalid_ssl_config to use new ca command

MikaelSmith:
Re-enable the `invalid_ssl_config` test, and update it to use the new
`puppetserver ca` command to generate a CA and certs. Uses only
`puppetserver ca setup` because other commands require a running CA
service.


donoghuc:
jenkins please test this

PR #727, opened 2018-09-26
(PCP-857) Allow lost message to restarted service

MikaelSmith:
Allow that a status request may be undelivered to pxp-agent immediately
after restarting the service. We've seen some occasions where the
service restarts but pcp-broker doesn't receive a disconnect meaning it
always looks like pxp-agent is connected; that can result in losing a
message and waiting 60 seconds before timing out. Add one retry in case
that happened, consistent with how other tools (such as Orchestrator)
implement resilience against lost connections.


MikaelSmith:
jenkins please test this on windows10ent-32a with agent 6.0.x

ekinanp:
jenkins please test this on windows10ent-32a with agent 6.0.x

MikaelSmith:
The first version passed PR tests. The 2nd was purely a change to appveyor.

PR #738, opened 2019-01-15
(PCP-833) Kill all pxp-agent processes during restart

lucywyman:
This modifies the pxp-agent sysv init script to use `killall pxp-agent` in place of `killproc` when stopping the pxp-agent service. `killproc` did not properly remove all pxp-agent processes, but did remove the pxp-agent pid file, which led to dangling processes that never got cleaned up. This PR also removes logic around using `kill` or `killproc` in the stop function, since it's now not necessary.

Note: the `killall` pattern is copied from sshd.

To reproduce the issue on CentOS 6:
1. Install the puppet-agent
1. Run `pgrep -x pxp-agent` to see the current process id
1. Remove the pxp-agent pid file at `/var/run/puppetlabs/pxp-agent.pid`
1. Restart the pxp-agent service: `service pxp-agent restart`
1. Run `pgrep -x pxp-agent` to see that there are now 2 process ids

### DIFF
--- a/acceptance/tests/service_stop_start.rb
+++ b/acceptance/tests/service_stop_start.rb
@@ -49,6 +49,18 @@ def assert_running
   end
 end
 
+def sysv?(platform)
+  if platform.variant =~ /redhat|centos/ && platform.version.to_i < 7
+    return true
+  elsif platform.variant =~ /sles/ && platform.version.to_i < 12
+    return true
+  elsif platform.variant =~ /ubuntu/ && platform.version.to_i < 15
+    return true
+  else
+    return false
+  end
+end
+
 agents.each_with_index do |agent, i|
   @agent = agent
 
@@ -64,6 +76,17 @@ agents.each_with_index do |agent, i|
   step 'C93069 - Service Stop (from running, with configuration)' do
     stop_service
     assert_stopped
+  end
+
+  if sysv?(@agent['platform'])
+    step '`service stop` stops all pxp-agent processes' do
+      start_service
+      assert_running
+      on(@agent, "rm -rf /var/run/puppetlabs/pxp-agent.pid")
+      on(@agent, "service pxp-agent restart")
+      pids = on(@agent, "pgrep -u `whoami` -f /opt/puppetlabs/puppet/bin/pxp-agent").stdout.split("\n")
+      assert_equal(1, pids.length, "Expected 1 pxp-agent process, got #{pids.length}")
+    end
   end
 
   # Solaris service administration will prevent the service from starting
--- a/ext/redhat/pxp-agent.init
+++ b/ext/redhat/pxp-agent.init
@@ -55,6 +55,15 @@ stop() {
             RETVAL=$?
         fi
     fi
+
+    pids=$(pgrep -u `whoami` -x $prog)
+    # Ensure we killed all pxp-agent processes owned by the current user
+    if [ -n "${pids}" ]; then
+        trap '' TERM
+        pkill -TERM  -u `whoami` -f $exec 2>/dev/null
+        trap TERM
+    fi
+
     echo
     [ $RETVAL = 0 ] && rm -f ${lockfile} ${pidfile}
     return $RETVAL
--- a/ext/debian/pxp-agent.init
+++ b/ext/debian/pxp-agent.init
@@ -32,7 +32,15 @@ start_pxp_agent() {
 }
 
 stop_pxp_agent() {
-    start-stop-daemon --stop --retry TERM/10/KILL/5 --quiet --oknodo --pidfile $pidfile --name $prog && rm -f "${pidfile}"
+    start-stop-daemon --stop --retry TERM/10/KILL/5 --quiet --oknodo --pidfile $pidfile --name $prog
+
+    pids=$(pgrep -u `whoami` -x $prog)
+    if [ -n "${pids}" ]; then
+        trap '' TERM
+        pkill -TERM  -u `whoami` -f $exec 2>/dev/null
+        trap TERM
+    fi
+    rm -f "${pidfile}"
 }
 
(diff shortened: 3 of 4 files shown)


ext/redhat/pxp-agent.init
```diff
@@ -46,15 +46,10 @@ start() {
...
-        killproc $pidopts $exec
-        RETVAL=$?
-    else
-        if [ -n "${pid}" ]; then
-            kill -TERM $pid >/dev/null 2>&1
-            RETVAL=$?
-        fi
-    fi
+    trap '' TERM
+    killall $prog 2>/dev/null
```
MikaelSmith: I would lean towards still restricting this to the current user. So
```
killall -u `whoami` $prog 2>/dev/null
```

ext/redhat/pxp-agent.init
```diff
@@ -46,15 +46,10 @@ start() {
 
 stop() {
     echo -n $"Stopping PXP agent: "
-    if [ "$USEINITFUNCTIONS" = "true" ]; then
-        killproc $pidopts $exec
```
MikaelSmith: The sshd implementation also does killproc first. I think we also lose the `[  OK  ]` message from `service` without this.

ext/redhat/pxp-agent.init
```diff
@@ -46,15 +46,10 @@ start() {
...
-    else
-        if [ -n "${pid}" ]; then
-            kill -TERM $pid >/dev/null 2>&1
-            RETVAL=$?
-        fi
-    fi
+    trap '' TERM
+    killall $prog 2>/dev/null
+    trap TERM
+    RETVAL=$?
```
MikaelSmith: I'm not sure this is capturing the right error result. It might be grabbing the result of `trap`.

ext/redhat/pxp-agent.init
lucywyman: Ah sorry - I was using that on the system I was tsting on but didn't seem to include it in the PR! Good catch :)

lucywyman:
jenkins please test this on centos5-64,centos6-64,sles11-64,ubuntu1404-64 with agent 5.5.x

lucywyman:
jenkins please test this on centos5-64a,centos6-64a,sles11-64a,ubuntu1404-64a with agent 5.5.x

lucywyman:
jenkins please test this on centos5-64a,centos6-64a,sles11-64a,ubuntu1404-64a with agent 5.5.x

MikaelSmith:
Test failed on all platforms. It looks like the test is working right. Not sure what's going on from the logs.

ext/redhat/pxp-agent.init
```diff
@@ -55,6 +55,15 @@ stop() {
             RETVAL=$?
         fi
     fi
+
+    pids=$(pgrep -u `whoami` -x $prog)
```
MikaelSmith: When the file is absent, `killproc` might fail and we don't get to this point. Although I'm not sure why it'd do that rather than capturing the status and returning it at line 69.

MikaelSmith: Does this latest version work with manual testing?

lucywyman:
@MikaelSmith  Yes - I've been trying to reproduce the issue manually, so far without success. The failed puppet agent run was most concerning, but I'm hoping it was a transient error....

lucywyman:
M'k - I think it's just not including the changes in this PR in the tests (so it's running the acceptance tests on the old pxp-agent). I made a puppet-agent branch that points to this PR for the pxp-agent component, and rebuilt the jenkins job using that branch. I'm not sure if it'll load the acceptance tests from this PR or if I need to trigger the jenkins run here (with the different vanagon github user and branch) in order for it to include those tests?

lucywyman:
jenkins please test this on centos5-64a,centos6-64a,sles11-64a,ubuntu1404-64a with agent lucywyman/pxp-agent-testing

ext/redhat/pxp-agent.init
```diff
@@ -55,6 +55,15 @@ stop() {
...
+    # Ensure we killed all pxp-agent processes owned by the current user
+    if [ -n "${pids}" ]; then
```
MikaelSmith: SLES (suse) has its own version that needs to be updated as well.

acceptance/tests/service_stop_start.rb
```diff
@@ -66,6 +78,20 @@ def assert_running
...
   end
 
+  if sysv?(@agent['platform'])
+    step '`service stop` stops all pxp-agent processes' do
+      start_service
+      assert_running
+      on(@agent, "rm -rf /var/run/puppetlabs/pxp-agent.pid")
+      # TODO remove after debugging
+      on(@agent, "cat /etc/rc.d/init.d/pxp-agent")
+      stop_service
```
MikaelSmith: Ok, this test is failing because it uses Puppet to stop the service. But when Puppet queries `service pxp-agent status`, it's going to see that it's stopped and not do anything. So `service pxp-agent stop` is never called.

I think we could change this test to restart the service without puppet (since we know it's sysv, we can use `service pxp-agent restart`). The case we really care about is upgrades, and it's not Puppet restarting it then.

lucywyman:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/Adhoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/482/testReport/

MikaelSmith:
I'm ok merging this as soon as you're confident it won't break existing CI.

MikaelSmith:
jenkins please test this on ubuntu1404-64a with agent 5.5.x

lucywyman:
jenkins please test this on ubuntu1404-64a with agent 5.5.x

PR #741, opened 2019-01-18
(PA-669) Disable pxp-agent by default on osx.

tavilarion:


caseywilliams:
I don't disagree with this, but I hope that the version of this PR we eventually merge has a commit message that describes why we're doing this

mihaibuzgau:
integration tests for this: https://github.com/puppetlabs/puppet/pull/7342

MikaelSmith:
Ideally this behavior would map to `enable => false`, and `enable => true` would configure pxp-agent to start after a reboot. There are some scenarios where pxp-agent is the best route to remediating an issue, and if it doesn't start after a reboot then you may no longer have that option. I'm not sure if that's possible, but I'd like someone to at least have explored it.

MikaelSmith:
It makes sense that pxp-agent should not be running and shouldn't start after a reboot when you install it. But after PE configures and starts it (which includes the equivalent of `puppet resource service pxp-agent ensure=running enabled=true`) it *should* start after a reboot. I'd like some confirmation that's the behavior we're getting, or confirmation that it can't be done.

tavilarion:
@MikaelSmith after some thorough research I figured out that setting Disabled to true in the plist will ignore the plist even after reboot until `puppet resource service pxp-agent ensure=running enabled=true` is executed. I believe this is the desired behaviour, please let me know if you have any concerns with this.

PR #742, opened 2019-04-09
(PCP-814) Stop writing task execution params to disk

smcclellan:
Potentially sensitive task execution parameters were being
written to spool-dir. This modifies the way the metadata
is written to enter `"null"` for the "request_params" key.

### DIFF
--- a/lib/src/request_processor.cc
+++ b/lib/src/request_processor.cc
@@ -217,8 +217,6 @@ RequestProcessor::~RequestProcessor()
 void RequestProcessor::processRequest(const RequestType& request_type,
                                       const PCPClient::ParsedChunks& parsed_chunks)
 {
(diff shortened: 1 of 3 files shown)


lib/src/request_processor.cc
```diff
@@ -217,8 +217,7 @@ RequestProcessor::~RequestProcessor()
 void RequestProcessor::processRequest(const RequestType& request_type,
                                       const PCPClient::ParsedChunks& parsed_chunks)
 {
-    LOG_TRACE("About to validate and process PXP request message: {1}",
-              parsed_chunks.toString());
+    LOG_TRACE("About to validate and process PXP request message");
```
MikaelSmith: Not sure I'd be concerned about logging sensitive stuff at trace. I don't think the remaining message is very valuable on its own, so I'd lean towards removing it altogether.

PR #746, opened 2019-05-21
(maint) Add reference for max-message-size

jdub55:
Adding a reference for max-message-size parameter as this was recently added to the pcp-broker but not referenced anywhere in the documentation for pxp-agent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,10 @@ Specifies whether to use PCP version 1 or 2. Only accepts '1' or '2'. Defaults t
 Maximum amount of time that may elapse when trying to establish a connection to
 the broker in seconds. Defaults to 5 seconds.
 
+**max-message-size (optional)**
+
+The maximum message size, specified in bytes. The current default is 64 MB as defined within pcp-broker. This value can be overriden and reduced to something smaller as desired.
+
 **ssl-ca-cert (required to connect)**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -349,6 +349,10 @@ Specifies whether to use PCP version 1 or 2. Only accepts '1' or '2'. Defaults t
 Maximum amount of time that may elapse when trying to establish a connection to
 the broker in seconds. Defaults to 5 seconds.
 
+**max-message-size (optional)**
```
MikaelSmith: This section of documentation is for the pxp-agent config specifically. I don't believe `max-message-size` is configurable there.

If you want to document it in the projects where it's configurable, seems like that should go into https://github.com/puppetlabs/pcp-broker/blob/master/doc/configuration.md, https://github.com/puppetlabs/clj-pcp-client/blob/master/README.md, and Orchestrator (I guess a PCP [ticket](https://tickets.puppet.com)).

Background on this are PCP-491 and PCP-480. The options were added in case we needed to be able to tweak them later, but the goal was to have messages be relatively small and if use-cases came up that violated that we'd try to solve them differently.

jdub55:
Thanks! Closing this and will open it under the pcp-broker project.

PR #745, opened 2019-05-21
(PCP-871) Extract a base module class for bolt support

caseywilliams:
In preparation for supporting run_command, upload_file, and run_script over PCP, these changes extract reusable functionality from the existing Task module into a base bolt module class. Future work will introduce other modules that extend this base class.

PR #748, opened 2019-05-23
(PCP-872) Move download to utility function

McdonaldSeanp:
This commit removes functionality for downloading files from the task module
and adds identical functionality to a base utility. The base utility will be
extendable to other modules for use in downloading files

This should unblock work for download_file and run_script modules

### DIFF
--- a/lib/src/util/bolt_helpers.cc
+++ b/lib/src/util/bolt_helpers.cc
@@ -1,0 +1,185 @@
+#include <pxp-agent/util/bolt_helpers.hpp>
+#include <pxp-agent/configuration.hpp>
+#include <pxp-agent/module.hpp>
+
+#include <boost/algorithm/hex.hpp>
+
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.util.bolt_helpers"
+#include <leatherman/logging/logging.hpp>
+
+#include <openssl/evp.h>
+
+namespace fs       = boost::filesystem;
+namespace alg      = boost::algorithm;
+namespace lth_curl = leatherman::curl;
+namespace lth_jc   = leatherman::json_container;
+namespace lth_loc  = leatherman::locale;
+
+namespace PXPAgent {
+namespace Util {
+
+
+    // NIX_DIR_PERMS is defined in pxp-agent/configuration
+    #define NIX_TASK_FILE_PERMS NIX_DIR_PERMS
+
+    // Downloads a file if it does not already exist on the filesystem. A check is made
+    // on the filesystem to determine if the file at destination already exists and if
+    // it already matches the sha256 provided with the file. If the file already exists
+    // the function immediately returns.
+    //
+    // If the file does not exist attempt to download with leatherman.curl. Once the
+    // download finishes a sha256 check occurs to ensure file contents are correct. Then
+    // the file is moved to destination with boost::filesystem::rename.
+    fs::path downloadFileFromMaster(const std::vector<std::string>& master_uris,
+                                uint32_t connect_timeout,
+                                uint32_t timeout,
+                                lth_curl::client& client,
+                                const fs::path& cache_dir,
+                                const fs::path& destination,
+                                const lth_jc::JsonContainer& file) {
+        auto filename = fs::path(file.get<std::string>("filename")).filename();
+        auto sha256 = file.get<std::string>("sha256");
+
+        if (fs::exists(destination) && sha256 == calculateSha256(destination.string())) {
+            fs::permissions(destination, NIX_TASK_FILE_PERMS);
+            return destination;
+        }
+
+        if (master_uris.empty()) {
+            throw Module::ProcessingError(lth_loc::format("Cannot download task. No master-uris were provided"));
+        }
+
+        auto tempname = cache_dir / fs::unique_path("temp_task_%%%%-%%%%-%%%%-%%%%");
+        // Note that the provided tempname argument is a temporary file, call it "tempA".
+        // Leatherman.curl during the download method will create another temporary file,
+        // call it "tempB", to save the downloaded file's contents in chunks before
+        // renaming it to "tempA." The rationale behind this solution is that:
+        //    (1) After download, we still need to check "tempA" to ensure that its sha matches
+        //    the provided sha. So the downloaded task is not quite a "valid" file after this
+        //    method is called; it's still temporary.
+        //
+        //    (2) It somewhat simplifies error handling if multiple threads try to download
+        //    the same file.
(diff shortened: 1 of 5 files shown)


McdonaldSeanp:
This commit is based on https://github.com/puppetlabs/pxp-agent/pull/745, that will need to be merged first

McdonaldSeanp:
Jenkins please test this on windows2012r2-64a,osx1013-64a,redhat7-64a,ubuntu1804-64a

McdonaldSeanp:
Jenkins please test this on windows2012r2-64a,osx1013-64a,redhat7-64a,ubuntu1804-64a

lib/src/util/bolt_helpers.cc
```diff
@@ -0,0 +1,195 @@
...
+    // the file is moved to destination with boost::filesystem::rename.
+    //
+    // If the destination param is empty: the function assumes the file should
+    // remain in the cache_dir and will be renamed to cache_dir/filename
+    fs::path downloadFileFromMaster(const std::vector<std::string>& master_uris,
+                                uint32_t connect_timeout,
+                                uint32_t timeout,
+                                lth_curl::client& client,
+                                const fs::path& cache_dir,
+                                const fs::path& destination,
```
speedofdark: IS there a reason that this isn't just called `filepath`? Is there an external API that needs this name?

lib/src/util/bolt_helpers.cc
```diff
@@ -0,0 +1,195 @@
...
+            fs::permissions(filepath, NIX_TASK_FILE_PERMS);
+            return filepath;
+        }
+
+        if (master_uris.empty()) {
+            throw Module::ProcessingError(lth_loc::format("Cannot download task. No master-uris were provided"));
+        }
+
+        auto tempname = cache_dir / fs::unique_path("temp_task_%%%%-%%%%-%%%%-%%%%");
+        // Note that the provided "file_path" argument is a temporary file, call it "tempA".
```
speedofdark: `file_path` or `filepath`?

lib/src/util/bolt_helpers.cc
```diff
@@ -0,0 +1,195 @@
...
+        if (!std::get<0>(download_result)) {
+            throw Module::ProcessingError(lth_loc::format(
+                "Downloading the task file {1} failed after trying all the available master-uris. Most recent error message: {2}",
+                file.get<std::string>("filename"),
+                std::get<1>(download_result)));
+        }
+
+        if (sha256 != calculateSha256(tempname.string())) {
+            fs::remove(tempname);
+            throw Module::ProcessingError(lth_loc::format("The downloaded {1}'s sha differs from the provided sha", filename));
```
speedofdark: Does `sha` need to be `SHA` in user visible messaging? Also, try reading this message out loud with an arbitrary filepath in place of the argument and see if the phrasing feels awkward to you. (Yes, I see that this is code that you moved, but anyways...)

lib/src/util/bolt_helpers.cc
```diff
@@ -0,0 +1,195 @@
...
+        std::string url = uri.get<std::string>("path");
+        auto params = uri.getWithDefault<lth_jc::JsonContainer>("params", lth_jc::JsonContainer());
+        if (params.empty()) {
+            return url;
+        }
+        auto curl_handle = lth_curl::curl_handle();
+        url += "?";
+        for (auto& key : params.keys()) {
+            auto escaped_key = std::string(
+            lth_curl::curl_escaped_string(curl_handle, key));
```
speedofdark: Is there a reason these lines are broken at the `(` the way they are? They look short enough without being wrapped.

lib/src/util/bolt_helpers.cc
McdonaldSeanp: destination can be left empty, in which case filepath defaults to cache_dir/filename.

lib/src/util/bolt_helpers.cc
McdonaldSeanp: filepath, typo

lib/src/util/bolt_helpers.cc
McdonaldSeanp: Oh, yeah, it should read:

`The downloaded file {1} has a SHA that does not match the provided SHA`

lib/src/util/bolt_helpers.cc
speedofdark: Understood; the question is why do we pass in one name when all we do is assign it to a different one? Is it because we want to keep it a `const`?

McdonaldSeanp: oh, yeah, now that I think about it you're probably right. There's no need for this to be a const and for the function not to just modify it.

lib/src/util/bolt_helpers.cc
McdonaldSeanp: No, I think you're correct that there's no need for the broken (

lib/src/util/bolt_helpers.cc
McdonaldSeanp: I don't like how I wrote this so I've just moved the destination compilation away from this function. Now destination is always where this function downloads to, and the calling function in task defines destination as cache_dir / filename.

That should be more explicit anyway

McdonaldSeanp:
Jenkins please test this on windows2012r2-64a,osx1013-64a,redhat7-64a,ubuntu1804-64a

McdonaldSeanp:
Jenkins please test this on windows2012r2-64a,osx1013-64a,redhat7-64a,ubuntu1804-64a

lib/src/util/bolt_helpers.cc
```diff
@@ -0,0 +1,188 @@
...
+    // Downloads a file if it does not already exist on the filesystem. A check is made
+    // on the filesystem to determine if the file at destination already exists and if
+    // it already matches the sha256 provided with the file. If the file already exists
+    // the function immediately returns.
+    //
+    // If the file does not exist attempt to download with leatherman.curl. Once the
+    // download finishes a sha256 check occurs to ensure file contents are correct. Then
+    // the file is moved to destination with boost::filesystem::rename.
+    //
+    // If the destination param is empty: the function assumes the file should
```
speedofdark: I don't see logic to support this last comment block. Didn't you move this elsewhere, like into the caller, per the other comments?

McdonaldSeanp: oh, yeah, whoops

McdonaldSeanp:
Passing tests: https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_pr/981/

The only change since then is a comment.

PR #751, opened 2019-05-31
(PE-26615) Fail gracefully when multiple Puppet runs stack up

nicklewis:
Previously, if a Puppet run was blocked by an in-progress run, it would
wait for the lockfile to disappear, attempt to run, and assume that
the run must have now succeeded.

This behavior was incorrect in multiple cases. First, if multiple runs
queue up, they will all wait for the lock and when it becomes free, one
of them will acquire it and the rest will fail. This case was unhandled,
and the result would indicate that no report was written. That is
technically true, but meaningless as the real issue was that the agent
didn't run.

The other case this could happen is if the lock was in place for more
than ten minutes, at which point the module will trigger a second agent
run attempt regardless, which can fail if the lock is still legitimate.

In both of those cases, the error message indicates
"last_run_report.yaml was not written", which is unhelpful at best and
profoundly misleading at worst.

We now explicitly check for the same failure conditions after the second
run as after the first: namely, whether the agent is disabled or locked.
If that is the case, then we report the failure as such rather than
attempting to read a non-existent report.

This also rewords the error message in the case where we _do_ attempt to
read a report that hasn't been updated. It's unclear whether this case
can actually happen under _normal_ operation, but it's a worthwhile
validation step nonetheless. Thus, rather than the purely symptomatic
and "last_run_report.yaml was not written", we now report a more generic
"unexpected failure", which more accurately conveys the reality of the
situation.

### DIFF
--- a/modules/pxp-module-puppet
+++ b/modules/pxp-module-puppet
@@ -13,6 +13,7 @@ module Pxp
       NoLastRunReport = "no_last_run_report"
       InvalidLastRunReport = "invalid_last_run_report"
       Disabled = "agent_disabled"
+      Locked = "agent_locked"
       FailedToStart = "agent_failed_to_start"
       NonZeroExit = "agent_exit_non_zero"
     end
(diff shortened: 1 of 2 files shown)


nicklewis:
I'm not 100% sure this is the correct branch to target the fix, but it seemed appropriate...

McdonaldSeanp:
The fix doesn't seem important enough to warrant support all the way back to 2016.4 agents, so this branch is fine.

MikaelSmith:
jenkins please test this on windows2012r2-64 with agent 5.5.x

PR #754, opened 2019-06-03
(PCP-867) Implement run_command over PCP for bolt support

caseywilliams:
Supports 'command run' for bolt over PCP. See individual commit messages.

### DIFF
--- a/acceptance/lib/pxp-agent/bolt_pxp_module_helper.rb
+++ b/acceptance/lib/pxp-agent/bolt_pxp_module_helper.rb
@@ -31,23 +31,23 @@ end
 # @yield [Array<Hash<String,String>>] Yields an array of response hashes:
 # For a successful response:
 #   ["<agent-pcp-identity-string>" => {
-#       "transaction_id" => "<uuid>"
+#        "transaction_id" => "<uuid>"
 #   }]
 # When there's been an RPC error, the result is different:
-#  ["<agent-pcp-identity-string>" =>  {
-#       "data" => {
-#           "description" => "<error-message>",
-#           "id" => "<uuid>",
-#           "transaction_id" => "<uuid>"
-#        },
-#       "envelope" => {
-#           "expires" => "<date-string>",
-#           "id" => "<uuid>",
-#           "message_type" => "http://puppetlabs.com/<message-type>",
-#           "sender" => "<broker-pcp-identity-string>",
-#           "targets" => ["<agent-pcp-identity-string", ...]
-#        }
-#  }]
+#   ["<agent-pcp-identity-string>" =>  {
+#        "data" => {
+#            "description" => "<error-message>",
+#            "id" => "<uuid>",
+#            "transaction_id" => "<uuid>"
+#         },
+#        "envelope" => {
+#            "expires" => "<date-string>",
+#            "id" => "<uuid>",
+#            "message_type" => "http://puppetlabs.com/<message-type>",
+#            "sender" => "<broker-pcp-identity-string>",
+#            "targets" => ["<agent-pcp-identity-string", ...]
+#         }
+#   }]
 def do_module_action(broker, target_identities, module_name, action_name, params,
                      blocking_request: false,
                      expected_response_type: 'http://puppetlabs.com/rpc_provisional_response')
@@ -69,12 +69,19 @@ def do_module_action(broker, target_identities, module_name, action_name, params
   yield(response_dataset) if block_given?
 end
 
-# Starts a task.
+# Makes a non-blocking request to start a task.
 # Block is executed on the "response_dataset" object from do_module_action.
-def run_task(broker, targets, task, files, input = {}, metadata = nil, **kwargs, &block)
+def run_task(broker, targets, task, files, input: {}, metadata: nil, **kwargs, &block)
   params = { task: task, input: input, files: files }
   params[:metadata] = metadata if metadata
   do_module_action(broker, agent_identities(targets), 'task', 'run', params, **kwargs, &block)
+end
+
+# Makes a non-blocking request to start a command.
+# Block is executed on the "response_dataset" object from do_module_action.
+def run_command(broker, targets, command, **kwargs, &block)
+  params = { command: command }
+  do_module_action(broker, agent_identities(targets), 'command', 'run', params, **kwargs, &block)
 end
 
 # Checks that a non-blocking request finished successfully.
@@ -108,25 +115,51 @@ end
 # Runs a non-blocking task action on targets, and confirms that it was successful.
 # Block is executed on the stdout string.
 def run_successful_task(broker, targets, task, files, input: {}, metadata: nil, **kwargs, &block)
-  run_task(broker, targets, task, files, input, metadata, **kwargs) do |datas|
-    ensure_successful(broker, targets, datas, **kwargs, &block)
+  run_task(broker, targets, task, files, input: input, metadata: metadata) do |response_dataset|
+    ensure_successful(broker, targets, response_dataset, **kwargs, &block)
   end
 end
 
 # Runs a non-blocking task action on targets, and confirms that it failed but did not error.
 # Block is executed on the stderr string.
 def run_failed_task(broker, targets, task, files, input: {}, metadata: nil, **kwargs, &block)
-  run_task(broker, targets, task, files, input, metadata) do |datas|
-    ensure_failed(broker, targets, datas, **kwargs, &block)
+  run_task(broker, targets, task, files, input: input, metadata: metadata) do |response_dataset|
+    ensure_failed(broker, targets, response_dataset, **kwargs, &block)
   end
 end
 
 # Starts a blocking task on targets, and confirms that it produced an RPC error
 # Block is executed on the RPC error message string.
 def run_errored_task(broker, targets, task, files, input: {}, metadata: nil)
-  run_task(broker, targets, task, files, input, metadata,
+  run_task(broker, targets, task, files, input: input, metadata: metadata,
            blocking_request: true,
            expected_response_type: "http://puppetlabs.com/rpc_error_message") do |response_dataset|
+    response_dataset.each { |data| yield data['description'] } if block_given?
+  end
+end
+
+# Runs a non-blocking command action on targets, and confirms that it was successful.
+# Block is executed on the stdout string.
+def run_successful_command(broker, targets, command, **kwargs, &block)
+  run_command(broker, targets, command) do |response_dataset|
+    ensure_successful(broker, targets, response_dataset, **kwargs, &block)
+  end
+end
+
+# Runs a non-blocking command action on targets, and confirms that it unsuccessful, but did not error.
+# Block is executed on the stderr string.
+def run_failed_command(broker, targets, command, **kwargs, &block)
+  run_command(broker, targets, command) do |response_dataset|
+    ensure_failed(broker, targets, response_dataset, **kwargs, &block)
+  end
+end
+
+# Runs a blocking command action on targets, and confirms that it unsuccessful, but did not error.
+# Block is executed on the stderr string.
+def run_errored_command(broker, targets, command)
+  run_command(broker, targets, command,
+              blocking_request: true,
+              expected_response_type: "http://puppetlabs.com/rpc_error_message") do |response_dataset|
     response_dataset.each { |data| yield data['description'] } if block_given?
   end
 end
--- a/acceptance/tests/pxp_module_bolt_command/run_command.rb
+++ b/acceptance/tests/pxp_module_bolt_command/run_command.rb
@@ -1,0 +1,23 @@
+require 'pxp-agent/bolt_pxp_module_helper.rb'
+
+test_name 'run_command task' do
+  step 'Ensure each agent host has pxp-agent running and associated' do
+    agents.each do |agent|
+      on agent, puppet('resource service pxp-agent ensure=stopped')
+      create_remote_file(agent, pxp_agent_config_file(agent), pxp_config_hocon_using_puppet_certs(master, agent))
+      on agent, puppet('resource service pxp-agent ensure=running')
+
+      assert(is_associated?(master, "pcp://#{agent}/agent"),
+             "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
+    end
+  end
+
+  step 'Run a command (`echo`) on agent hosts' do
+    agents.each do |agent|
+      cmd = agent.platform =~ /windows/ ? 'cmd /c echo hello' : 'echo hello'
+      run_successful_command(master, [agent], cmd) do |stdout|
+        assert_equal('hello', stdout.chomp)
+      end
+    end
+  end # test step
+end # test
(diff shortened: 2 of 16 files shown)


caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
This is ready for review -- this PR has gotten kind of large, so it may help to review the commits one by one.

I'm not sure how to fix the appveyor failures -- the failing unit test (which just calls echo.exe) passes for me on vmpooler machine with cygwin, where I just built this same revision -- if anyone has ideas, I'd very much appreciate it.

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
Looks like something is up with the package signing process in Jenkins for windows. I'll rekick once it's resolved.

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

lib/src/modules/command.cc
```diff
@@ -0,0 +1,69 @@
...
+           params.type("command") == lth_jc::DataType::String);
+
+    const fs::path& results_dir = request.resultsDir();
+
+    // Try to split the raw command text into executable and arguments, as
+    // leatherman::execution expects:
+    std::vector<std::string> command_parts;
+    auto raw_command = params.get<std::string>("command");
+
+    boost::split(command_parts,
```
McdonaldSeanp: Will this function be affected by windows executables with spaces in their path? i.e. something like 
`C:\Program Files\OpenSSL\bin\openssl.exe`

McdonaldSeanp: Additionally, do we want to fail early here if the executable does not include an extension on windows?

On windows the file extension is required to call any executable, will leatherman execute look for an extension? or just fail? I would assume we would just fail in that scenario since guessing file extensions seems fishy at best.

caseywilliams: It definitely won't handle that correctly - I'm not sure what a good heuristic is to split the user input up into executable and arguments in this case - would it be reasonable to expect the user to quote pathnames that have spaces in them within the command parameter, like `{"command": "\"C:\\Program Files\\OpenSSL\\bin\\openssl.exe\" ..."}`?

McdonaldSeanp: Yes, if the path has spaces I believe it's reasonable to expect quotes.

caseywilliams: And yeah, it looks like leatherman execution is checking for file extensions [here](https://github.com/puppetlabs/leatherman/blob/master/execution/src/windows/execution.cc#L57)

exe/CMakeLists.txt
```diff
@@ -9,19 +9,19 @@ add_executable(pxp-agent main.cc)
...
-    list(APPEND TASK_WRAPPER_LIBS ${CMAKE_THREAD_LIBS_INIT})
+    list(APPEND EXECUTION_WRAPPER_LIBS ${CMAKE_THREAD_LIBS_INIT})
 endif()
 
-add_executable(task_wrapper task_wrapper.cc)
-target_link_libraries(task_wrapper ${TASK_WRAPPER_LIBS})
-install(TARGETS task_wrapper DESTINATION bin)
+add_executable(execution_wrapper execution_wrapper.cc)
+target_link_libraries(execution_wrapper ${EXECUTION_WRAPPER_LIBS})
+install(TARGETS execution_wrapper DESTINATION bin)
```
MikaelSmith: There's a small chance someone has directly used the task_wrapper executable - I remember some conversations related to using it for some invocation logic, maybe with Choria. We should call out the rename in release notes when we introduce this work.

lib/src/util/bolt_module.cc
```diff
@@ -112,22 +121,48 @@ void BoltModule::callNonBlockingAction(
         const Util::CommandObject &command,
         ActionResponse &response
 ) {
-    // TODO: throw if storage_ isn't available
-    auto exec = run(command);
-    // Stdout / stderr output should be on file; read it
+    // Guaranteed by Configuration
```
MikaelSmith: So much cleaner, 👍

acceptance/lib/pxp-agent/bolt_pxp_module_helper.rb
```diff
@@ -50,12 +50,32 @@ def do_module_action(broker, target_identities, module_name, action_name, params
   yield(datas) if block_given?
 end
 
+def run_command(broker, targets, command, metadata, **kwargs, &block)
```
MikaelSmith: As we add more of these helpers, it might be useful to separate out the `run_successful` part as something that invokes a function passed to it and confirms it succeeded, then build `run_successful_*` from it and the associated `run_*` function.

acceptance/tests/pxp_module_bolt_command/run_command.rb
```diff
@@ -0,0 +1,23 @@
...
+      on agent, puppet('resource service pxp-agent ensure=running')
+
+      assert(is_associated?(master, "pcp://#{agent}/agent"),
+             "Agent #{agent} with PCP identity pcp://#{agent}/agent should be associated with pcp-broker")
+    end
+  end
+
+  step 'Run a command (`echo`) on agent hosts' do
+    agents.each do |agent|
+      cmd = agent.platform =~ /windows/ ? 'cmd /c echo hello' : 'echo hello'
```
MikaelSmith: My recollection is that pxp-agent has always invoked the executable directly rather than running in the context of a shell. Bolt over other transports has different semantics. I wouldn't change it in this PR, but do we want to rethink how `run_command` works with pxp-agent to more closely mirror Bolt with other transports?

lib/src/modules/command.cc
MikaelSmith: Agreed that quoting paths or arguments with spaces should be expected. https://stackoverflow.com/questions/21335041/how-to-split-a-string-using-shell-like-rules-in-c has some suggestions how, looks like https://www.boost.org/doc/libs/1_65_0/doc/html/boost/program_options/split_unix.html might cover it. Windows makes that trickier, but it might work well enough.

Would anything be easier if we made Bolt split the string?

MikaelSmith: It turns out run_command is kind of weird, we're accepting something shell-specific but not running it in a shell. So we need to define what the characteristics are. Looks like bolt_shim went with whatever the semantics of https://www.rubydoc.info/stdlib/open3/Open3.capture3 are.

acceptance/tests/pxp_module_bolt_command/run_command.rb
MikaelSmith: Based on investigation around shell splitting, I'm no longer certain that's the case.

caseywilliams: I don't have enough experience within this project to have much of an opinion, but @branan ^ this seems worth discussing sooner rather than later

McdonaldSeanp: @MikaelSmith the use of cmd.exe in this acceptance test isn't because we are running in the shell for run_command. There just isn't an echo executable on windows, so there was no simple way to create simple stdout without executing in cmd.exe first.

The actual invocation of the command to be used should be using leatherman::execute

McdonaldSeanp: Oh, wait, I guess maybe what your suggesting is we _should_ be executing in the shell?

MikaelSmith: The question was just inspired by the example. I think it's confusing to folks when the answer to that question changes depending on what transport they're using, but I might be confusing this with the Docker transport (where it can't assume a shell because there might not be a shell executable).

acceptance/lib/pxp-agent/bolt_pxp_module_helper.rb
speedofdark: This seems like a good path forward for using common methods. 👍

lib/tests/unit/modules/command_test.cc
```diff
@@ -0,0 +1,99 @@
...
+
+    ActionRequest request { RequestType::NonBlocking, ECHO_COMMAND_CONTENT };
+    fs::path spool_path { SPOOL_DIR };
+    auto results_dir = (spool_path / request.transactionId()).string();
+    fs::create_directories(results_dir);
+    request.setResultsDir(results_dir);
+
+    auto response = mod.executeAction(request);
+
+    SECTION("executes a command") {
```
speedofdark: This is a successful execution; would it make sense to have ones that failed in various ways to sure the command validation mechanisms all worked?

acceptance/tests/pxp_module_bolt_command/run_command.rb
speedofdark: If we add some unsuccessful command invokes at the unit level, we can probably get away with doing happy path testing at the acceptance level. Otherwise, we should consider some unhappy path tests here (failed and pxp_errored)? As we're not currently mucking with the protocol, we don't need to explore that space.

caseywilliams:
Thanks for the feedback, all -- I've added many more unit test cases and handling for whitespace in executable and argument names, but I still need to rework the acceptance tests.

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

caseywilliams:
jenkins please test this on osx1012-64a,windows2012r2-64a

lib/src/modules/command.cc
```diff
@@ -0,0 +1,71 @@
...
+    const auto params = request.params();
+
+    assert(params.includes("command") &&
+           params.type("command") == lth_jc::DataType::String);
+
+    const fs::path& results_dir = request.resultsDir();
+
+    // Try to split the raw command text into executable and arguments, as leatherman::execution
+    // expects, taking care to retain spaces within quoted executables or argument names:
+    auto raw_command = params.get<std::string>("command");
```
MikaelSmith: Looks like this was pulled from program_options::split_unix. Might be worth calling out the origin.

lib/tests/unit/modules/command_test.cc
```diff
@@ -0,0 +1,216 @@
...
+    Modules::Command mod { PXP_AGENT_BIN_PATH, STORAGE };
+
+#ifdef _WIN32
+    static const std::string echo_params { "{ \"command\": \"cmd.exe /c echo hello\" }" };
+#else
+    static const std::string echo_params { "{ \"command\": \"echo hello\" }" };
+#endif
+
+    auto request = command_request(echo_params);
+    auto response = mod.executeAction(request);
```
speedofdark: It appears that `executeAction` has sufficient unit tests via `task_test.cc` (and few scattered across `module_test.cc` and `ping_test.cc`); we also cover external modules via `external_module_test.cc`.

caseywilliams: are you suggesting some of these test cases don't need to be here? I'm not sure how else to test this code outside of calling executeAction to do so. I will try to improve the test descriptions in general.

speedofdark:
No, I'm saying that it appears we do not need to add tests for
executeAction because its already quite well covered.

lib/tests/unit/modules/command_test.cc
speedofdark: @caseywilliams looking for feedback: is there a better way for me to have phrased this such that it would have been less confusing?Thank you.

caseywilliams:
I've split the acceptance refactor into its own PR at #761, so that other tickets are unblocked. I will need to update this PR's acceptance tests accordingly.

lib/tests/unit/modules/command_test.cc
caseywilliams: Sure, I did not consider validating `executeAction`'s exiting behavior to be part of this ticket, since that method's been around for a while

caseywilliams:
jenkins please test this on windows2012r2-64a,osx1012-64a

exe/execution_wrapper.cc
```diff
@@ -52,10 +61,11 @@ int main(int argc, char *argv[])
...
 #endif
-        ofs << lth_loc::format("Task '{1}' failed to run: {2}", task_executable, e.what());
+        ofs << lth_loc::format("Executable '{1}' failed to run: {2}", executable, e.what());
         exitcode = 127;
     }
 
+    // Write the exit code; at this point, stdout and stderr are already written
 #ifdef _WIN32
     lth_file::atomic_write_to_file(std::to_string(exitcode), params.get<std::string>("exitcode"));
 #else
```
caseywilliams: I haven't changed this behavior in this PR, but I'm wondering why there is a write to the exitcode file here in the task wrapper, but also a PID callback that writes to the same file on the task module's call to `leatherman::execution::execute`. But I notice that if I remove the callback from the task module, some unit tests do fail. I haven't included a similar PID callback for the command module here, so I'm just relying on the task wrapper to write the exit code. I'm curious if there's something I'm missing here.

MikaelSmith: Why do you say they write to the same file? I see the task module writing out a `pid` file (which would be the process ID of the process it just started). The task wrapper writes out a `exitcode` file when it's done (which is the exit code of that same process).

caseywilliams: Sorry, you're right, I was totally confused when I commented - I'll work on resolving the rest of the failures here today

caseywilliams:
jenkins please test this on windows2012r2-64a,osx1012-64a

caseywilliams:
jenkins please test this on windows2012r2-64a,osx1012-64a

caseywilliams:
I've pushed another update to add two more unit test cases, but acceptance tests passed [here](https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_pr/1013/). I think this is finally ready for review and/or merging again.

acceptance/lib/pxp-agent/bolt_pxp_module_helper.rb
```diff
@@ -108,29 +115,55 @@ def ensure_failed(broker, targets, response_dataset, max_retries: 30, query_inte
...
+
+# Runs a non-blocking command action on targets, and confirms that it unsuccessful, but did not error.
+# Block is executed on the stderr string.
+def run_failed_command(broker, targets, command, **kwargs, &block)
+  run_command(broker, targets, command) do |response_dataset|
+    ensure_failed(broker, targets, response_dataset, **kwargs, &block)
+  end
+end
+
+# Runs a non-blocking command action on targets, and confirms that it unsuccessful, but did not error.
```
speedofdark: the comment says non-blocking, but the param on [L161](https://github.com/puppetlabs/pxp-agent/pull/754/commits/85058699d4a6346660e57bc9e40d755cf3262d98#diff-b29b1cee2f7efe4d0cf5fb1ab71bc054R161) says `blocking_request: true` ?

caseywilliams: Ah right, I fixed this in https://github.com/puppetlabs/pxp-agent/pull/761, but after I'd moved those changes from here to there. I've pushed a fix.

PR #756, opened 2019-06-07
(PA-2267) Update pxp tests to check /proc/1/exe for systemd (1.9.x)

GabrielNagy:
Previous to this commit, two tests: run_puppet_umask and
multiple_pxp_agent_daemon were attempting to use the 'service' command to check
for systemd. This will not work for newer fedora versions since sysvinit was
removed and the 'service' command went with it.

Thus we check if the file that /proc/1/exe points to is systemd

Merging this to 1.9.x also, as we need it for puppet-agent 5.5.x.

### DIFF
--- a/acceptance/tests/multiple_pxp_agent_daemon.rb
+++ b/acceptance/tests/multiple_pxp_agent_daemon.rb
@@ -8,7 +8,7 @@ test_name 'Start pxp-agent daemon with pidfile present' do
   end
 
   applicable_agents = applicable_agents.reject do |agent|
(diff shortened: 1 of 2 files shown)


GabrielNagy:
Passing run with this commit on Fedora 30:

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/Adhoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/609/RMM_COMPONENT_TO_TEST_NAME=pxp_agent,SLAVE_LABEL=beaker,TEST_TARGET=fedora30-64a/

PR #766, opened 2019-07-22
(PCP-874) Add functionality for recursive downloads in download_file

McdonaldSeanp:
This work should expand the download_file module in pxp-agent to implement taking an array of files/directories/symlinks and creating them all on the system.

The schema for download_file already takes an array (previously it just operated on the first element), so the only update required for the schema is a field to differentiate between files/dirs/symlinks.

### DIFF
--- a/acceptance/tests/pxp_module_bolt_download_file/download_file.rb
+++ b/acceptance/tests/pxp_module_bolt_download_file/download_file.rb
@@ -22,6 +22,10 @@ def test_file_exists(agent, file)
   assert_match(/ensure\s*=>\s*'file',/, on(agent, puppet("resource file #{file}")).stdout)
 end
 
+def test_dir_exists(agent, dir)
+  assert_match(/ensure\s*=>\s*'directory',/, on(agent, puppet("resource file #{dir}")).stdout)
+end
+
 def test_file_does_not_exist(agent, file)
   assert_match(/ensure => 'absent'/, on(agent, puppet("resource file #{file}")).stdout)
 end
@@ -63,8 +67,9 @@ test_name 'download file tests' do
     on master, puppet('resource service puppetserver ensure=running')
   end
 
-  step 'execute successful download_file' do
+  step 'execute successful download_file with files and directories' do
     suts.each do |agent|
+      test_dir = windows?(agent) ? "C:/Windows/Temp/test_dir#{rand(10000000)}" : "/opt/test_dir#{rand(10000000)}"
       test_files = [
         test_file_destination(agent),
         test_file_destination(agent),
@@ -74,14 +79,32 @@ test_name 'download file tests' do
       test_files.each do |file|
         request << download_file_entry(@sha256, "/download-test-files/#{@filename}", file, 'file')
       end
+      request << download_file_entry('', '', '', test_dir, 'directory')
       run_successful_download(master,
                               agent,
                               request)
       test_files.each do |file|
         test_file_exists(agent, file)
       end
+      test_dir_exists(agent, test_dir)
       teardown {
         clean_files(agent, test_files)
+      }
+    end
+  end
+
+  step 'execute download_file for a file with a destination directory that doesnt exist yet' do
+    suts.each do |agent|
+      test_dir = windows?(agent) ? "C:/Windows/Temp/test_dir#{rand(10000000)}" : "/opt/test_dir#{rand(10000000)}"
+      test_file = test_dir + "/testing_file#{rand(10000000)}.txt"
+      run_successful_download(master,
+                              agent,
+                              [download_file_entry(@filename, @sha256, "/download-test-files/#{@filename}", test_file, 'file')])
+      test_file_exists(agent, test_file)
+      teardown {
+        clean_files(agent, [test_file])
+        # Remove the test directory too.
+        assert_match(/ensure => 'absent'/, on(agent, puppet("resource file #{test_dir} ensure=absent force=true")).stdout)
       }
     end
   end
--- a/acceptance/tests/pxp_module_bolt_download_file/download_file.rb
+++ b/acceptance/tests/pxp_module_bolt_download_file/download_file.rb
@@ -68,7 +68,7 @@ test_name 'download file tests' do
     on master, puppet('resource service puppetserver ensure=running')
   end
 
-  step 'execute successful download_file with files and directories' do
+  step 'execute successful download_file with files,symlinks,directories' do
     suts.each do |agent|
       test_dir = test_dir_destination(agent)
       test_symlink = test_file_destination(agent)
@@ -93,6 +93,8 @@ test_name 'download file tests' do
       test_file_resource_exists(agent, test_symlink, 'link')
       teardown {
         clean_files(agent, test_files)
+        assert_match(/ensure => 'absent'/, on(agent, puppet("resource file #{test_dir} ensure=absent force=true")).stdout)
+        assert_match(/ensure => 'absent'/, on(agent, puppet("resource file #{test_symlink} ensure=absent force=true")).stdout)
       }
     end
   end
--- a/acceptance/tests/pxp_module_bolt_download_file/download_file.rb
+++ b/acceptance/tests/pxp_module_bolt_download_file/download_file.rb
@@ -5,25 +5,25 @@ require 'puppet/acceptance/environment_utils.rb'
 PUPPETSERVER_CONFIG_FILE = '/etc/puppetlabs/puppetserver/conf.d/webserver.conf'
 
 def test_file_destination(agent)
-  windows?(agent) ? 'C:/testing_file.txt' : '/opt/testing_file.txt'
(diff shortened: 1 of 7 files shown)


McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

lib/tests/unit/modules/download_file_test.cc
```diff
@@ -191,6 +222,16 @@ TEST_CASE("Modules::DownloadFile::callAction", "[modules]") {
         REQUIRE(fs::is_regular_file(fs::path(TEST_FILE_DIR + "/file.txt")));
     }
 
+    SECTION("Correctly returns exitcode 1 if the new symlink requested is already a file") {
+        // FAILURE_DIR_EXISTS_AS_FILE_CONTENT will fail because it will make a request to make a
```
speedofdark: comment doesn't match code at line 228.

McdonaldSeanp:

lib/src/modules/download_file.cc
```diff
@@ -113,31 +119,71 @@ namespace Modules {
...
-    } catch (Module::ProcessingError& e) {
-      LOG_ERROR("Failed to download {1}; {2}", destination, e.what());
-      response.output = write_download_results(results_dir, EXIT_FAILURE, "", lth_loc::format("Failed to download {1}; {2}", destination, e.what()));
+    for (auto this_file : files) {
+      auto destination = fs::path(this_file.get<std::string>("destination"));
+      auto file_type = this_file.get<std::string>("file_type");
+      if (file_type == "file") {
+        try {
+          if (!fs::exists(destination.parent_path())) {
+            Util::createDir(destination.parent_path());
```
branan: Why is this necessary?

lib/src/modules/download_file.cc
```diff
@@ -113,31 +119,71 @@ namespace Modules {
...
+            return failure_response(request, results_dir, lth_loc::format("Destination {1} already exists and is not a directory!", destination));
+          }
+        } else {
+          try {
+            Util::createDir(destination);
+          } catch (fs::filesystem_error& e) {
+            return failure_response(request, results_dir, lth_loc::format("Failed to create directory {1}; {2}", destination, e.what()));
+          }
+        }
+      } else if (file_type == "symlink") {
```
branan: Have we confirmed how Bolt and Puppet behave for symlinks (on POSIX and Windows) so we can be consistent?

lib/src/util/bolt_helpers.cc
```diff
@@ -188,6 +188,11 @@ namespace Util {
     fs::permissions(dir, NIX_DIR_PERMS);
   }
 
+  void createSymLink(const fs::path& destination, const fs::path& source) {
+    fs::create_symlink(destination, source);
+    fs::permissions(destination, NIX_DIR_PERMS);
```
branan: This is a race. In theory someone could access the symlink with default permissions in between creation and the permissions fix. We should be able to specify the permissions at creation time (on POSIX with UMASK, on Windows just as part of the CreateFile IIRC). Dunno how much of a refactor there is to achieve that...

lib/src/modules/download_file.cc
McdonaldSeanp: There's an acceptance test that uses puppet to check that puppet is reporting the symlink as:

`ensure => link`

So I _think_ they behave the same.

lib/src/modules/download_file.cc
McdonaldSeanp: Now that I think about it I think this is in the wrong place. downloadFileFromMaster doesn't create the directory itself, it downloads the file to a cachedir and renames it (using fs::rename()) if the SHA matches. So if the underlying destination dir for the file doesn't exist yet it will break.

Though after looking at it I think downloadFileFromMaster _should_ be creating the underlying destination dir if it doesn't exist, so that it can do so only after the matching SHA is determined.

lib/src/util/bolt_helpers.cc
McdonaldSeanp: So after some digging on this:

It looks like there are ways for POSIX nodes to specify umask of the pxp-agent service, which means any file/symlink/dir created with boost should match the umask of the service process. That means between the file being created and the call to fs::permissions it looks like the user can still control the security of the files/symlinks/dirs.

For windows: all permissions are inherited by the containing directory, meaning there is no scenario where files/symlinks will have perms more open than the containing directories.

Since download_file isn't supporting specialized permissions from the user I believe the combination of those things means we don't need to worry about the race.

McdonaldSeanp:

acceptance/tests/pxp_module_bolt_download_file/download_file.rb
```diff
@@ -63,33 +68,74 @@ def test_file_does_not_exist(agent)
     on master, puppet('resource service puppetserver ensure=running')
   end
 
-  step 'execute successful download_file' do
-    clean_files(suts)
+  step 'execute successful download_file with files and directories' do
```
speedofdark: files, links, directories? Or is that too pedantic?

acceptance/tests/pxp_module_bolt_download_file/download_file.rb
```diff
@@ -63,33 +68,74 @@ def test_file_does_not_exist(agent)
...
+      request << download_file_entry('', '', test_files[0], test_symlink, 'symlink')
+      run_successful_download(master,
+                              agent,
+                              request)
+      test_files.each do |file|
+        test_file_resource_exists(agent, file, 'file')
+      end
+      test_file_resource_exists(agent, test_dir, 'directory')
+      test_file_resource_exists(agent, test_symlink, 'link')
+      teardown {
```
speedofdark: does something else take care of cleaning non-file resources?

lib/tests/unit/modules/download_file_test.cc
```diff
@@ -58,17 +60,68 @@ static const std::string CRT { "mock_crt" };
...
+                                                                    "},"
+                                                                    "\"sha256\":\"\","
+                                                                    "\"destination\":\"%1%/file.txt\","
+                                                                    "\"link_source\":\"\","
+                                                                    "\"file_type\":\"directory\""
+                                                                "}"
+                                                            "]"
+                                                       "}") % TEST_FILE_DIR;
+
+// Create a failure scenario by using a request with a "directory" file type where
```
speedofdark: directory file type?

McdonaldSeanp: Yeah, the field in the schema I used to identify if it was a file/dir/link is called "file_type". I can change that.

acceptance/tests/pxp_module_bolt_download_file/download_file.rb
McdonaldSeanp: Oh, whoops, no.

lib/src/modules/download_file.cc
```diff
@@ -113,31 +119,68 @@ namespace Modules {
...
-                    destination,
-                    files[0]);
-      response.output = write_download_results(results_dir, EXIT_SUCCESS, "downloaded", "");
-    } catch (Module::ProcessingError& e) {
-      LOG_ERROR("Failed to download {1}; {2}", destination, e.what());
-      response.output = write_download_results(results_dir, EXIT_FAILURE, "", lth_loc::format("Failed to download {1}; {2}", destination, e.what()));
+    for (auto this_file : files) {
+      auto destination = fs::path(this_file.get<std::string>("destination"));
+      auto content_type = this_file.get<std::string>("content_type");
+      if (content_type == "file") {
```
caseywilliams: `content_type` seems a little strange as the name of the parameter, because it makes me think about the actual contents of the file/symlink/directory rather than the type of thing it is - I think `file_type` or `kind` or something like that might work.

But do we have to add a type param at all? I'm wondering if we could use a symlink if the `link_source` param is set, and otherwise use the same strategy as before?

(If this is here and called content_type to match some other code somewhere else, ignore me)

caseywilliams: Oh, nevermind about the second part, I was looking at the changeset as a whole, I didn't realize we kind of need the new parameter to deal with directories

lib/src/util/bolt_helpers.cc
caseywilliams: Can you leave a comment near this code with a summary of that info for later?

lib/src/modules/download_file.cc
McdonaldSeanp: I had this var named "file_type" before, but Jonathan pointed out that it made things confusing to talk about (i.e. "file_type":"file" is awkward language to use).

I'll change it to `kind` though. I think that's probably a good idea.

PR #767, opened 2019-07-29
(maint) refactor download_file schema to remove filename

McdonaldSeanp:
It turns out the "filename" part of the schema for the download_file module in
pxp-agent is unnecessary. Since the filename can be fetched from the
destination field (since it contains the filename), the downloadFileFromMaster
utility can just fetch the filename from destination using boost::filesystem's
.filename().

This commit does two things:

1. Updates the downloadFileFromMaster utility to fetch the filename from the
destination parameter, rather than grabbing it from the file parameter.

2. Removes the filename part of the download_file schema.

Task file downloads are unaffected: since the task download creates a
destination using the cache_dir + the filename, downloadFileFromMaster will
still fetch the correct filename.


McdonaldSeanp:

PR #772, opened 2019-09-10
(PCP-883) Allow powershell task parameters with `type` in the name

donoghuc:
Previously, when parsing xml in the `Get-ContentAsJson` implementation a regex used to filter out xml elements with whos `MemberType` `Properties` had a name that was preceeded with `__` or contained the string `type`. Filtering on the string `type` caused task parameters that contain that string to be ignored, preventing users from being able to effectively pass powershell task paramters with the `type` in the name. Unfortunately, with the current implementation powershell parameters with the name `type` will be clobbered by a `Property` called type. However this commit updates the regex to *only* filter out the property that exactly matches `type`. Note that this work allows powershell task parameters with `type` in the name, but not the name `type` and only affects powershell <= 2.

### DIFF
--- a/acceptance/files/complex-task.ps1
+++ b/acceptance/files/complex-task.ps1
@@ -46,7 +46,12 @@ param(
 
   [Parameter(Mandatory = $False)]
   [Switch]
-  $ArgSwitch
+  $ArgSwitch,
+
+  # Parameters may have the string `type` in them, but not have the name `type`
+  [Parameter(Mandatory = $False)]
+  [String]
+  $types
 )
 
 function ConvertTo-String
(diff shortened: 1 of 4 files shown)


branan:
[](https://xkcd.com/1171)

acceptance/files/complex-task.ps1
```diff
@@ -47,6 +47,11 @@ param(
   [Parameter(Mandatory = $False)]
   [Switch]
   $ArgSwitch
```
branan: I think you need a comma after this

donoghuc: yeah, oops. copy paste problems.

PR #773, opened 2019-09-13
(PCP-844) Update the command module to use a shell

McdonaldSeanp:
This commit updates the "run" action in the command pxp-agent module to wrap
all commands to run in a shell. This feature mistakenly assumed that bolt
commands ran with no shell, when that's not the case, thus we need to change
the implementation to wrap everything in a shell.

This commit wraps everything in a shell by creating the command structure with
the shell as the command to execute in leatherman. Then the arguments will
include the actual command passed from the PCP message.

For windows:
Bolt uses WinRM, which defaults to powershell as the shell (and that isn't
believed to be configurable) so we default to powershell.exe

For Posix systems:
We use the posix defined getpwuid (https://linux.die.net/man/3/getpwuid) to
find the default shell for the calling pxp-agent processes user.

### DIFF
--- a/lib/tests/unit/modules/command_test.cc
+++ b/lib/tests/unit/modules/command_test.cc
@@ -87,7 +87,7 @@ TEST_CASE("Modules::Command::callAction successfully", "[modules]") {
     Modules::Command mod { PXP_AGENT_BIN_PATH, STORAGE };
 
 #ifdef _WIN32
-    static const std::string echo_params { "{ \"command\": \"cmd.exe /c echo hello\" }" };
+    static const std::string echo_params { "{ \"command\": \"write-host hello\" }" };
 #else
     static const std::string echo_params { "{ \"command\": \"echo hello\" }" };
 #endif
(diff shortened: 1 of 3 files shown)


McdonaldSeanp:

McdonaldSeanp:

McdonaldSeanp:

lib/tests/unit/modules/command_test.cc
```diff
@@ -209,45 +218,3 @@ TEST_CASE("Modules::Command::callAction where the command fails", "[modules]") {
         }
     }
 }
-
-TEST_CASE("Modules::Command::buildCommandObject where the executable contains spaces", "[modules]") {
```
caseywilliams: Is it not worth testing commands with spaces in them anymore?

McdonaldSeanp: Not anymore no. The command no longer needs to be split based on which string should go in the "Executable" part since the actual executable is the shell interpreter, and everything else including the command to run is an argument.

PR #778, opened 2019-10-09
(PCP-886) Catch EXDEV error on file rename, retry with copy

McdonaldSeanp:
This commit updates the download helpers to catch the EXDEV error code
during boost::filesystem::rename. This error occurs when rename is
called with a source and destination on different filesystems. In this
case, we can retry the file move by using copy and then remove rather
than rename.


McdonaldSeanp:

PR #784, opened 2020-02-14
(PCP-875)(PCP-891) Add apply module to pxp-agent

donoghuc:
This commit adds a new internal apply module. The apply module has a single `apply` action which will apply a catalog on an agent. The result of an apply is a JSON report of the puppet run. Currenly the ruby shim is managed as a HEREDOC in the apply.cc source. This should be managed as a shipped file in pxp-agent packaging. We will do that work in subsequent tickets.

This commit updates the ruby shim and the apply module C++ code to enable downloading plugins in to the temporary environment the catalog is applied in. Currently, the C++ code is responsible for ensuring a cache directory is available for the shim to download plugins in to. The shim is responsible for ensuring the plugins are up to date and copying them in to the expected location in the temporary environment.

We pass through the connection information from the pxp-agent through to the shim. We use this for pluginsync and then clear all setting before applying the compiled catalog to ensure isolation. Currently PXP does not honor CRL. We need to enable that in pxp-agent and then we can pass that through to the shim as well.

In order to download plugins from puppetserver with puppet in the apply shim we need CRL. This commit makes CRL a required parameter for pxp-agent for use with the apply module. Plumbing in the agent to honor the CRL in the other built in modules as well as connecting to the PCP broker will be subsequently implemented. It is tracked with https://tickets.puppetlabs.com/browse/PCP-893


lib/src/modules/apply.cc
```diff
@@ -0,0 +1,234 @@
...
+  #   Puppet::ModuleTool::Tar.instance.unpack(plugins, moduledir, user)
+  # end
+
+  env = Puppet.lookup(:environments).get('production')
+  # Needed to ensure features are loaded
+  env.each_plugin_directory do |dir|
+    $LOAD_PATH << dir unless $LOAD_PATH.include?(dir)
+  end
+
+  if (conn_info = args['_target'])
```
donoghuc: I should comment this out and add a TODO. We do not yet pass along _target

lib/tests/unit/modules/apply_test.cc
```diff
@@ -0,0 +1,108 @@
...
+TEST_CASE("Modules::Apply::hasAction", "[modules]") {
+    Modules::Apply mod { PXP_AGENT_BIN_PATH, MASTER_URIS, CA, CRT, KEY, "", 10, 20, MODULE_CACHE_DIR, STORAGE };
+    SECTION("correctly reports false") {
+        REQUIRE(!mod.hasAction("foo"));
+    }
+
+    SECTION("correctly reports true") {
+        REQUIRE(mod.hasAction("apply"));
+    }
+}
```
donoghuc: Given we do not have an agent (and ruby interpreter) for acceptance tests we cant really do much more than this super minimal set of unit tests. Unless anybody has a better idea? I thought about putting some bash in a $tasks-cache/apply_ruby_shim/apply_ruby_shim.rb file with a bash shebang that echod back the input, BUT that would never work on windows (and is not a meaningful test).

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,234 @@
...
+  Puppet.settings.use(:main)
+
+  ## TODO: ENABLE PLUGINS
+  # Tempfile.open('plugins.tar.gz') do |plugins|
+  #   File.binwrite(plugins, Base64.decode64(args['plugins']))
+  #   user = Etc.getpwuid.nil? ? Etc.getlogin : Etc.getpwuid.name
+  #   Puppet::ModuleTool::Tar.instance.unpack(plugins, moduledir, user)
+  # end
+
+  env = Puppet.lookup(:environments).get('production')
```
donoghuc: TODO: use env from shim arguments.

donoghuc: I actually dont think this does anything. We set up an isolated environment for this execution but puppet needs a "real" environment, so we use `production`. I asked the bolt team yesterday in slack but it did not get any feedback.

lib/src/modules/apply.cc
donoghuc: ah, right. I forgot about that. WIll remove.

lib/src/modules/apply.cc
```diff
@@ -80,43 +90,51 @@ begin
...
+  #  ensure
+  #    pool.close
+  #  end
+  #end
+
+  # Queston about caching... It appears that the file resource type wont need to fully re-download every file in the destdirs. So all the CPP code needs to do 
+  # is to make sure a cache dir named after an environment is available and puppet will do the rest? Does this approach make sense? IE cache by env with a TTL.
+  # HACK: todo figure out a better way, specifically:
+  # 1. Is shelling out (spawning a new proccess) under a leatherman scripting container OK and will it in any way contaminate state in the rest of the shim script
+  # 2. The bolt plugin method packages up the plugins a bit differently than plugin_sync (they are effectively extracted from modules), is copying them
```
donoghuc: The cache should be the concatenation of `environment`-`version`.

lib/src/modules/apply.cc
```diff
@@ -80,43 +90,51 @@ begin
...
+  # from cache to /var/lib OK?
+  plugin_cache = args['plugin_cache']
+  remote_env = args['environment']
+  begin
+    # shell out and let puppet do this operation as if it is in a "real" environment, this should be safe bc we are controlling the output dirs
+    stdout_str, stderr_str, status = Open3.capture3("/opt/puppetlabs/bin/puppet plugin download --plugindest #{plugin_cache} --pluginfactdest #{plugin_cache} --environment #{remote_env}")
```
donoghuc: Potentially guard against command injection.

donoghuc: We should not give up on doing this in process. If we continue to use sub-process is to *really* make sure that we are isolated.

lib/src/modules/apply.cc
```diff
@@ -80,43 +90,51 @@ begin
...
+    # Copy libs into our temp environment (puppet_root/var/lib will already be on the load path, ok to move there?)
+    FileUtils.copy_entry(plugin_cache, "#{File.join(puppet_root, 'var', 'lib')}")
```
donoghuc: The other option is to add the libs in the cachedir to the $LOAD_PATH. This would allow us to *not* copy over the entire plugin set each time.

donoghuc: Things to consider: Locking the cache dir so we dont delete it out from under an apply. AND/OR perhaps touching file that are used so mtime is updated and ttls are not reached bc puppet is smart enough to not re-download files.

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,251 @@
...
+  #    pool.close
+  #  end
+  #end
+
+  plugin_cache = args['plugin_cache']
+  remote_env = args['environment']
+  begin
+    # shell out and let puppet do this operation as if it is in a "real" environment, this should be safe bc we are controlling the output dirs
+    # TODO: make invocation cross platform https://github.com/puppetlabs/pxp-agent/blob/579b883a5545f9a0895571d5f8cde329be5981ab/modules/pxp-module-puppet#L377-L384
+    stdout_str, stderr_str, status = Open3.capture3("/opt/puppetlabs/bin/puppet plugin download --plugindest #{plugin_cache} --pluginfactdest #{plugin_cache} --environment #{remote_env}")
```
donoghuc: Setting `--plugindir` and `--pluginfactdir` is a bad bug. They need to be separate so the file resource does not conflict causing exactly *not* the caching behavior we want.

donoghuc:
I still need to add an acceptance test that applies a catalog that *requires* plugins and pluginfacts. I am currently focusing on taking over work on https://github.com/puppetlabs/orchestrator/pull/1289 so that I can get an end-to-end environment to develop a good module to test pluginsync.

donoghuc:
Added acceptance tests.

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,252 @@
...
+        "properties": {
+            "environment": {
+                "type": "string"
+            },
+            "version": {
+                "type": "string"
+            },
+            "catalog": {
+                "type": "object"
+            },
```
donoghuc: Would like to talk about this schema. See https://github.com/puppetlabs/orchestrator/pull/1298/

lib/inc/pxp-agent/modules/apply.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <leatherman/locale/locale.hpp>
+#include <leatherman/curl/client.hpp>
+
+namespace PXPAgent {
+namespace Modules {
+
+class Apply : public PXPAgent::Util::BoltModule, public PXPAgent::Util::Purgeable {
+    public:
+        Apply(const boost::filesystem::path& exec_prefix,
+              const std::vector<std::string>& master_uris,
```
branan: This (and the settings below) are what we need to have for the various SSL settings in the ruby-side code. Maybe the best fix is to plumb these through and use them to set Puppet's settings, rather than fighting with loading settings multiple times?

donoghuc: Damn thats a really good idea. Let me see if I can do that!

lib/src/modules/apply.cc
branan: This name might end up in the report? We may have to make it not-production. Unfortunately Puppet has some weirdness about creating environments not named production out of thin air...

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,252 @@
...
+  Puppet.settings.use(:main)
+
+  # Given we set up an empty env, it will be 'production' in this case
+  # Note we *do not* want to set this to the environment we are applying the catalog for
+  env = Puppet.lookup(:environments).get('production')
+
+  # Ensure custom facts are available for provider suitability tests
+  facts = Puppet::Node::Facts.indirection.find(SecureRandom.uuid, environment: env)
+
+  report = if Puppet::Util::Package.versioncmp(Puppet.version, '5.0.0') > 0
```
branan: We know we're gonna have puppet > 6 in our environment

donoghuc: lol yeah that was copy paste ;)

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,252 @@
...
+           end
+
+  overrides = { current_environment: env,
+                loaders: Puppet::Pops::Loaders.new(env) }
+
+  Puppet.override(overrides) do
+    catalog = Puppet::Resource::Catalog.from_data_hash(args['catalog'])
+    catalog.environment = env.name.to_s
+    catalog.environment_instance = env
+    if defined?(Puppet::Pops::Evaluator::DeferredResolver)
```
branan: Again, we know we'll have Puppet 6

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,252 @@
...
+  # Puppet::Util::Log.newdestination(:console)
+  # Puppet[:log_level] = :info
+
+  # Sync plugins to cache. Note that we are explicitly *not* using an isolated confdir
+  # because we need to be able to communicate with puppetserver 
+  plugin_cache = args['plugin_cache']
+  plugin_dest = File.join(plugin_cache, 'plugins')
+  pluginfact_dest = File.join(plugin_cache, 'pluginfacts')
+  Puppet[:plugindest] = plugin_dest
+  Puppet[:pluginfactdest] = pluginfact_dest
```
branan: Whenever possible, we should be adding settings to the cli - that way any derived settings or implicit early init does the right thing

lib/src/modules/apply.cc
donoghuc: I seem to have been able to get  `Puppet::Node::Environment.remote(args['environment'])` to work for everything. I will investigate that further.

lib/inc/pxp-agent/modules/apply.hpp
```diff
@@ -40,6 +40,10 @@ class Apply : public PXPAgent::Util::BoltModule, public PXPAgent::Util::Purgeabl
       boost::filesystem::path exec_prefix_;
 
       std::vector<std::string> master_uris_;
+      std::string ca_;
+      std::string crt_;
+      std::string key_;
+      std::string proxy_;
```
donoghuc: TODO: pipe through proxy.

lib/src/modules/apply.cc
```diff
@@ -0,0 +1,271 @@
...
+# Create temporary directories for all core Puppet settings so we don't clobber
+# existing state or read from puppet.conf. Also create a temporary modulepath.
+# Additionally include rundir, which gets its own initialization.
+puppet_root = Dir.mktmpdir
+moduledir = File.join(puppet_root, 'modules')
+Dir.mkdir(moduledir)
+plugin_cache = args['plugin_cache']
+plugin_dest = File.join(plugin_cache, 'plugins')
+pluginfact_dest = File.join(plugin_cache, 'pluginfacts')
+# Use isolated directories in puppet_root *except* confdir so we can connect to puppetserver
```
donoghuc: TODO: delete this comment. It is no longer accurate.

lib/inc/pxp-agent/modules/apply.hpp
donoghuc: We needed to add in CRL to make this work. We have decided this is the way to go and will move forward by first requiring CRL for apply module and follow up in https://tickets.puppetlabs.com/browse/PCP-893 with honoring CRL in other modules and communications with PCP broker.

lib/src/modules/apply.cc
donoghuc: done

lib/src/modules/apply.cc
donoghuc: I dont see it in the report. I am really tempted to use the "battle tested" (at least in bolt) method of using `'production'` here.

lib/src/modules/apply.cc
donoghuc: Updated every setting to prefer CLI over settings where possible.

lib/src/modules/apply.cc
donoghuc: yeep. done

lib/src/modules/apply.cc
McdonaldSeanp: Yeah, I think it's fine to leave this with a comment about the discussion we've had here. Just something to describe why this says production and that we don't see anything in the report to make us think it needs to be different.

donoghuc:
jenkins please test this on redhat6-64,solaris10-SPARC,solaris11-64,ubuntu1604-64,windows2008r2-64,windows2012r2-64

lib/src/modules/apply.cc
donoghuc: Given we do not have `version` plumbed through puppetserver yet we need to just cache on enviroment for now.

lib/src/modules/apply.cc
donoghuc: We are able to do this in process. This is no longer a problem.

donoghuc:
jenkins please test this on redhat6-64,solaris10-SPARC,solaris11-64,ubuntu1604-64,windows2008r2-64,windows2012r2-64

PR #787, opened 2020-03-16
(PCP-894) Add `prep` action to `apply` module

donoghuc:
WIP- this is the first cpp module with multiple actions. This first commit is just feeling out how to build up schema validation for that.

### DIFF
--- a/lib/src/modules/apply.cc
+++ b/lib/src/modules/apply.cc
@@ -32,9 +32,11 @@ namespace Modules {
     }
     )" };
 
+    static const std::string prep_ACTION { "prep" };
+
     // TODO: Actually manage this file in packaging (or perhaps fetch it from puppetserver).
     // For now just copy it to the cache dir each time unless there is already a copy
-    static const std::string RUBY_SHIM {
+    static const std::string APPLY_RUBY_SHIM {
 R"(#! /opt/puppetlabs/puppet/bin/ruby
 # frozen_string_literal: true
 
@@ -62,18 +64,19 @@ setting_keys = Puppet::Settings::REQUIRED_APP_SETTINGS.append(:rundir)
 cli_base = (setting_keys).flat_map do |setting|
   ["--#{setting}", File.join(puppet_root, setting.to_s.chomp('dir'))]
 end
-
-# There will always be at least a single master URI
-# TODO: make sure comma separated is what --server_list expects
-server_list = args['master_uris'].map { |uri| URI.parse(uri).host }.join(',')
-
-cli_settings_pluginsync = [
+cli_base.concat([
   '--modulepath',
   moduledir,
   '--plugindest',
   plugin_dest,
   '--pluginfactdest',
-  pluginfact_dest,
+  pluginfact_dest
+])
+# There will always be at least a single master URI
+# TODO: make sure comma separated is what --server_list expects
+server_list = args['master_uris'].map { |uri| URI.parse(uri).host }.join(',')
+
+cli_settings_pluginsync = [
   '--localcacert',
   args['ca'],
   '--hostcert',
@@ -105,7 +108,7 @@ begin
 
   Puppet.initialize_settings(cli_base + cli_settings_pluginsync)
 
-  remote_env_for_plugins = Puppet::Node::Environment.remote(args['catalog']['environment'])
+  remote_env_for_plugins = Puppet::Node::Environment.remote(args['environment'])
   downloader = Puppet::Configurer::Downloader.new(
     "plugin",
     Puppet[:plugindest],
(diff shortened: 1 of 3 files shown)


lib/src/modules/apply.cc
```diff
@@ -129,7 +132,7 @@ begin
 
   # Now fully isolate again
   Puppet.settings.send(:clear_everything_for_tests)
-  Puppet.initialize_settings(cli_base + cli_settings_pluginsync)
```
donoghuc: That was kind of a bad mistake in PCP-891 first iteration :grimacing:

lib/src/modules/apply.cc
```diff
@@ -206,17 +336,20 @@ exit exit_code
     {
         module_name = "apply";
         actions.push_back(apply_ACTION);
+        actions.push_back(prep_ACTION);
+        PCPClient::Schema apply_input_schema { apply_ACTION, lth_jc::JsonContainer { apply_ACTION_INPUT_SCHEMA } };
+        input_validator_.registerSchema(apply_input_schema);
 
-        PCPClient::Schema input_schema { apply_ACTION, lth_jc::JsonContainer { apply_ACTION_INPUT_SCHEMA } };
-        input_validator_.registerSchema(input_schema);
+        PCPClient::Schema prep_input_schema { prep_ACTION };
```
donoghuc: Is this how `PCPClient::Schema` is intended to work? This is the first built-in cpp module with multiple module actions?

lib/src/modules/apply.cc
```diff
@@ -225,43 +358,55 @@ exit exit_code
...
-        // TODO: remove this when we figure out how to manage ruby shim
-        const std::string ruby_shim_cache_dir = "apply_ruby_shim";
-        const std::string ruby_shim_script_name = "apply_ruby_shim.rb";
+
+        const std::string ruby_shim_cache_dir = "apply_ruby_shims";
         const fs::path& cache_dir = module_cache_dir_->createCacheDir(ruby_shim_cache_dir);
-        auto ruby_shim_path = cache_dir / ruby_shim_script_name;
-        if (!fs::exists(ruby_shim_path)) {
-          lth_file::atomic_write_to_file(RUBY_SHIM, ruby_shim_path.string(), NIX_DOWNLOADED_FILE_PERMS, std::ios::binary);
+
```
donoghuc: Does this fit with general module design? There are no other built-in cpp modules that have multiple actions implemented for a single module. Do we just handle the logic in buildCommandObject? Or are we supposed to have separate implementations of buildCommandObject for each action?

lib/src/modules/apply.cc
```diff
@@ -179,6 +182,133 @@ ensure
...
+# Proxy will always be a string (even if its empty)
+parsed_proxy = URI.parse(args['proxy'])
+proxy_flags.each do |uri_method, setting_flag|
+  val = parsed_proxy.send(uri_method)
+  cli_settings_pluginsync << setting_flag << val unless val.nil?
+end
+
+exit_code = 0
+begin
+
```
donoghuc: Once we pull these out into shipped files we can look in to sharing some code between `prep` and `apply`.

lib/src/modules/apply.cc
McdonaldSeanp: This all looks correct to me

lib/src/modules/apply.cc
```diff
@@ -225,43 +358,55 @@ exit exit_code
...
         const fs::path& cache_dir = module_cache_dir_->createCacheDir(ruby_shim_cache_dir);
-        auto ruby_shim_path = cache_dir / ruby_shim_script_name;
-        if (!fs::exists(ruby_shim_path)) {
-          lth_file::atomic_write_to_file(RUBY_SHIM, ruby_shim_path.string(), NIX_DOWNLOADED_FILE_PERMS, std::ios::binary);
+
+        std::string ruby_shim_script_name;
+        fs::path apply_ruby_shim_path;
+        std::string plugin_cache_name;
+
+        if (params.includes("catalog")) {
```
McdonaldSeanp: We should see if there's a more robust way to decide between the two actions.

lib/src/modules/apply.cc
McdonaldSeanp: It might make more sense to have one shim that can take the action to perform as a param.

lib/src/modules/apply.cc
McdonaldSeanp: I think this is the first module with multiple actions, so we're in new territory here.

Personally I think we should stick with one buildCommandObject until someone sees a good reason to split it

lib/src/modules/apply.cc
donoghuc: https://github.com/puppetlabs/pxp-agent/blob/master/lib/inc/pxp-agent/action_request.hpp#L35 yeah i meant to comment there must be a better way. I think ^^ should do it.

lib/src/modules/apply.cc
donoghuc: For sure. Updated to a single shim. I started the work thinking they would be pretty different, turns out they are not.

lib/src/modules/apply.cc
donoghuc: Updated to use that.

PR #789, opened 2020-03-23
(maint) make ssl-crl optional

robdaemon:
For backwards compatibility, the ssl-crl parameter has to be optional.

Reintroduces the add apply module to pxp-agent work


lib/src/modules/apply.cc
```diff
@@ -78,12 +78,14 @@ cli_settings_pluginsync = [
...
   '--hostprivkey',
   args['key'],
-  '--hostcrl',
-  args['crl'],
   '--server_list',
   server_list
 ]
 
+if args['crl']
+  cli_settings_pluginsync << '--hostcrl' << args['crl']
```
donoghuc: This will always be required. We should error out in the CPP module code if apply action is requested.

donoghuc: I think for this PR we just need to make sure there is no error when we load config, but it *does* error when we try to use apply module action.

McdonaldSeanp: Yep, that sounds right to me

PR #791, opened 2020-03-26
(PCP-894) move the shim to a external file

robdaemon:
Creates a libexec folder to house the apply Ruby shim and other scripts


robdaemon:
this isn't ready yet, it doesn't take Windows into account but I'm getting there

robdaemon:
There isn't a good portable way that I know of. You need `argv[0]`, and that's only available in `main`.

My first pass was making a function in `configuration.cc` and declaring it `extern` in `apply.cc` but we have unused functions set as errors.

donoghuc:
Deprioritizing this work for now.

PR #793, opened 2020-04-07
(maint) Use remote env everywhere

donoghuc:
Previously we looked up a default env for application. This caused report environemnt to not provide a value for environment name in apply report. By using remote environemnt everywhere we do not rely on default env.

### DIFF
--- a/lib/src/modules/apply.cc
+++ b/lib/src/modules/apply.cc
@@ -50,6 +50,7 @@ require 'uri'
 
 ## TODO: Option to read from a file is for debugging. Only read from stdin in production.
 args = JSON.parse(ARGV[0] ? File.read(ARGV[0]) : STDIN.read)
+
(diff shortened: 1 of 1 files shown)


lib/src/modules/apply.cc
```diff
@@ -153,19 +153,19 @@ begin
   env = Puppet.lookup(:environments).get('production')
```
donoghuc, line 153: This is what is being replaced. If we move forward w/this we need to delete this line.

lib/src/modules/apply.cc
```diff
@@ -177,8 +177,7 @@ begin
     puts JSON.pretty_generate(report.to_data_hash)
     exit_code = report.exit_status != 1
   else
-    facts.name = facts.values['clientcert']
```
donoghuc, line 180: this is wrong. Revert this part.

PR #786, opened 2020-03-09
(PCP-893) Honor CRL when communicating with broker/puppetserver

donoghuc:
This commit updates pxp-agent to use a new required `ssl-crl` configuration option. This option represents the CRL to validate when communicating with PCP brokers or puppet master/server.

PR #800, opened 2020-04-22
(PCP-896) Acceptance puppetserver setup to fixture

johnduarte:
This PR adds a fixture directory for a 'bolt' environment.  This
directory contains the bolt tasks, scripts, and download files used for
various bolt tests.

An additional setup step is added to rsync this environment to the
puppetserver.  This will provide the pxp-agent tests the necessary
puppetserver configuration without having to setup and restart the
puppetserver in an ad-hoc basic per test.

The server setup step is also added to the 'aio' options file to ensure
that it is executed as part of the setup suite for an 'aio'
installation.

All applicable tests that configured and restarted puppetserver have
been updated to use the common environment that is created during
the beaker setup stage.


johnduarte:

acceptance/setup/common/070_Setup_Server_fixtures.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  step 'stop puppetserver' do
+    on master, puppet('resource service puppetserver ensure=stopped')
+  end
+
+  PUPPETSERVER_CONFIG_FILE = '/etc/puppetlabs/puppetserver/conf.d/webserver.conf'
+
+  fixtures = File.absolute_path('files')
+  test_env = 'bolt'
+  env_path = ''
```
donoghuc, line 13: This is more for my own beaker edification. I have been thinking of a pattern for dealing with variables we need in a "global" (ie shared between steps, teardown etc) context. The way I was thinking about it is that for variables we only ever need to read (in this case PUPPETSERVER_CONFIG_FILE, fixtures and test_env) we would use just "global" references to a variable. In the case where steps want to mutate the value should we use instance variables? Even though it appears to be safe enough to mutate the variables defined outside the step scope it seems like maybe it would be more clear to the test author which vars are read only and which ones may change? 

This is not something i've really seen a concrete pattern for, its just something i've been thinking about. I don't really understand what happens in the scope of a `test_name` and I think that just setting random instance variables could actually be dangerous.

johnduarte: @donoghuc I am happy to iterate on this to get a better pattern out of it.  The lowest common denominator for these types of variables is that they are used within a 'test_name' block and are _usually_ scoped within that block.

My bias is to scope these variables as narrowly as required.  Here are my thoughts on the available variable options given my understanding of their scoping within beaker.

* Globals ("$foo = 'bar'"): Beaker uses a lot of modules that extend beaker and tend to be very loose about variable scoping.  Within test cases using global variables could cause unintended side effects by leaking them into the broader beaker stack.  I would advise strongly against using them under any circumstances.

* Constants ("FOO = 'bar'):  I haven't seen this used much within beaker tests.  I would expect this to apply to the scope of the `TestCase` object (defined at the file level).  Since these are immutable, this would be a nice pattern to use for immutable values that apply to the entire test file.  They should probably be declared outside the scope of the 'test_name' block to make that clear.

* Instance variables ("@foo = 'bar'):  In the context of a beaker test, instance variables are associated with the `TestCase` object.  This scope is broader than the 'test_name' block in a test case file.  If multiple "tests" defined by 'test_name' blocks are included in a file, these instances would bleed across them.  This is almost always undesirable.

* Local variables ("foo = bar"):  These are _usually_ declared in the scope of a test 'step' block, but there are often cases when a variable needs to be declared outside of a 'step' block to make it available to multiple 'step' blocks within a 'test_name' block.  I realize that it is not as easy to read and pick them out when they scope at the 'test_name' block, but they are the most restrictive of the available options.  The mutability point you bring up is a valid concern though.  It would be reasonable to `freeze` local variables used across 'step' blocks as soon as the desired value is assigned to protect against this.

donoghuc: Well this explanation is beyond and away helpful! Moving forward local variables seem for sure to be the way to go. Maybe constants outside of the test_name block. Really appreciate the depth.

donoghuc:
This got merged https://github.com/puppetlabs/pxp-agent/pull/786

Maybe we should target 1.9.x (I think that's the branch that feeds 5.x puppet agent) and then mergeup.

johnduarte:
> Maybe we should target 1.9.x (I think that's the branch that feeds 5.x puppet agent) and then mergeup.

The only test from these changes that is part of 1.9.x is the proxy test.  If we re-target, we will still need to open another PR for the other test updates after the merge up takes place.  I am fine with that strategy if that works best.

The other option is to leave this PR targeted at master and cherry pick https://github.com/puppetlabs/pxp-agent/pull/800/commits/26b00c73a7ca4e4d5ebab1670b03fe7ad69c6d87 and https://github.com/puppetlabs/pxp-agent/pull/800/commits/d3c3c086ea824d183e061eefbf6b9f95e2a8406c .

I have no particular preference.  Let me know which way you would prefer to proceed @donoghuc .  Thanks.

donoghuc:
I dont have a preference either. I think the task/download test https://github.com/puppetlabs/pxp-agent/blob/1.9.x/acceptance/tests/tasks/task_purged_from_cache.rb would use it too on the 1.9.x branch. I was more pointing out that we merged in more code that follows the "old" pattern https://github.com/puppetlabs/pxp-agent/pull/786/files#diff-c714da86f83b9fbf56eac941fb01367cR79-R104 . Whatever you find easier works for me :smiley:

johnduarte:
:man_facepalming: oh, I see that the task_download test has been moved between 1.9.x and master.

I will close this PR and target the changes at 1.9.x.

PR #806, opened 2020-05-08
(maint) Only run crl_revoke test on external agent

johnduarte:
This commit removes the master from the agent hosts used in the
`crl_revocation` test.  If the master is included in the test the
master's ssl cert will be revoked centrally causing SSL authentication
to fail.


johnduarte:

PR #808, opened 2020-05-15
(maint) Remove unused Apply constructor params

donoghuc:
The `Apply` module does not need to use the curl client to download anything from master/puppetserver. this commit removes the client and it's configuration params.


donoghuc:

PR #810, opened 2020-05-29
(PCP-897) allow setting of http proxy when running as a service

ciprianbadescu:
Before this commit, when running pxp-agent as a service
 it was not possible to specify a http proxy

 With this change, `http_proxy` and `https_proxy` variables are passed from
 /etc/sysconfig/pxp-agent(RedHat) or /etc/default/pxp-agent(Debian) to `pxp-agent`
 process environment while running as a service

### DIFF
--- a/ext/debian/pxp-agent.default
+++ b/ext/debian/pxp-agent.default
@@ -2,3 +2,5 @@
 
 # Startup options
 #PXP_AGENT_OPTIONS=--loglevel=debug
+#http_proxy="http://10.0.0.1:3128"
+#https_proxy="http://10.0.0.1:3128"
--- a/ext/debian/pxp-agent.init
+++ b/ext/debian/pxp-agent.init
@@ -28,6 +28,8 @@ reload_pxp_agent() {
 
 start_pxp_agent() {
     mkdir -p $piddir $logdir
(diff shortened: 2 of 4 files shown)


ciprianbadescu:
setting `http_proxy` and `https_proxy` environment variables controls proxy values used by underlying http(s) implementation of libcurl (https://curl.haxx.se/libcurl/c/libcurl-env.html).

this PR/jira ticket is about allowing you to pass those environment variables from /etc/sysconfig/pxp-agent

donoghuc:
Is there another use case for libcurl than downloading task/script content from master/puppetserver? https://github.com/puppetlabs/pxp-agent/blob/master/lib/src/modules/task.cc#L171 for example?

donoghuc:
I commented on the Jira ticket. I feel like if there is a bug with https://tickets.puppetlabs.com/browse/PCP-848 we should fix it. This seems like an incomplete solution.

donoghuc:
Closing this due to conversation in https://tickets.puppetlabs.com/browse/PCP-897 . It looks like the wrong settings were being used. Specifically setting proxy in puppet.conf does not cascade to proxy settings in pxp-agent.conf.

PR #811, opened 2020-06-05
(maint) Allow plan for dev build to not use bundler

McdonaldSeanp:
This commit updates the dev build plan to include an optional no_bundler
parameter that will ignore use of bundler when attempting the agent build


steveax:
I understand the change, but not the why. Can you expand on that a bit?

McdonaldSeanp:
@steveax: @donoghuc had mentioned that he primarily runs bolt using bundler, so that he can run on the edge bolt builds. Running bolt in that way and having the tasks call bundler caused problems, so we just added an option to avoid the whole bundler situation in the case you are doing something similar.

PR #812, opened 2020-06-08
(PE-29137) Make use of Curl threadsafe

donoghuc:
The use of the Leatherman Curl client is not thread safe (https://github.com/puppetlabs/leatherman/blame/master/curl/inc/leatherman/curl/client.hpp#L271) when module actions that use the client to dowload content from puppetserver were run concurrently a curl error was raised and the action would fail. This comit synchronizes the use of the curl client accross and within modules. Specifically the helper methods that implement curl were moved to the ModuleCacheDir class where a mutex is maintained to synchonize the use of the curl client. The ModuleCacheDir class was chosen because it is instantiated in the RequestProcesser and a pointer to the object is already passed to modules that need to download content from the master. This ensures that the mutex synchonizes the use of the client both between and within modules.


donoghuc:
jenkins please test this on centos7-64a,ubuntu1804-64a

lib/src/modules/task.cc
```diff
@@ -150,7 +150,7 @@ Task::Task(const fs::path& exec_prefix,
     task_download_connect_timeout_ { task_download_connect_timeout },
     task_download_timeout_ { task_download_timeout },
 #ifdef _WIN32
-    features_ { "puppet-agent", "powershell" }
+    features_ { "puppet-agent", "powershell" },
```
McdonaldSeanp, line 153: This change broke the appveyor build:

the #ifdef clause here is switching out how features_ is set for windows vs. Unix. In the case of Windows builds this will be the last line rather than part of a list.

lib/inc/pxp-agent/modules/task.hpp
```diff
@@ -8,6 +8,7 @@
 #include <pxp-agent/util/purgeable.hpp>
 #include <pxp-agent/util/bolt_module.hpp>
 
+#include <cpp-pcp-client/util/thread.hpp>
```
McdonaldSeanp, line 11: I don't _think_ this line should be required, since all the thread work is done within the modulecachedir class which already includes this header.

lib/src/util/bolt_helpers.cc
```diff
@@ -14,8 +14,8 @@
 
 namespace fs       = boost::filesystem;
 namespace alg      = boost::algorithm;
-namespace lth_curl = leatherman::curl;
-namespace lth_jc   = leatherman::json_container;
+// namespace lth_curl = leatherman::curl;
```
McdonaldSeanp, line 17: I'm assuming you were testing that we could remove these lines. Looks like the jenkins build worked so we should be 👍 to remove them now.

lib/src/modules/task.cc
```diff
@@ -376,14 +375,12 @@ Util::CommandObject Task::buildCommandObject(const ActionRequest& request)
...
     auto file = selectTaskFile(files, implementation);
-
-    auto task_file = Util::getCachedFile(master_uris_,
-                                       task_download_connect_timeout_,
-                                       task_download_timeout_,
-                                       client_,
-                                       module_cache_dir_->createCacheDir(file.get<std::string>("sha256")),
-                                       file);
-
+    auto task_file = module_cache_dir_->getCachedFile(master_uris_,
```
McdonaldSeanp, line 378: Nice :) looks pretty clean to me.

lib/src/modules/task.cc
donoghuc: Thanks, yeah that was an artifact left over from testing what happens if the mutex was managed as an instance variable on individual module objects. Will clean up.

McdonaldSeanp:
Oh, heh, whoops.

Looks like there was a WIP label on this. Hopefully the review is still a little helpful :)

donoghuc:
Orch testing is ready once other threading bugs have been solved. https://github.com/puppetlabs/orchestrator/pull/1381

MikaelSmith:
I think I just ran into this as `easy handle already used in multi handle`, so thanks!

donoghuc:
Yeah, that is the error I was reliably seeing resulting from this bug. https://gist.github.com/donoghuc/4b5242201d28eb2525c0e0ce7fd3da12#easy-handle-already-used-in-multi-handle has a bit more detail.

PR #817, opened 2020-08-14
(PE-28827) Handle max-message-size agent side

donoghuc:
DRAFT for discussion


lib/src/request_processor.cc
```diff
@@ -149,9 +150,20 @@ void nonBlockingActionTask(std::shared_ptr<Module> module_ptr,
...
+    LOG_WARNING("action_metadata {1}", action_metadata_string);
+    LOG_WARNING("action_metadata size {1}", action_metadata_string.size());
+    LOG_WARNING("max_message_size {1}", max_message_size);
+    LOG_WARNING("message_to_large {1}", message_to_large);
     if (response.action_metadata.get<bool>("notify_outcome")) {
-        if (response.action_metadata.get<bool>("results_are_valid")) {
+        if (response.action_metadata.get<bool>("results_are_valid") && !message_to_large) {
             connector_ptr->sendNonBlockingResponse(response);
+        } else if (message_to_large){
+            err_msg = lth_loc::format("Message size: {1} exceeded max-message-size {2}", action_metadata_string.size(), max_message_size);
```
MikaelSmith, line 164: Would an admin be able to see the message that it failed to send somewhere?

Ideally we'd add multi-part messages to the protocol, but this seems like it would at least make the failure clearer.

Does pcp-broker send a response when it rejects a message?

McdonaldSeanp: > Would an admin be able to see the message that it failed to send somewhere?
If you're referring to the PE admin, then yes, the goal here is to have the failure propagate to orchestrator and fail in the resulting status of the action.

> Ideally we'd add multi-part messages to the protocol, but this seems like it would at least make the failure clearer.
We've talked a bunch about that, I'm not convinced that our services are built in a way that can handle this much data loaded in to memory.

Updating the protocol to stream the response in parts is also a large enough chunk of work that we would need to legitimately plan and prioritize it. At this point we just need better error handling so that it fails in a consistent way and folks can come to us with their use-case for large result processing. 

> Does pcp-broker send a response when it rejects a message?
No, nothing happens in pcp-broker when this occurs save one logged error. It silently continues.

donoghuc: I think the first question was more related to "can an admin see the message that was too large to send". We figured instead of logging the big message, we would just assume the admin could look in the spool dir and look at the data there.

donoghuc:
jenkins please test this

donoghuc:
jenkins please test this

donoghuc:
jenkins please test this on redhat7-64a,windows2016-64a

donoghuc:
jenkins please test this on redhat7-64a,windows2016-64a

McdonaldSeanp:
jenkins please test this on redhat7-64a,windows2016-64a

donoghuc:
jenkins please test this on redhat7-64a,windows2016-64a

McdonaldSeanp:
jenkins please test this on redhat7-64a,windows2016-64a

PR #819, opened 2020-09-14
(maint) Update max-message-size calculation

barriserloth:
This commit updates the max-message-size to be 64*1024*1024 rather than
64*1012*1012 in order to more accurately calculate the number of bytes
in 64 megabytes.


McdonaldSeanp (approved):

McdonaldSeanp:
jenkins please test this

donoghuc (approved):
wow, i'm bad a dimensional analysis

PR #823, opened 2020-11-06
(maint) Use cp -rL when backing up ca directory

joshcooper:
The `cp -r` command does not follow symlinks. So the command to copy the ssl
directory to `/tmp/ssl-bkp` results in a new symlink being created
`/tmp/ssl-bkp/ca` that points to `/etc/puppetlabs/puppetserver/ca`.

So later the test tries to copy the backup to the destination:

    cp /tmp/ssl-bkp/ca/ca_crl.pem /etc/puppetlabs/puppet/ssl/ca/ca_crl.pem

and it fails because the source and destination files are actually the same
file (inode).

Instead use `cp -rL` which follows symlinks. As a result of the change, the
backup directory contains a copy of the ca directory after following symlinks.

### DIFF
--- a/acceptance/tests/crl_revocation.rb
+++ b/acceptance/tests/crl_revocation.rb
@@ -28,7 +28,7 @@ test_name 'When certs have been revoked in CRL' do
   end
 
   step 'Create a CRL with the master/broker cert revoked' do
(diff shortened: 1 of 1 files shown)


donoghuc (approved):

joshcooper:
Jenkins please test this

PR #826, opened 2020-11-24
(PA-3490) Add syslog logging to pxp-agent

GabrielNagy:
This is mostly for AIX where we have to log through its syslog daemon in
order to benefit from log rotation.

To enable logging to syslog, the user needs to specify "syslog" as the
value for the "logfile" option.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -282,41 +282,6 @@ HW::ParseResult Configuration::parseOptions(int argc, char *argv[])
     return parse_result;
 }
 
-#ifdef HAS_LTH_SYSLOG
-lth_log::syslog_facility string_to_syslog_facility(std::string facility)
-{
-    lth_log::syslog_facility fac;
-    try {
-        // NOTE(ale): ignoring HorseWhisperer's vlevel ("-v" flag)
-        const std::map<std::string, lth_log::syslog_facility> option_to_syslog_facility {
-            { "kern", lth_log::syslog_facility::kern },
-            { "user", lth_log::syslog_facility::user },
-            { "mail", lth_log::syslog_facility::mail },
-            { "daemon", lth_log::syslog_facility::daemon },
-            { "auth", lth_log::syslog_facility::auth },
-            { "syslog", lth_log::syslog_facility::syslog },
-            { "lpr", lth_log::syslog_facility::lpr },
-            { "news", lth_log::syslog_facility::news },
-            { "uucp", lth_log::syslog_facility::uucp },
-            { "cron", lth_log::syslog_facility::cron },
-            { "local0", lth_log::syslog_facility::local0 },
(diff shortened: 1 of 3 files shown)


lib/src/configuration.cc
```diff
@@ -337,6 +338,25 @@ std::string Configuration::setupEventLogLogging(std::string loglevel)
...
 }
 
+std::string Configuration::setupSyslogLogging(std::string loglevel)
+{
+#ifdef _WIN32
+    throw Configuration::Error {
+      lth_loc::format("eventlog is available only on POSIX platforms") };
+#else  // _WIN32
+#ifdef HAS_LTH_SYSLOG
+    lth_log::setup_syslog_logging("pxp-agent");
```
GabrielNagy, line 348: This initializes the logger with a garbled ID, such as `/ò*[10485804]` instead of `pxp-agent[10485804]`, and I don't understand why.

GabrielNagy: https://stackoverflow.com/a/27627466/6648276

TIL

GabrielNagy:
adhoc build: http://builds.delivery.puppetlabs.net/puppet-agent/2e66387bd0db56b22eaf2f28fe423bdebaa52ba8/artifacts/aix/7.1/puppet6/ppc/puppet-agent-6.4.2.3492.g2e66387b-1.aix7.1.ppc.rpm

lib/src/configuration.cc
```diff
@@ -337,6 +375,34 @@ std::string Configuration::setupEventLogLogging(std::string loglevel)
...
+#ifdef HAS_LTH_SYSLOG
+    std::string facility = HW::GetFlag<std::string>("syslog-facility");
+    lth_log::syslog_facility fac = string_to_syslog_facility(facility);
+
+    // Some syslog implementations won't write to a non-existent logfile,
+    // so create an empty file if it does not exist.
+    if (!fs::exists(logfile_)) {
+        boost::nowide::ofstream file(logfile_.c_str());
+    }
+
```
ciprianbadescu, line 393: I'm not sure about this one, since you don't really know where the log will be redirected now.
A better place may be where the syslog configures facility->file mapping

lib/src/configuration.cc
```diff
@@ -280,6 +282,41 @@ HW::ParseResult Configuration::parseOptions(int argc, char *argv[])
...
+        };
+        fac = option_to_syslog_facility.at(facility);
+    } catch (const std::out_of_range& e) {
+        throw Configuration::Error {
+            lth_loc::format("invalid syslog facility: '{1}'", facility) };
+    }
+    return fac;
+}
+#endif
+
```
ciprianbadescu, line 319: I think this may better belong in leatherman, and pass string to `setup_syslog_logging`

ciprianbadescu (approved):

PR #828, opened 2020-12-11
(PE-30708) Support `both` as valid task input method

donoghuc:
The default input method is already `both` (when there is no entry for input_method). This commit adds support for explicitly specifying `both` in task metadata.

### DIFF
--- a/lib/src/modules/task.cc
+++ b/lib/src/modules/task.cc
@@ -455,7 +455,7 @@ ActionResponse Task::callAction(const ActionRequest& request)
     auto task_input_method = task_execution_params.includes("input_method") ?
         task_execution_params.get<std::string>("input_method") : std::string{""};
 
-    static std::set<std::string> input_methods{{"stdin", "environment", "powershell"}};
+    static std::set<std::string> input_methods{{"stdin", "environment", "powershell", "both"}};
     if (!task_input_method.empty() && input_methods.count(task_input_method) == 0) {
         throw Module::ProcessingError {
             lth_loc::format("unsupported task input method: {1}", task_input_method) };
@@ -469,9 +469,9 @@ ActionResponse Task::callAction(const ActionRequest& request)
                                        client_,
                                        task_execution_params.get<std::vector<lth_jc::JsonContainer>>("files"));
 
-    // Use powershell input method by default if task uses .ps1 extension.
-    if (task_input_method.empty() && task_file.extension().string() == ".ps1") {
-        task_input_method = "powershell";
+    // If input_method is unset use the default "powershell" for a powershell task or "both" for any other task
+    if (task_input_method.empty()) {
+      task_input_method = task_file.extension().string() == ".ps1" ? "powershell" : "both";
     }
 
     std::map<std::string, std::string> task_environment;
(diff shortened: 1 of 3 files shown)


donoghuc:
@McdonaldSeanp looks like my unit test is passing on appveyor but there is some pester test issue. able to run that locally?

lib/src/modules/task.cc
```diff
@@ -485,11 +485,11 @@ ActionResponse Task::callAction(const ActionRequest& request)
         // Pass input on stdin ($input)
         task_input = task_execution_params.get<lth_jc::JsonContainer>("input").toString();
     } else {
-        if (task_input_method.empty() || task_input_method == "stdin") {
```
nicklewis, line 488: I think this `task_input_method.empty()` is implementing the `both` default. It might be better to actually _set_ `task_input_method` to `both` if it's empty (I think somewhere around here we already default it to `powershell` in some cases, so there's a pattern for it) and then not have to check `empty()` anymore.

donoghuc: wow, that is a much better solution. thanks

lib/src/modules/task.cc
```diff
@@ -453,9 +453,9 @@ ActionResponse Task::callAction(const ActionRequest& request)
 {
     auto task_execution_params = request.params();
     auto task_input_method = task_execution_params.includes("input_method") ?
-        task_execution_params.get<std::string>("input_method") : std::string{""};
+        task_execution_params.get<std::string>("input_method") : std::string{"both"};
```
nicklewis, line 456: Applying this default _here_ breaks the line where we set `powershell` as the default for `.ps1` files 

https://github.com/puppetlabs/pxp-agent/blob/1.9.x/lib/src/modules/task.cc#L473

I think we probably want to keep it as an empty string _here_ and then set `both` just _after_ we would potentially set `powershell`. So that code would become: if it has no default, then set it to `powershell` if it's `.ps1` and `both` otherwise.

That would also address the _other_ issue this introduces which is that if the task metadata explicitly sets `"input_method": ""`, it was previously treated as `both` but will now be treated as an unknown input method. Admittedly, there's no good _reason_ to set it to that, but it is a behavior change. I will check what Bolt does in that case. An error may actually be consistent.

donoghuc: Ah, that is what i get for updating this with half my attention on it. :flushed: will update when i can actually focus on it.

lib/src/modules/task.cc
```diff
@@ -474,6 +474,11 @@ ActionResponse Task::callAction(const ActionRequest& request)
         task_input_method = "powershell";
     }
 
+    // If input_method is unset use the default "both"
```
nicklewis, line 477: Should this be folded into the above conditional?

```cpp
if (task_input_method.empty()) {
  if (task_file.extension().string() == ".ps1") {
    task_input_method = "powershell";
  } else {
    task_input_method == "both";
  }
}
```
or
```cpp
if (task_input_method.empty()) {
  task_input_method = task_file.extension().string() == ".ps1" ? "powershell" : "both";
}
```

nicklewis (approved):

PR #829, opened 2020-12-15
(maint) Merge 1.9.x to master

ciprianbadescu:
* commit '40a666c4a7b5cea133892a3be6f24b51e164f791':
  (maint) Pin back pester version
  (PE-30708) Support `both` as valid task input method
  (packaging) Bump to version '1.9.15' [no-promote]

 * lib/tests/unit/modules/task_test.cc - manually adapted

Conflicts:
 * CMakeLists.txt - keep main version
 * appveyor.yml - pin pester version
 * lib/src/modules/task.cc - manually solved conflicts

### DIFF
--- a/lib/tests/unit/modules/task_test.cc
+++ b/lib/tests/unit/modules/task_test.cc
@@ -292,6 +292,35 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
 #endif
     }
 
+    SECTION("passes input on both stdin and env when method is both") {
+        auto echo_txt =
+#ifndef _WIN32
+            (DATA_FORMAT % "\"0632\""
+                         % "\"task\""
+                         % "\"run\""
+                         % "{\"task\":\"test::both\", \"input\":{\"message\":\"hello\"}, \"input_method\": \"both\", \"files\" : [{\"sha256\": \"823c013467ce03b12dbe005757a6c842894373e8bcfb0cf879329afb5abcd543\", \"filename\": \"multi\"}]}").str();
+#else
+            (DATA_FORMAT % "\"0632\""
+                         % "\"task\""
+                         % "\"run\""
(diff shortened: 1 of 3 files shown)


lib/tests/unit/modules/task_test.cc
```diff
@@ -292,6 +292,35 @@ TEST_CASE("Modules::Task::executeAction", "[modules][output]") {
 #endif
     }
 
+    SECTION("passes input on both stdin and env when method is both") {
+        auto echo_txt =
+#ifndef _WIN32
+            (DATA_FORMAT % "\"0632\""
+                         % "\"task\""
+                         % "\"run\""
+                         % "{\"task\":\"test::both\", \"input\":{\"message\":\"hello\"}, \"input_method\": \"both\", \"files\" : [{\"sha256\": \"823c013467ce03b12dbe005757a6c842894373e8bcfb0cf879329afb5abcd543\", \"filename\": \"multi\"}]}").str();
```
ciprianbadescu, line 301: I don't know if it was necessary, but test failed until I added `task` key

donoghuc: an yeah this is required in main https://github.com/puppetlabs/pxp-agent/blob/dd5b091971e97864873a0644cc58caeb2f2e88f6/lib/src/modules/task.cc#L356 i'm kind of suprised how different the task implementation is between the two. perhaps i targetted the wrong stream here... @McdonaldSeanp

McdonaldSeanp: @donoghuc yeah, you probably could have just targeted main since it looks like pxp-agent main is going to both agent 6.x and 7.x streams.

Your PR would have landed changes for pxp-agent 5.x, which would be PE 2018.1.

It was likely unnecessary to fix this for 2018.1, but I don't think it hurts. If these changes still work for main we're fine.

donoghuc: ah, my bad. i thought 1.9.x went to 6 and main went to 7. backporting this wont hurt anything. thanks.

donoghuc (approved):

PR #830, opened 2021-02-18
(PE-30624) Deprecate master_uris, add primary_uris

barriserloth:
This commit updates pxp-agent as part of the work to deprecate the
`master_uris` parameter. As of this commit, using `master_uris` should
still work, but using the new `primary_uris` should also now work.
pxp-agent will now look for both of those config values when setting
that param.

### DIFF
--- a/lib/src/configuration.cc
+++ b/lib/src/configuration.cc
@@ -230,9 +230,9 @@ HW::ParseResult parseArguments(const int argc, char* const argv[])
 
 HW::ParseResult Configuration::parseOptions(int argc, char *argv[])
 {
-    // If parsing options, clear the previous value for master_uris.
+    // If parsing options, clear the previous value for primary_uris.
     // This is primarily for testing.
-    master_uris_.clear();
+    primary_uris_.clear();
 
     // Remember the path to the pxp-agent executable used to start
     // this process, it is supposed to be used to start executables
@@ -440,7 +440,7 @@ const Configuration::Agent& Configuration::getAgentConfiguration() const
     agent_configuration_ = Configuration::Agent {
         HW::GetFlag<std::string>("modules-dir"),
         broker_ws_uris_,
-        master_uris_,
+        primary_uris_,
         HW::GetFlag<std::string>("pcp-version"),
         HW::GetFlag<std::string>("ssl-ca-cert"),
         HW::GetFlag<std::string>("ssl-cert"),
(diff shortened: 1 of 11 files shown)


donoghuc (commented):
I think we should update tests to use the new setting. We can also have a test that shows master-uri still works.

barriserloth:
Sounds good, also it looks like the make commands are failing for reasons I don't understand and I'm not sure if that's related to my changes or not so I'll take a look

lib/src/configuration.cc
```diff
@@ -861,16 +861,16 @@ void Configuration::parseConfigFile()
...
-            // TODO: add `servers`, which will populate broker-ws-uris and master-uris
+            // TODO: add `servers`, which will populate broker-ws-uris and primary-uris
             //       using default ports; expects hostnames only
             if (key == "broker-ws-uris") {
                 wrap(key, lth_loc::translate("array of strings"), [&]() {
                     broker_ws_uris_ = config->get_string_list(key);
                 });
                 continue;
-            } else if (key == "master-uris") {
+            } else if (key == "master-uris" || key == "primary-uris") {
```
nicklewis, line 871: It looks like whichever one is parsed last will win here. I think we should try to prefer `primary-uris` if they're both set.

barriserloth:
After copious amount of force-pushing, I think this PR is now ready for review again

donoghuc (changes requested):
Sorry @barriserloth i think we got confused. i was thinking that 1.9.x fed in to 6.x and main into 7, but it looks like main goes to both 6 and 7. So i think you need to target main. Also when you do i think you can make some updates in the task and apply module to just go ahead and use primary instead of master in the source there as well.

barriserloth:
@donoghuc Cool, should be updated now

lib/src/modules/apply.cc
```diff
@@ -265,7 +265,8 @@ exit exit_code
 
         const auto plugin_cache = module_cache_dir_->createCacheDir(plugin_cache_name);
         params.set<std::string>("plugin_cache", plugin_cache.string());
-        params.set<std::vector<std::string>>("master_uris", master_uris_);
+        params.set<std::vector<std::string>>("master_uris", primary_uris_);
```
donoghuc, line 268: I dont think we need this anymore.

donoghuc:
Jenkins please test this

lib/src/modules/apply.cc
barriserloth: I wasn't sure if removing this would cause some issue with backwards compatibility since I don't fully understand how it gets used, but I can try dropping it

donoghuc: you pass whatever you want the arg name through and its used https://github.com/puppetlabs/pxp-agent/pull/830/commits/b4ee333f111dbbeb7ae72dfe42cfaeca15299d3a#diff-47d2fffa351567b475e3243c3a80d4b751e29a9057e424a3709cebee6ebfc47fR78 so just get rid of master_uris in both places.

barriserloth: Ah yep that's easy. Thanks

barriserloth: Done!

barriserloth:
Jenkins please test this

donoghuc (approved):

PR #831, opened 2021-05-03
(PE-31415) Set apply ruby shim environment to passed-in environment

ekinanp:
This ensures that spurious messages of the form 'Local environment..
doesn't match...' are not printed when the apply ruby shim is executed
and the passed-in environment is different from production.

Signed-off-by: Enis Inan <enis.inan@puppet.com>


ekinanp:
I tested this by adding the exact line here to the apply ruby shim on a master and things worked. The message that was previously printed is no longer printed so this should be good to go.

PR #832, opened 2021-05-04
(PE-31871) Extract apply shim to a file

nicklewis:
Previously, this shim existed only as a hardcoded Ruby string in a C++ 
source file. That file would be written when needed and purged after a 
period of disuse. But changes to the content weren't necessarily picked up,
as it would only be rewritten after being purged for inactivity. That led
to a problem where pxp-agent installations which had been upgraded would
fail to run apply/apply_prep, because they had an old copy of the shim. The
old version expected to receive a master_uris argument but pxp-agent had
been updated to pass primary_uris instead.

We now manage this file as a regular file and include it during the 
build/packaging process so that it will be automatically upgraded by the 
package manager.

### DIFF
--- a/exe/CMakeLists.txt
+++ b/exe/CMakeLists.txt
@@ -9,6 +9,9 @@ add_executable(pxp-agent main.cc)
 target_link_libraries(pxp-agent libpxp-agent)
 install(TARGETS pxp-agent DESTINATION bin)
 
+install(FILES apply_ruby_shim.rb DESTINATION bin
+    PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+
 set(EXECUTION_WRAPPER_LIBS ${Boost_LIBRARIES} ${LEATHERMAN_LIBRARIES})
 if (CMAKE_SYSTEM_NAME MATCHES "AIX")
     find_package(Threads)
--- a/exe/apply_ruby_shim.rb
+++ b/exe/apply_ruby_shim.rb
@@ -1,0 +1,148 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'json'
+require 'puppet'
+require 'puppet/configurer'
+require 'securerandom'
+require 'tempfile'
+require 'uri'
+
+## TODO: Option to read from a file is for debugging. Only read from stdin in production.
+args = JSON.parse(ARGV[0] ? File.read(ARGV[0]) : STDIN.read)
+
+# Create temporary directories for all core Puppet settings so we don't clobber
+# existing state or read from puppet.conf. Also create a temporary modulepath.
+# Additionally include rundir, which gets its own initialization.
+puppet_root = Dir.mktmpdir
+moduledir = File.join(puppet_root, 'modules')
+Dir.mkdir(moduledir)
+plugin_cache = args['plugin_cache']
+plugin_dest = File.join(plugin_cache, 'plugins')
+pluginfact_dest = File.join(plugin_cache, 'pluginfacts')
+# Use isolated directories in puppet_root
+setting_keys = Puppet::Settings::REQUIRED_APP_SETTINGS.append(:rundir)
+cli_base = (setting_keys).flat_map do |setting|
+  ["--#{setting}", File.join(puppet_root, setting.to_s.chomp('dir'))]
+end
+cli_base.concat([
+  '--modulepath',
+  moduledir,
+  '--plugindest',
+  plugin_dest,
+  '--pluginfactdest',
+  pluginfact_dest
+])
+# There will always be at least a single master URI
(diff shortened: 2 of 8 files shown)


nicklewis:
This depends on, but does not block, https://github.com/puppetlabs/pxp-agent-vanagon/pull/23.

nicklewis:
Either this PR or https://github.com/puppetlabs/pxp-agent/pull/831 will need to be rebased after the other is merged to ensure that the change is picked up by the new file.

ekinanp (approved):
Changes here seem to make sense.

exe/CMakeLists.txt
```diff
@@ -9,6 +9,9 @@ add_executable(pxp-agent main.cc)
 target_link_libraries(pxp-agent libpxp-agent)
 install(TARGETS pxp-agent DESTINATION bin)
 
+install(FILES apply_ruby_shim.rb DESTINATION bin
+    PERMISSIONS OWNER_WRITE OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
```
McdonaldSeanp, line 13: I don't know that it's actually any sort of risk, but is there a reason to allow world read/execute?

McdonaldSeanp (changes requested):
Just one question

exe/CMakeLists.txt
nicklewis: I copied those permissions from the ps1 helpers https://github.com/puppetlabs/pxp-agent/blob/main/exe/CMakeLists.txt#L34-L35 They also match the permissions for everything else in that directory

```
# ls -l /opt/puppetlabs/puppet/bin/
total 3588
-rwxr-xr-x 1 root root    4820 May  4 19:33 apply_ruby_shim.rb
-rwxr-xr-x 1 root root   19944 May  4 19:45 augmatch
-rwxr-xr-x 1 root root   11384 May  4 19:45 augparse
-rwxr-xr-x 1 root root   24200 May  4 19:45 augtool
-rwxr-xr-x 1 root root     544 Apr 28 18:12 bundle
-rwxr-xr-x 1 root root     546 Apr 28 18:12 bundler
-rwxr-xr-x 1 root root    6224 Apr 28 18:07 c_rehash
-rwxr-xr-x 1 root root  223200 May  4 19:44 curl
-rwxr-xr-x 1 root root    6071 Apr 28 18:10 curl-config
-rwxr-xr-x 1 root root    5078 Apr 28 18:12 erb
-rwxr-xr-x 1 root root  172192 May  4 19:44 execution_wrapper
lrwxrwxrwx 1 root root      53 May  4 20:21 eyaml -> /opt/puppetlabs/puppet/lib/ruby/vendor_gems/bin/eyaml
-rwxr-xr-x 1 root root     270 May  4 19:44 facter
-rwxr-xr-x 1 root root   15624 May  4 19:45 fadot
-rwxr-xr-x 1 root root     560 Apr 28 18:12 gem
-rwxr-xr-x 1 root root    6448 May  4 19:44 hiera
-rwxr-xr-x 1 root root     522 Apr 28 18:12 irb
-rwxr-xr-x 1 root root  653600 May  4 19:44 openssl
-rwxr-xr-x 1 root root     179 May  4 19:44 puppet
-rwxr-xr-x 1 root root     590 Apr 12 19:42 puppet-backup
-rwxr-xr-x 1 root root 2290880 May  4 19:44 pxp-agent
-rwxr-xr-x 1 root root     528 Apr 27 02:24 r10k
-rwxr-xr-x 1 root root     601 Apr 28 18:12 racc
-rwxr-xr-x 1 root root     605 Apr 28 18:12 racc2y
-rwxr-xr-x 1 root root     528 Apr 28 18:12 rake
-rwxr-xr-x 1 root root     528 Apr 28 18:12 rdoc
-rwxr-xr-x 1 root root     524 Apr 28 18:12 ri
-rwxr-xr-x 1 root root    7200 May  4 19:45 ruby
-rwxr-xr-x 1 root root     528 Apr 28 18:14 thor
-rwxr-xr-x 1 root root     939 May  4 19:44 wrapper.sh
-rwxr-xr-x 1 root root    1666 Apr 28 18:12 xml2-config
-rwxr-xr-x 1 root root   19808 May  4 19:44 xmlcatalog
-rwxr-xr-x 1 root root   75856 May  4 19:44 xmllint
-rwxr-xr-x 1 root root    2450 Apr 28 18:13 xslt-config
-rwxr-xr-x 1 root root   28200 May  4 19:44 xsltproc
-rwxr-xr-x 1 root root     605 Apr 28 18:12 y2racc
```

I don't _really_ know why, but it seems to be the norm.

nicklewis: I guess not having that permission would create friction if we wanted to run pxp-agent as non-root. We don't support that now, but it seems like a reasonable practice for the installed binaries to be runnable by anyone.

McdonaldSeanp: 👍 cool, that's a reasonable reason

McdonaldSeanp (approved):

PR #833, opened 2021-05-19
(PE-31216) Disallow null bytes in task output

nicklewis:
Previously, pxp-agent would accept task output containing null bytes and
would return it to orchestrator. This caused a database error as null
bytes cannot be written to postgres. These null bytes are also not
allowed by the task spec, so this behavior was incorrect.

The root issue is that the isValidUTF8 helper, which uses rapidjson to
validate each character, considers null bytes to be valid. Technically
this isn't wrong, as the null character is a UTF-8 character, but it's
not a valid character in a null-terminated string.

We now supplement the regular validity check with an additional check
that the string doesn't contain any null bytes.

This also fixes a bug where we could return a result containing invalid
UTF-8 in the case where both stdout and stderr had invalid UTF-8 output.
We would validate that stdout had invalid UTF-8 and would generate an
error including the value from stderr without validating it as UTF-8. We
now simply omit the stderr output as it's unlikely to be useful in this
case.

### DIFF
--- a/lib/src/util/bolt_module.cc
+++ b/lib/src/util/bolt_module.cc
@@ -74,11 +74,8 @@ void BoltModule::processOutputAndUpdateMetadata(PXPAgent::ActionResponse &respon
                   response.prettyRequestLabel(), output);
         std::string execution_error {
                 lth_loc::format("The task executed for the {1} returned invalid "
-                                "UTF-8 on stdout - stderr:{2}",
-                                response.prettyRequestLabel(),
-                                (response.output.std_err.empty()
-                                 ? lth_loc::translate(" (empty)")
-                                 : "\n" + response.output.std_err)) };
+                                "UTF-8 on stdout",
+                                response.prettyRequestLabel()) };
         response.setBadResultsAndEnd(execution_error);
     }
 }
--- a/lib/src/util/utf8.cc
+++ b/lib/src/util/utf8.cc
@@ -4,6 +4,7 @@
 #include <rapidjson/stream.h>
 #endif
 
+#include <algorithm>
 #include <string>
 
 namespace PXPAgent {
@@ -18,7 +19,15 @@ namespace Util {
                 return false;
             }
         }
-        return true;
+
+        // rapidjson::UTF8<char>::Validate accepts null bytes as valid UTF-8.
+        // They technically are valid since they're the null character. But
+        // null characters aren't valid in a string, so we want to disallow
+        // them regardless.
+        bool has_null = std::any_of(s.begin(), s.end(), [](char c) {
+                            return c == 0;
+                        });
+        return !has_null;
     }
 }  // namespace Util
 }  // namespace PXPAgent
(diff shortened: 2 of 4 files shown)


lib/src/util/utf8.cc
```diff
@@ -18,7 +19,15 @@ namespace Util {
...
             }
         }
-        return true;
+
+        // rapidjson::UTF8<char>::Validate accepts null bytes as valid UTF-8.
+        // They technically are valid since they're the null character. But
+        // null characters aren't valid in a string, so we want to disallow
+        // them regardless.
+        bool has_null = std::any_of(s.begin(), s.end(), [](char c) {
+                            return c == 0;
```
nicklewis, line 28: Note that the only time a null byte is allowed in UTF-8 is for a single-byte null character. A null byte is _not_ allowed as part of a multibyte character.

https://en.wikipedia.org/wiki/UTF-8#Encoding

nicklewis:
Tested this against https://github.com/puppetlabs/orchestration-control/blob/6e2f92749e69446bf4abcf371755a8576d69f19c/site/bad_output/tasks/init.rb which was cited on the ticket as an example of a task that causes this error.

```
Errored on wily-invention.delivery.puppetlabs.net
  Error: The task executed for the non blocking 'task run' request (transaction 9) returned invalid UTF-8 on stdout
```

donoghuc:
jenkins please test this on ubuntu1804-64a,windows2016-64a

lib/src/util/bolt_module.cc
```diff
@@ -74,11 +74,8 @@ void BoltModule::processOutputAndUpdateMetadata(PXPAgent::ActionResponse &respon
                   response.prettyRequestLabel(), output);
         std::string execution_error {
                 lth_loc::format("The task executed for the {1} returned invalid "
-                                "UTF-8 on stdout - stderr:{2}",
-                                response.prettyRequestLabel(),
-                                (response.output.std_err.empty()
-                                 ? lth_loc::translate(" (empty)")
-                                 : "\n" + response.output.std_err)) };
```
donoghuc, line 81: So a change here will be the error message will no longer attempt to print out stderr? Seems fine

nicklewis: Yeah. Specifically there was a bug here where if both stdout _and_ stderr included invalid UTF-8, the result would include invalid UTF-8 because it included stderr without first validating it. I opted to just remove the stderr bit rather than adding that validation, on the basis that a) it seems likely that if stdout is invalid UTF-8, stderr probably is too, b) including stderr doesn't seem especially valuable in this case (since the task didn't _fail_, stderr isn't going to help explain what the problem was), c) I'm lazy.

lib/src/util/utf8.cc
```diff
@@ -18,7 +19,15 @@ namespace Util {
...
+                        });
+        return !has_null;
```
donoghuc, line 30: Ah, so `std::any_of` will be false for an empty string http://www.cplusplus.com/reference/algorithm/any_of/  nice

donoghuc (approved):
This looks good to me, assuming the acceptance tests pass (wanted to make sure we were not matching on a message that is expecting stderr in an error message) i think this is g2g.

nicklewis:
jenkins please test this on ubuntu1804-64a,windows2016-64a

donoghuc:
jenkins please test this on ubuntu1804-64a,windows2016-64a

nicklewis:
jenkins please test this on ubuntu1804-64a,windows2016-64a

nicklewis:
jenkins please test this on ubuntu1804-64a,windows2016-64a

PR #835, opened 2021-09-03
(PA-3974) Increase DH key size for pcp-broker

GabrielNagy:
Add the 'jdk.tls.ephemeralDHKeySize=2048' JVM option when running pcp-broker, as the default DH key size is too small for openssl 1.1.1k FIPS.

### DIFF
--- a/acceptance/lib/pxp-agent/test_helper.rb
+++ b/acceptance/lib/pxp-agent/test_helper.rb
@@ -57,10 +57,6 @@ def expect_file_on_host_to_contain(host, file, expected, seconds=30)
   end
 end
 
-def append_jvm_limits(command)
-  "export JVM_OPTS=\"-Xms2g -Xmx2g\"; export LEIN_JVM_OPTS=\"-Xms2g -Xmx2g\"; #{command}"
-end
-
 # Some helpers for working with a pcp-broker 'lein tk' instance
 def run_pcp_broker(host, instance=0)
   timeout = 120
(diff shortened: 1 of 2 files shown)


McdonaldSeanp (approved):

acceptance/lib/pxp-agent/test_helper.rb
```diff
@@ -68,7 +72,8 @@ def run_pcp_broker(host, instance=0)
 
   lein_command = "cd #{GIT_CLONE_FOLDER}/pcp-broker#{instance}; export LEIN_ROOT=ok; \
      lein with-profile #{LEIN_PROFILE} tk </dev/null >/var/log/puppetlabs/pcp-broker.log.#{SecureRandom.uuid} 2>&1 &"
-  lein_command = append_jvm_limits(lein_command) if host[:gke_container]
+  lein_command = append_jvm_opts('-Xms2g -Xmx2g', lein_command) if host[:gke_container]
```
donoghuc, line 75: Is this method used anywhere else? Breaking this in to a method that does some complicated string manipulation makes this pretty hard to reason about. It looks like the only special case is if the host is a GKE cluster right? So could we just do something simple like:

```ruby
jvm_opts = ['-Djdk.tls.ephemeralDHKeySize=2048']
jvm_opts << '-Xms2g -Xmx2g' if host[:gke_container]
lein_command = "export JVM_OPTS='#{jvm_opts.join(' ')}'; export LEIN_JVM_OPTS='#{jvm_opts.join(' ')}'; #{lein_command}"
'''
```

acceptance/setup/common/050_Setup_Broker.rb
```diff
@@ -55,7 +59,7 @@
   # some time and will eat into the polling period we allow for the broker to start
   for i in 0..NUM_BROKERS-1
     lein_command = "cd #{GIT_CLONE_FOLDER}/pcp-broker#{i}; export LEIN_ROOT=ok; lein with-profile #{LEIN_PROFILE} deps"
-    lein_command = append_jvm_limits(lein_command) if master[:gke_container]
+    lein_command = append_jvm_opts('-Xms2g -Xmx2g', lein_command) if master[:gke_container]
```
donoghuc, line 62: ah, i see this is where that method is used. I think its simple enough just to conditionally append it here without using a general method https://github.com/puppetlabs/pxp-agent/pull/835/files#r701994227

donoghuc:
Thanks so much @GabrielNagy for tracking this down! Excellent work.

steveax:
Yes, thank you! 🙏

PR #836, opened 2021-10-28
(PCP-906) Configure WebSocket++ loggers

nicklewis:
We now pass the log level and stream through to the cpp-pcp-client
Connector classes which in turn use them to configure logging on the
websocketpp connection. This entails storing the parsed log level on the
Configuration::Agent object and providing a method to retrieve the log
stream object.

This depends on https://github.com/puppetlabs/cpp-pcp-client/pull/257

### DIFF
--- a/lib/tests/common/mock_connector.hpp
+++ b/lib/tests/common/mock_connector.hpp
@@ -8,6 +8,10 @@
 #include <pxp-agent/action_request.hpp>
 #include <pxp-agent/action_response.hpp>
 
+#define LEATHERMAN_LOGGING_NAMESPACE "puppetlabs.pxp_agent.test"
+
+#include <leatherman/logging/logging.hpp>
+
 #include <atomic>
 #include <vector>
 #include <string>
@@ -55,7 +59,8 @@ static Configuration::Agent AGENT_CONFIGURATION { MODULES,
                                                   15,    // ping interval
                                                   30,    // task download connection timeout
                                                   120,   // task download timeout
-                                                  64 * 1024 * 1024 };  // default max-message-size
+                                                  64 * 1024 * 1024,  // default max-message-size
+                                                  leatherman::logging::log_level::none };
 
 static const std::string VALID_ENVELOPE_TXT {
     " { \"id\" : \"123456\","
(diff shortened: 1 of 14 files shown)


beechtom:
The PXPConnector initializer lists need to be updated for the unit tests [here](https://github.com/nicklewis/pxp-agent/blob/PCP-906/lib/tests/unit/pxp_connector_v1_test.cc#L11) and [here](https://github.com/nicklewis/pxp-agent/blob/PCP-906/lib/tests/unit/pxp_connector_v2_test.cc#L11).

lib/inc/pxp-agent/pxp_connector.hpp
```diff
@@ -9,6 +9,12 @@ struct ParsedChunks;
 class Schema;
 }  // namespace PCPClient
 
+namespace leatherman {
+  namespace logging {
+    enum class log_level;
+  }
+}
+
```
beechtom, line 17:
```suggestion
```

I don't think this is needed anymore?

nicklewis:
Updated the tests and removed that unneeded forward declaration

lib/tests/common/mock_connector.hpp
```diff
@@ -55,7 +59,8 @@ static Configuration::Agent AGENT_CONFIGURATION { MODULES,
                                                   15,    // ping interval
                                                   30,    // task download connection timeout
                                                   120,   // task download timeout
-                                                  64 * 1024 * 1024 };  // default max-message-size
+                                                  64 * 1024 * 1024, // default max-message-size
```
beechtom, line 62:
```suggestion
                                                  64 * 1024 * 1024,  // default max-message-size
```

Just to make the linter happy. 🙃

beechtom (approved):
Looks like tests aren't passing since they're not picking up the changes in cpp-pcp-client. Not sure if that'll be fixed automatically now that the changes there are merged.

Other than that, lgtm. 👍

nicklewis:
It looks like the version of cpp-pcp-client is set in .travis.yml and will need to be updated once that change is released. I have no idea how that works though...

nicklewis:
I tagged cpp-pcp-client which should make a release happen automatically, so I updated the version here. Will need to rekick once the release has actually gone through though.

PR #838, opened 2022-05-05
(PA-4487) Add make in the pre-suite

cthorn42:
If a VM doesn't have make installed then Lein will fail to build. Historically our EL platforms
had make installed my default, but the new GCE images we'll be switching to does not have it installed.


cthorn42:
There is a pcp-broker log in the description of PA-4487 that shows what happens when make isn't installed and the pre-suite is ran.

joshcooper (approved):

PR #840, opened 2023-02-03
(PA-4906) Updates SSL test, acceptance Gemfile

mhashizume:


acceptance/Gemfile
```diff
@@ -20,14 +18,14 @@ gem "beaker-vagrant", *location_for(ENV['BEAKER_VAGRANT_VERSION'] || "~> 0")
 gem "beaker-vmpooler", *location_for(ENV['BEAKER_VMPOOLER_VERSION'] || "~> 1")
 gem "beaker-gke", *location_for(ENV['BEAKER_GKE_VERSION'] || "~> 0.0.3")
 gem 'rake'
-gem 'scooter', '~> 3.2.17'
+gem 'scooter', '~> 4'
```
joshcooper, line 21:
```suggestion
gem 'scooter', '~> 4.0'
```
So that we don't install 5.0 when it comes out?

mhashizume: Good catch, thank you!

joshcooper (approved):

acceptance/Gemfile
```diff
@@ -1,11 +1,9 @@
-# Specifies a gem mirror; duplicated in acceptance setup
-# to ensure a similar environment on acceptance hosts.
-source ENV['GEM_SOURCE'] || 'https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/'
+source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 def location_for(place, fake_version = nil)
-  if place =~ /^((?:git[:@]|https:)[^#]*)#(.*)/
+  if place.is_a?(String) && place =~ /^((?:git[:@]|https:)[^#]*)#(.*)/
```
donoghuc, line 4: I think we have defaults for these being strings. I've had luck with this definition:

```ruby
def location_for(place, fake_version = nil)
  case place
  when /^(git:[^#]*)#(.*)/
    [fake_version, { git: Regexp.last_match(1), branch: Regexp.last_match(2), require: false }].compact
  when %r{^file://(.*)}
    ['>= 0', { path: File.expand_path(Regexp.last_match(1)), require: false }]
  else
    [place, { require: false }]
  end
end
```

acceptance/Gemfile
```diff
@@ -1,11 +1,9 @@
-# Specifies a gem mirror; duplicated in acceptance setup
-# to ensure a similar environment on acceptance hosts.
-source ENV['GEM_SOURCE'] || 'https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/'
+source ENV['GEM_SOURCE'] || 'https://rubygems.org'
```
donoghuc, line 1: What was the problem with defaulting to artifactory?

acceptance/tests/invalid_ssl_config.rb
```diff
@@ -29,9 +29,9 @@
   end
 
   step 'Create alternate CA on the master and copy to agent temp folder' do
-    master_temp_ca_dir = master.tmpdir('alternate_ssl')
+    master_temp_ca_dir = master.tmpdir('alternate_ca')
     master_temp_conf = File.join(master_temp_ca_dir, 'puppet.conf')
-    create_remote_file(master, master_temp_conf, "ssldir = #{master_temp_ca_dir}")
+    create_remote_file(master, master_temp_conf, "cadir = #{master_temp_ca_dir}\nssldir = #{master_temp_ca_dir}")
```
donoghuc, line 34: Ah, so configuring both `cadir and ssldir` to the same location yields the "old" behavior?

mhashizume: Yes, before puppetserver 7 the cadir and ssldir were in the same location. This change preserves that behavior, as the new behavior (separating them out into `etc/puppetlabs/puppet/ssl/ca` and
`/etc/puppetlabs/puppetserver/ca`) doesn't really matter for the sake of this test.

acceptance/Gemfile
mhashizume: Definitely not a necessary change, but it's a convention that we follow in most of the other Puppet repositories I've seen (Puppet, Facter, Hiera). I assume the original intent was that you can set `GEM_SOURCE` in CI for Artifactory, but if you're outside of the Puppet internal network you can fall back to rubygems.org, which should be available anywhere.

acceptance/Gemfile
mhashizume: Yeah, typically `place` _should_ be a string, though I have run into people using arrays for gem versions ([I ran into this](https://github.com/voxpupuli/beaker-puppet/blob/94be1761ea48d775509a2c3dfe38d41d3ecb6786/Gemfile#L17) last week). I was modeling this off of what we did in Puppet a few years back: https://github.com/puppetlabs/puppet/commit/d84b38898527d2c94c1112402a085729c81a6236

donoghuc (approved):

acceptance/Gemfile
donoghuc: Cool, whatever works! thanks for updating these!!1

PR #845, opened 2023-11-30
(PE-37243) update lein to v2.9.0

steveax:
In a previous commit [1], the jdk version used in the acceptance tests was updated to v11 and the version of lein that the test was using (2.7.1) incompatible with jdk11. This commit updates to the version of lein that runs elsewhere in CI.

[1]: 9db00cfee8c019f6a7590d76dbdc46886f760f4a


rileynewton (approved):
👍

joshcooper (approved):

PR #846, opened 2024-01-11
(maint) ensure int32_t type defined

h0tw1r3:
int32_t type def is optional in C++11, some vendors ship gcc suite with them automatically included, some don't.

observed `'uint32_t' does not name a type` with gcc 13, which will ship with Ubuntu 24.

https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes

### DIFF
--- a/lib/inc/pxp-agent/pxp_connector.hpp
+++ b/lib/inc/pxp-agent/pxp_connector.hpp
@@ -3,6 +3,7 @@
 #include <vector>
 #include <string>
 #include <functional>
+#include <cstdint>
 
 namespace PCPClient {
 struct ParsedChunks;


donoghuc (commented):
We use `uint32_t` all over the pxp-agent rpc protocol stack, In those cases we may rely on a bit on the compiler exporting some names in headers like https://github.com/puppetlabs/pxp-agent/blob/9b19bab0ebbaaeaaced1ff9e213167478be4a2fa/lib/inc/pxp-agent/pxp_connector_v1.hpp#L10 . 

If we need to fully switch to boost for these types it will need to be a bigger effort across multiple projects. 

Did you try explicitly including `#include <cstdint>` in this file? 

I'm unsure how the rest of the codebase including cpp-pcp-client etc would work if they are still attempting to use a native one.

h0tw1r3:
@donoghuc that makes sense. Updated PR based on your recommendation, [which is also the recommended way to support gcc-13's c++ standard library](https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes).

donoghuc (approved):
Thanks @h0tw1r3 If you find other places in this, or its puppetlabs deps ping me!

PR #847, opened 2024-01-17
(PE-36763) Migrate CURLOPT_PROTOCOLS{,_STR}

tlehman:
This PR removes the [deprecated CURLOPT_PROTOCOLS option](https://curl.se/libcurl/c/CURLOPT_PROTOCOLS.html) and replaces it with the new [CURLOPT_PROTOCOLS_STR option](https://curl.se/libcurl/c/CURLOPT_PROTOCOLS_STR.html).

Jira: https://perforce.atlassian.net/browse/PE-36763

Leatherman PR: https://github.com/puppetlabs/leatherman/pull/339

### DIFF
--- a/lib/src/modules/file.cc
+++ b/lib/src/modules/file.cc
@@ -94,7 +94,7 @@ namespace Modules {
     client_.set_ca_cert(ca);
     client_.set_client_cert(crt, key);
     client_.set_client_crl(crl);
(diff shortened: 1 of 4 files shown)


AriaXLi:
hank you for reporting this issue. While we agree this is likely an improvement, we do not anticipate addressing this any time soon due to other issues demanding precedence. As such, this ticket will be closed.

PR #851, opened 2024-02-17
Update Beaker exit_code method call

mhashizume:


acceptance/lib/pxp-agent/bolt_pxp_module_helper.rb
```diff
@@ -139,7 +139,7 @@ def ensure_failed(broker, targets, response_dataset, max_retries: 30, query_inte
     check_non_blocking_response(broker, identity, transaction_id, max_retries, query_interval) do |result|
       assert_equal('failure', result['status'], 'Action was expected to fail')
       assert_equal(nil, result['stdout'], 'Successful action expected to have no output on stdout')
-      assert_equal(1, result['exitcode'], "Successful action expected to have exit code #{exit_code}")
+      assert_equal(1, result['exitcode'], "Successful action expected to have exit code #{result.exit_code}")
```
joshcooper, line 142: I think maybe this needs to be `result['exitcode']` as I don't think `result` is a `Beaker::Result`, but maybe just a hash?

joshcooper (approved):

PR #853, opened 2024-03-15
(PE-37481) Return result of downloadFileFromMaster directly on retry

donoghuc:
In the original implementation of the retry logic (https://github.com/puppetlabs/pxp-agent/pull/852/files) a while loop using pointers was used to control max retries. On some platforms this approach resulted in broken pointers. Instead, use a for loop and return result of downloadFileFromMaster directly when it does not raise.

### DIFF
--- a/lib/src/module_cache_dir.cc
+++ b/lib/src/module_cache_dir.cc
@@ -296,27 +296,28 @@ namespace PXPAgent {
       // files remain in the cache_dir rather than being written out to a destination
       // elsewhere on the filesystem.
       auto destination = cache_dir / fs::path(file.get<std::string>("filename")).filename();
-      // retry logic for failed downloadFileFromMaster calls
-      // 2^5 = 32, so the maximum time it will wait is cumulatively: 2 + 4 + 8 + 16 + 32 = 62 seconds
+
+      // Try to download the file at most retry_count times. Catch any error and retry, on the last
+      // loop through throw the last caught error.
       int retry_count = 5;
       int retry_wait = 1;
-      fs::path* dlpath = nullptr;
-      while (dlpath == nullptr && retry_count > 0) {
+      for (int i = 1; i <= retry_count; i++) {
         try {
           LOG_DEBUG("getCachedFile: try max #{1} times", retry_count);
-          fs::path p = downloadFileFromMaster(master_uris, connect_timeout, timeout, client, cache_dir, destination, file);
-          dlpath = &p;
-        } catch (std::runtime_error &e) {
+          // Return early on success
+          return downloadFileFromMaster(master_uris, connect_timeout, timeout, client, cache_dir, destination, file);
+        }
+        catch (std::runtime_error &e) {
+          if (i == retry_count) {
+            throw Module::ProcessingError{lth_loc::format("Failed to download file with: {1}", e.what())};
+          }
           LOG_ERROR("getCachedFile: (std::runtime_error) {1}, waiting {2} seconds", e.what(), retry_wait);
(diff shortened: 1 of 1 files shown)


donoghuc:
For example on redhad 8, you can see that the pointer returned has lost its `fs::path` value:
```
2024-03-15 14:00:45.867743 ERROR puppetlabs.pxp_agent.request_processor:186 - Failed to execute the task for the non blocking 'task run' request (transaction 125). Unexpected error: boost::filesystem::copy_file: No such file or directory: "7^^3^?
```
I believe you were testing only on windows @tlehman , which is why this got past your manual verification.

lib/src/module_cache_dir.cc
```diff
@@ -296,27 +296,29 @@ namespace PXPAgent {
...
-      // retry logic for failed downloadFileFromMaster calls
-      // 2^5 = 32, so the maximum time it will wait is cumulatively: 2 + 4 + 8 + 16 + 32 = 62 seconds
+
+      // Try to download the file at most retry_count times. Catch any error and retry, on the last
+      // loop through throw the last caught error.
       int retry_count = 5;
       int retry_wait = 1;
-      fs::path* dlpath = nullptr;
-      while (dlpath == nullptr && retry_count > 0) {
+      for (int i = 0; i <= retry_count; i++) {
```
donoghuc, line 304: I suppose this is actually trying 6 times. I could start at i=1.

lib/src/module_cache_dir.cc
```diff
@@ -296,27 +296,29 @@ namespace PXPAgent {
...
           retry_count -= 1;
           retry_wait *= 2;
-          if(retry_count == 0 && dlpath == nullptr) {
-            throw Module::ProcessingError{lth_loc::format("Failed to download file with: {1}", e.what())};
-          }
         }
       }
-      return *dlpath;
+    // This line should never be hit (either returns early or raises), just satisfying warning about no return
+    return downloadFileFromMaster(master_uris, connect_timeout, timeout, client, cache_dir, destination, file);
```
donoghuc, line 322: This for sure seems silly, I added it because without I got a warning about this function not returning when the signature has a non-void return type. I'm not sure what the correct pattern here is when returning early from a loop. If i understand correctly the two outcomes are 1. return early or 2 raise exception. Not sure how this line could be reached. I'm kind of surprised it appeases the warning.

tlehman: Yeah this was why I had it be a pointer, but let's see if your changes get this unblocked

tlehman (approved):
LGTM

lib/src/module_cache_dir.cc
```diff
@@ -296,27 +296,29 @@ namespace PXPAgent {
...
+          return downloadFileFromMaster(master_uris, connect_timeout, timeout, client, cache_dir, destination, file);
+        }
+        catch (std::runtime_error &e) {
+          if (i == retry_count) {
+            throw Module::ProcessingError{lth_loc::format("Failed to download file with: {1}", e.what())};
+          }
           LOG_ERROR("getCachedFile: (std::runtime_error) {1}, waiting {2} seconds", e.what(), retry_wait);
           auto wait_seconds = pcp_util::chrono::seconds(retry_wait);
           pcp_util::this_thread::sleep_for(wait_seconds);
           retry_count -= 1;
```
mcdonaldseanp, line 317: Since you're testing the iterator in the for loop now we shouldn't also decrement the retry_count

donoghuc: oh, jeeze yeah thats a problem 😅

tlehman (approved):

